PR #4, opened 2016-10-10
Feature/elastic search

melsawy:
The branch include implementation for issues #5157, #5047 and #5149 which related to Pender as annotation, editing title/description, claims extensions to Media
Implementation details:
 - Dropped `data` column from media and account and added Pender data as a new annotations called `embed`
 - Migrate existing data to annotations
 - Dropped `title` and `description` columns from media 
 - Made `Media.url` optional
 - Allow user to update Pender data through `information` method that accept a json format. 
	*Example `info = {title: 'new title', description: 'new description'}`
 - Modify `jsondata` to be updated with overridden keys.
 - Added `Embed.quote` attribute to support quotes through `information` method that accept a json format. 
	*Example to add claims extension `info = {title: 'new title', description: 'new description', quote: 'Media quote'}`


app/controllers/api/v1/search_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+  module V1
+    class SearchController < Api::V1::BaseApiController
+      include SearchDoc
+
+      skip_before_filter :authenticate_from_token!
+
+      def create
+        query_string = params[:query]
+        repository = Elasticsearch::Persistence::Repository.new
+        repository.index = 'checkdesk_application_development_annotations'
```
caiosba: The index name should not be hard-coded... try to read this property: https://github.com/meedan/check-api/blob/develop/app/models/concerns/annotation_base.rb#L66

app/controllers/api/v1/search_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+        repository.type = 'annotation'
+        result = repository.search(query: { query_string: {query: query_string}})
```
caiosba: Do you need to re-instantiate the repository? I was hoping that you could use `Annotation.search`, as we do on annotation_base

db/migrate/20160927085541_remove_data_from_accounts.rb
```diff
@@ -0,0 +1,13 @@
+class RemoveDataFromAccounts < ActiveRecord::Migration
+  def change
+    pender = Bot.where(name: 'Pender').last
+    Account.find_each do |account|
+      em = Embed.new
+      em.embed = account.data
+      em.annotated = account
+      em.annotator = pender unless pender.nil?
+      em.save
```
caiosba: Please replace this `save` by `save!`

db/migrate/20161005065727_remove_title_and_description_from_medias.rb
```diff
@@ -0,0 +1,6 @@
+class RemoveTitleAndDescriptionFromMedias < ActiveRecord::Migration
+  def change
+    remove_column :medias, :title, :string
+    remove_column :medias, :description, :string
```
caiosba: `remove_column` doesn't take a third argument

test/controllers/graphql_controller_test.rb
```diff
@@ -105,11 +106,23 @@ def setup
...
+    p = create_project team: @team
+    m = create_valid_media project_id: p.id
+    query = "query GetById { media(id: \"#{m.id}\") { jsondata(context: \"#{p}\") } }"
+    post :create, query: query
+    assert_response :success
+    # calling without context
+    query = "query GetById { media(id: \"#{m.id}\") { jsondata(context: nil) } }"
+    post :create, query: query
+    assert_response :success
+  end
```
caiosba: I was expecting something different for this test. Two medias are created, for two different projects. One query for each project, then we assert that only the relevant media was returned. Then a third case for when a context is not provided. Also, the client doesn't send it as `nil`, but as an empty string.

test/controllers/search_controller_test.rb
```diff
@@ -0,0 +1,14 @@
...
+
+class SearchControllerTest < ActionController::TestCase
+  def setup
+    super
+    @controller = Api::V1::SearchController.new
+  end
+
+  test "should get create" do
+    #get :create
+    #assert_response :success
```
caiosba: Why this test is missing?

test/models/bot_test.rb
```diff
@@ -0,0 +1,16 @@
...
+  test "should create bot" do
+    assert_difference 'Bot.count' do
+      create_bot
+    end
+  end
+
+  test "should not save bot without name" do
+    project = Project.new
+    assert_not project.save
+  end
```
caiosba: Should be "bot" instead of "project" here :)

app/controllers/api/v1/search_controller.rb
melsawy: no will use Annotation.search

db/migrate/20161005065727_remove_title_and_description_from_medias.rb
melsawy: Done

test/controllers/search_controller_test.rb
melsawy: Will add this test when back to issue #4830

test/models/bot_test.rb
melsawy: Done

PR #6, opened 2016-10-16
Feature/5242 refactor embed annotation

melsawy:
* Added new fields to embed annotations: 
  - `title`: string
  - `description`: string
  - `username`: string
  - `published_at`: integer (UNIX timestamp)
  - `quote`: string
* Populate the above fields by reading them from the Pender data
* Re-write migration

### DIFF
--- a/app/models/annotations/embed.rb
+++ b/app/models/annotations/embed.rb
@@ -1,11 +1,31 @@
 class Embed
   include AnnotationBase
 
-  attribute :embed, String, presence: true
-  validates_presence_of :embed
+  attribute :title, String
+  attribute :description, String
+  attribute :username, String
+  attribute :published_at,  Integer
+  attribute :quote, String
+  attribute :embed, String
+  validate :exist_of_title_or_embed
 
   def content
-    { embed: self.embed }.to_json
+    {
+      title: self.title,
+      description: self.description,
+      username: self.username,
+      published_at: self.published_at,
+      quote: self.quote,
+      embed: self.embed
+    }.to_json
+  end
+
+  private
+
+  def exist_of_title_or_embed
+    if self.embed.blank? and self.title.blank?
+      errors.add(:base, "Should fill at least one filed [title or embed]")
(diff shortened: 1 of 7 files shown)


app/models/annotations/embed.rb
```diff
@@ -1,11 +1,31 @@
...
+      quote: self.quote,
+      embed: self.embed
+    }.to_json
+  end
+
+  private
+
+  def exist_of_title_or_embed
+    if self.embed.blank? and self.title.blank?
+      errors.add(:base, "Should fill at least one filed [title or embed]")
```
caiosba: Typo: "filed"

app/models/media.rb
```diff
@@ -43,14 +43,13 @@ def jsondata(context = nil)
   def data(context = nil)
     #TODO:: change the assumsion for a one Pender result
     em_pender = self.annotations('embed').last
-    embed = em_pender.embed unless em_pender.nil?
+    embed = JSON.parse(em_pender.embed) unless em_pender.nil?
     em_u = self.annotations('embed')
     context_id = context.nil? ? nil : context.id
     em_u.reverse.each do |obj|
       if obj.context_id.to_i == context_id.to_i
```
caiosba: Instead of loading all annotations in memory and checking the context of each, I suggest to filter by context right away. The method allows that: https://github.com/meedan/check-api/blob/develop/app/models/concerns/annotation_base.rb#L41-L43. It's the second parameter of the `annotations` method.

infojunkie: There's a difference between the logic in this loop and the logic of the `annotations` method. Specifically, passing a `nil` context here only returns annotations with a nil context, not all annotations. So I suggest to keep them separate for now to complete the ticket and tag this code with another TODO.

melsawy: @infojunkie  passing a nil context return all annotations not annotation with context nil
we should fix `annotations('embed', nil)` to return annotations with nil context only

PR #11, opened 2016-10-20
Fix ES search results size to 10000

infojunkie:


infojunkie:
@caiosba plz review

lib/elasticsearch_relation.rb
```diff
@@ -20,7 +19,7 @@ def all
   end
 
   def count
-    Annotation.search(@params.merge({ size: 0 })).total
+    Annotation.search(@params).total
```
caiosba: If we have more than 10.000 the `total` will be wrong... I think you shouldn't change this line... or is there a reason I'm not seeing?

lib/elasticsearch_relation.rb
```diff
@@ -1,12 +1,11 @@
 # This class tries to reproduce an ActiveRecord relation
 class ElasticsearchRelation
   def initialize(params = {})
-    @params = params
+    @params = params.merge({ size: 10000 })
```
caiosba: I suggest the opposite: `{ size: 10000 }.merge(params)`, this way we can change it when calling

infojunkie: thanks

lib/elasticsearch_relation.rb
infojunkie: I tested that the number set in `size` has no effect on the `total` result. E.g. I set `size` to 10 and it still returned a correct total of 14.

caiosba: Crazy, but thanks

PR #15, opened 2016-10-21
Feature/5203 email link

infojunkie:


app/mailers/team_user_mailer.rb
```diff
@@ -5,7 +5,7 @@ def request_to_join(team, requestor, origin)
     if team && requestor
       @team = team
       @requestor = requestor
-      @url = URI.join(origin, "/members")
+      @url = origin.blank? ? '' : URI.join(origin, "/members")
```
infojunkie: @caiosba I'm not happy about checking for blank `origin` here because it should never happen in a real environment. I had to do this because tests create teams without `origin`. Adding a default value here https://github.com/meedan/check-api/blob/develop/lib/sample_data.rb#L272 causes subdomain checking to fail.

PR #30, opened 2016-11-04
Ticket #5390

melsawy:
Created a new annotation `undetermined` when `ProjectMedia` created

### DIFF
--- a/app/models/project_media.rb
+++ b/app/models/project_media.rb
@@ -5,7 +5,7 @@ class ProjectMedia < ActiveRecord::Base
   belongs_to :media
   belongs_to :user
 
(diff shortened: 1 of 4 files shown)


app/models/project_media.rb
```diff
@@ -31,6 +31,16 @@ def project_id_callback(value, mapping_ids = nil)
     mapping_ids[value]
   end
 
+  def set_initial_media_status
+    st = Status.new
+    st.annotated = self.media
+    st.context = self.project
+    st.annotator = self.user
+    st.status = 'undetermined'
```
infojunkie: The default status is now obtained by the static method `Status.default_id`

PR #51, opened 2017-01-03
Feature/5222 refactor annotation

melsawy:
- Refactor Annotation to refer to ProjectMedia / ProjectSource
- Moved methods  of `Media` which take a context to `ProjectMedia`
- Removed `cached_annotations`
- Show media page given `ProjectMediaId` so the URL format will be `project/id/media/pm_id`
- List medias on project page and search page using `ProjectMedia` instead of `Media`
- Added a new type `project_media_type.rb` for graphql 
- Refactor search code to return `ProjectMedias` instead of `Medias`
- Migrated media annotations and set annotated with `ProjectMedia` object
- Fix all tests


app/graph/types/project_media_type.rb
```diff
@@ -0,0 +1,77 @@
...
+  field :updated_at, types.String
+  field :created_at, types.String
+  field :media_id, types.Int
+  field :project_id, types.Int
+  field :user_id, types.Int
+  field :permissions, types.String
+  field :project do
+    type -> { ProjectType }
+
+    resolve -> (project_media, _args, _ctx) {
```
infojunkie: We should no longer need `_ctx` anywhere, right?

app/graph/types/project_media_type.rb
```diff
@@ -0,0 +1,77 @@
...
+    }
+  end
+
+  connection :tags, -> { TagType.connection_type } do
+    resolve ->(project_media, _args, _ctx) {
+      call_method_from_context(project_media, :tags)
+    }
+  end
+
+  instance_exec :jsondata, &GraphqlCrudOperations.field_with_context
```
infojunkie: We should no longer need `field_with_context`, right?

app/models/concerns/annotation_base.rb
```diff
@@ -24,8 +24,8 @@ def define_annotation_relation_method
           query = self.annotation_query(type, context)
           klass = (type.blank? || type.is_a?(Array)) ? Annotation : type.camelize.constantize
           relation = klass.where(query)
-          relation = relation.where(context_id: nil) if context == 'none'
-          relation = relation.where.not(context_id: nil) if context == 'some'
+          #relation = relation.where(context_id: nil) if context == 'none'
```
infojunkie: Completely deleting unneeded lines is better than commenting them.

app/graph/types/project_media_type.rb
caiosba: It's part of the signature provided by the `graphql` gem... actually, we should only be using `_ctx`, instead of `ctx` (which means that we should not be using the GraphQL context anymore)

infojunkie: OK cool. I thought it had to do with our own "context" which we're trying to remove... A good reminder that choosing names that are too generic can lead to confusion.

app/models/concerns/media_information.rb
```diff
@@ -6,60 +6,39 @@ module MediaInformation
   def set_information
```
infojunkie: is `MediaInformation` still needed? maybe we can consider merging it back inside `ProjectMedia`. Also, `information` is not a suitable name for what this thing is, because it's too generic a name. What is this thing that we call `information`?

app/models/project_media.rb
```diff
@@ -1,11 +1,16 @@
...
 
   belongs_to :project
   belongs_to :media
   belongs_to :user
+  has_annotations
 
-  after_create :set_initial_media_status, :add_elasticsearch_data
+  include MediaInformation
+
+  after_create :set_quot_information, :set_initial_media_status, :add_elasticsearch_data
```
infojunkie: typo: `set_quot_information` should be `set_quote_information`.

app/models/project_media.rb
```diff
@@ -58,8 +63,8 @@ def add_elasticsearch_data
     ms.team_id = p.team.id
     ms.project_id = p.id
     ms.set_es_annotated(self)
-    ms.status = m.last_status(self.project)
-    data = m.data(self.project)
+    ms.status = self.last_status
+    data = self.data
```
infojunkie: Like `information`, `data` is too generic a name. We need a more descriptive name for what this thing is.

app/models/project_media.rb
```diff
@@ -69,4 +74,50 @@ def add_elasticsearch_data
     #ElasticSearchWorker.perform_in(1.second, YAML::dump(ms), YAML::dump({}), 'add_parent')
   end
 
+  def cached_annotations(type = nil)
```
infojunkie: These annotations are no longer cached, and we should rename the method accordingly.

app/models/project_media.rb
```diff
@@ -69,4 +74,50 @@ def add_elasticsearch_data
...
+    self.annotations.where(annotation_type: type)
+  end
+
+  def get_media_annotations(type = nil)
+    em = self.media.annotations.where(annotation_type: type).last
+  end
+
+  def jsondata
```
infojunkie: If this is only used by GraphQL `ProjectMedia` type, we can easily have the code there call `to_json` instead of a specialized method just for that.

app/models/project_media.rb
```diff
@@ -69,4 +74,50 @@ def add_elasticsearch_data
...
+    self.created_at.to_i.to_s
+  end
+
+  def overriden_embed_attributes
+    %W(title description username)
+  end
+
+  private
+
+  def set_quot_information
```
infojunkie: typo

app/graph/types/project_media_type.rb
```diff
@@ -0,0 +1,128 @@
...
+  end
+
+  connection :tags, -> { TagType.connection_type } do
+    resolve ->(project_media, _args, _ctx) {
+      call_method_from_context(project_media, :tags)
+    }
+  end
+
+  instance_exec :media, &GraphqlCrudOperations.field_verification_statuses
+  instance_exec :jsondata, &GraphqlCrudOperations.field_with_context
```
infojunkie: Same comment as for `data` and `information`: this name needs to be more precise and descriptive. Actually, since `jsondata` is just `data.as_json`, this should be renamed to whatever will replace `data` and perform `as_json` right here.

app/graph/mutations/media_mutations.rb
```diff
@@ -4,8 +4,7 @@ module MediaMutations
     account_id: 'int',
     project_id: 'int',
```
infojunkie: why is `project_id` part of `media_mutations` ?

infojunkie: actually, we may no longer need to expose a `Media` mutation, since all handling will happen though `ProjectMedia`.

melsawy: `project_id` is part of  `media_mutations` to assign a media to project.
if we remove `Media` mutation then we will need to pass URL or Quote to `ProjectMedia` mutation to create a media object then associate it to `ProjectMedia`

app/models/concerns/annotation_base.rb
```diff
@@ -24,8 +24,6 @@ def define_annotation_relation_method
           query = self.annotation_query(type, context)
           klass = (type.blank? || type.is_a?(Array)) ? Annotation : type.camelize.constantize
           relation = klass.where(query)
-          relation = relation.where(context_id: nil) if context == 'none'
-          relation = relation.where.not(context_id: nil) if context == 'some'
           relation.order('id DESC')
```
caiosba: The idea of those conditions was to get the annotations given a project, which is not a context anymore, but an annotated, right? So, should we really remove those two lines or should we replace `context_id` by `annotated_id`?

melsawy: Removed those lines

app/graph/mutations/media_mutations.rb
melsawy: Yes Caio and agreed with Karim to do this later, for now I created a ticket #5587

app/graph/types/media_type.rb
```diff
@@ -35,61 +35,14 @@
...
-  connection :annotations, -> { AnnotationType.connection_type } do
-    argument :context_id, types.Int
-
-    resolve ->(media, args, ctx) {
-      media.cached_annotations(annotation_types, media_context(media, args, ctx))
-    }
-  end
-
-  field :annotations_count do
+  field :pm_dbid do
```
melsawy: I called pm_dbid with context_project  `media.pm_dbid(ctx[:context_project])` to return ProjectMediaId as redirect URL format is `project/id/media/pm_id` and should remove this method after #5587 implementation

app/models/annotations/status.rb
```diff
@@ -92,8 +93,21 @@ def update_elasticsearch_status
...
+      values = Status.possible_values(annotated, context)
       errors.add(:base, 'Status not valid') unless values[:statuses].collect{ |s| s[:id] }.include?(self.status)
     end
   end
+
+  def get_annotated_and_context
+    if self.annotated_type == 'ProjectMedia' || self.annotated_type == 'ProjectSource'
+      annotated = self.annotated.media if self.annotated.respond_to?(:media)
+      annotated = self.annotated.source if obj.nil? and self.annotated.respond_to?(:source)
+      context = self.annotated.project if self.annotated.respond_to?(:project)
```
melsawy: `ProjectMedia` and `ProjectSource` context will be project otherwise it'll be `self.context`
also next step is to remove `context` from annotations

app/graph/types/project_type.rb
```diff
@@ -23,7 +23,11 @@
     }
   end
 
-  connection :medias, MediaType.connection_type, property: :eager_loaded_medias
+  connection :medias, -> { ProjectMediaType.connection_type } do
```
melsawy: Created a ticket #5587

PR #54, opened 2017-01-24
Refactoring/5240 single status annotation

melsawy:
1- Added a new method `get_annotations_log` to get annotation from annotation plus status versions
2- Updated permissions to allow user to update status and destroy versions
3- Migrated existing data to single status
4- Created a new graphql type `version` (use this type to delete status from log)
5- Override graphql annotations ID to solve same annotation ID for all status 
6- Added required tests


app/models/ability.rb
```diff
@@ -53,6 +53,11 @@ def owner_perms
     can :destroy, [Annotation, Comment, Tag, Status, Flag] do |obj|
       obj.get_team.include? @context_team.id
     end
+    can :destroy, PaperTrail::Version do |obj|
+      s = nil
```
infojunkie: don't need this

test/controllers/graphql_controller_test.rb
```diff
@@ -632,7 +640,7 @@ def setup
     query = "query { project(id: \"#{p.id}\") { project_medias(first: 10000) { edges { node { permissions, annotations(first: 10000) { edges { node { permissions } }  } } } } } }"
     @request.headers.merge!({ 'origin': 'http://team.localhost:3333' })
 
-    assert_queries (2 * n + 15) do
+    assert_queries (5 * n ) do
```
infojunkie: can we optimize this again?

PR #57, opened 2017-02-02
Ticket #5640

melsawy:


app/models/annotations/status.rb
```diff
@@ -96,6 +96,7 @@ def destroy
       Annotation.find(self.id).destroy
     else
       widget.paper_trail.without_versioning do
+        widget.origin = self.origin
```
infojunkie: Are we still using an `origin` that is passed from the controller all the way to the models? @caiosba ?

caiosba: Yes... we didn't remove it when we removed current user, current team and current project. But we should have done the same to origin too. BTW, it is used only on Slack notifications and e-mail notifications. Actually, since we don't have subdomains anymore, I think that we don't need to receive the origin from the client. We can read it from `checkdesk_client` config (unless we want to use the same backend for multiple clients... do we?)

infojunkie: Let's use `checkdesk_client` for now.

caiosba: Ok, I created a ticket.

PR #59, opened 2017-02-06
Fix/5630 broken links

melsawy:
- Fix slack notification links
- Replace `origin` by `checkdesk_client` config


caiosba:
Sawy, the changes here look good. But there are still a few places that need to be changed:

- `app/controllers/api/v1/graphql_controller.rb`: no need to set the origin on the context
- `config/application.rb`: the CORS origin should be a string instead of a regular expression
- `test/controllers/graphql_controller_test.rb`: remove all lines like this: `@request.headers.merge!({ 'origin': '...'  })`
- `app/controllers/api/v1/confirmations_controller.rb`: string instead of regular expression

PR #60, opened 2017-02-06
Fix/5665 incorrect attribution

melsawy:


app/models/annotations/status.rb
```diff
@@ -105,6 +105,10 @@ def destroy
 
   private
 
+  def set_annotator
+    self.annotator = User.current unless User.current.nil?
+  end
+
```
caiosba: We already have this method here: https://github.com/meedan/check-api/blob/develop/app/models/concerns/annotation_base.rb#L266. Why we need it here too?

melsawy: Override it for status only to set annotator with current user as base one not set annotator if it set and this behavior not working with revisions and always show creator on annotations logs https://mantis.meedan.com/view.php?id=5665#c19477

infojunkie: I agree with Sawy: this is actually a behaviour of the "singleton" annotation (of which status is the only example currently). Eventually, we will move singleton-specific code to a child class of Annotation that Status will inherit from.

PR #76, opened 2017-02-22
Feature/5160 move reports

melsawy:
- Add projects list to `/me` endpoint
- Update ES after move media to other project
- Add tests

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -83,7 +83,6 @@ class User < ActiveRecord::Base
       current_team: self.current_team,
       teams: self.user_teams,
       team_ids: self.team_ids,
-      projects: self.user_projects,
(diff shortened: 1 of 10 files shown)


app/models/user.rb
```diff
@@ -114,6 +115,14 @@ def user_teams
     teams.to_json
   end
 
+  def user_projects
+    projects = Hash.new
+    self.teams.each do |t|
```
caiosba: Will this include teams that the user requested access but is not approved yet?

melsawy: yes will change it

melsawy: Done with a unit test

caiosba: Thanks!

PR #85, opened 2017-03-06
Ticket #5755: Added info on logger about access denied on Admin UI

danielafeitosa:


infojunkie:
Thanks Daniela. Let's merge that tomorrow after deployment to live.

PR #87, opened 2017-03-10
Ticket #5645: Password reset flow

danielafeitosa:
- On Admin UI, a "Send reset password instructions" button appears to
  every user with blank provider (blank === email)
- Clicking this button triggers the regular password reset flow: email
  is sent to user with a temporary link to reset their password
- The e-mail is sent in English
- The user receives an email with a link to reset password and click on
  it
- The user sees a page with the fields "New password" and "Confirm new
  password"
- If the password is OK (8 characters length and password/password
  confirmation match), the user sees the json response for success
  `{:message=>{:type=>"success"}, :status=>200}`
- If the password is not OK the user needs to fill in the fields again


infojunkie:
Looks good to me

PR #95, opened 2017-03-23
Feature/5781 csv export

danielafeitosa:


app/models/annotations/dynamic.rb
```diff
@@ -19,15 +19,8 @@ class Dynamic < ActiveRecord::Base
...
-
-      @fields ||= self.fields
-
-      @fields.each do |f|
-        response = f.value if f.field_name =~ /^response_/
-        note = f.value if f.field_name =~ /^note_/
-        task = Task.find(f.value).label if f.field_name =~ /^task_/
-      end
+      response, note, task = self.values(['response', 'note', 'task'], '-').values_at('response', 'note', 'task')
+      task = Task.find(task).label
```
caiosba: I think that those two lines are leading to the failing test

infojunkie: The code assumes that the task is always there (as it used to before)

app/models/annotations/dynamic.rb
```diff
@@ -19,15 +19,8 @@ class Dynamic < ActiveRecord::Base
 
   def slack_message
     if !self.set_fields.blank? && self.annotation_type =~ /^task_response/
-      response = note = task = '-'
-
-      @fields ||= self.fields
```
caiosba: This was added for performance reasons, why was it removed?

infojunkie: I removed it because its purpose was not clear - plz mark performance optimizations with a comment. I will bring it back.

PR #102, opened 2017-03-28
 Ticket #5872: Edit README to include info about new settings on Admi…

danielafeitosa:
…n UI

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,8 +98,8 @@ Clients should send the `Accept-Language` header in order to get localized conte
 
 #### Add new settings fields
 
-* Create a method on model to receive the data and set it to the setting field
-   On `app/models/team.rb`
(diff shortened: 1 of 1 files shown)


infojunkie:
Looks good. Some comments:
- The paragraph 
>   For `string` fields the content should be displayed on a `text_field` and have an example for the field. Just include `render_settings('field')` and add an example in a partial file, like `app/views/rails_admin/main/_suggested_tags.html.erb`. Also, the `formatted_value` should be included like `formatted value { bindings[:object].get_suggested_tags }`.

is mistakenly enclosed in a code block.

- Try to keep both 
> `show` block
and
> `edit` block 
in the same indentation level, ideally the first level.

- Edit
> Create a method on model to receive the data and set it to the setting field On `app/models/team.rb`
to 
> Create a method on model to receive the data and set it to the setting field. For example, in `app/models/team.rb`:

PR #108, opened 2017-04-05
Ticket #5755: Add AdminAbility to be used on Admin UI

danielafeitosa:
The abilities used on Admin UI will not use context_team and check
permissions based on the teams that the logged in user is admin


caiosba:
If there are things in common between those two ability files, can we put the common methods in a `module` and `include` it in both?

PR #109, opened 2017-04-18
Feature/5654 machine translation

melsawy:


app/workers/machine_translation_worker.rb
```diff
@@ -0,0 +1,41 @@
+class MachineTranslationWorker
+  include Sidekiq::Worker
+
+  def perform(target, author)
+    target = YAML::load(target)
+    author = YAML::load(author)
+    field = DynamicAnnotation::Field.joins(:annotation).where('annotations.annotation_type' => 'language', 'annotations.annotated_type' => target.class.name, 'annotations.annotated_id' => target.id.to_s, field_type: 'language').first
+    src_lang = field.value unless field.nil?
```
infojunkie: don't need `unless` since `src_lang` will get a value anyway

app/workers/machine_translation_worker.rb
```diff
@@ -0,0 +1,41 @@
...
+    unless mt.nil? or src_lang.nil?
+      translations = []
+      text = target.text
+      # Assume languages format is ['en', 'ar', ...]
+      languages = target.project.settings[:languages] unless target.project.settings.nil?
+      languages = languages - [src_lang] unless languages.nil?
+      languages.each do |lang|
+        begin
+          response = AlegreClient::Request.get_mt(CONFIG['alegre_host'], { text: text, from: src_lang, to: lang }, CONFIG['alegre_token'])
+          mt_text = response['type'] == 'mt' ? response['data'] : nil
```
infojunkie: Need to log (here or in `AlegreClient`) if we got something other than a good response

app/workers/machine_translation_worker.rb
```diff
@@ -0,0 +1,41 @@
...
+        rescue
+          mt_text = nil
+        end
+        unless mt_text.nil?
+          lang_name = TwitterCldr::Shared::LanguageCodes.to_language(lang, :iso_639_1)
```
infojunkie: We don't need to store the language name in the annotation, since we don't know in advance the locale which should show the name of the target language.

app/models/concerns/project_media_creators.rb
```diff
@@ -49,6 +49,21 @@ def create_annotation
     end
   end
 
+  def create_mt_annotation
+    bot = Bot::Alegre.default
+    if !bot.nil? && bot.should_classify?(self.text)
+      languages = self.project.settings[:languages] unless self.project.settings.nil?
```
caiosba: I suggest to replace this line by `languages = self.project.get_languages`, which already handles if settings are not present (among other things) and makes the code simpler.

app/workers/machine_translation_worker.rb
```diff
@@ -0,0 +1,41 @@
...
+    target = YAML::load(target)
+    author = YAML::load(author)
+    field = DynamicAnnotation::Field.joins(:annotation).where('annotations.annotation_type' => 'language', 'annotations.annotated_type' => target.class.name, 'annotations.annotated_id' => target.id.to_s, field_type: 'language').first
+    src_lang = field.value unless field.nil?
+    mt = target.annotations.where(annotation_type: 'mt').last
+    unless mt.nil? or src_lang.nil?
+      translations = []
+      text = target.text
+      # Assume languages format is ['en', 'ar', ...]
+      languages = target.project.settings[:languages] unless target.project.settings.nil?
```
caiosba: Same as above: I suggest to replace this line by languages = self.project.get_languages, which already handles if settings are not present (among other things) and makes the code simpler.

app/workers/machine_translation_worker.rb
```diff
@@ -0,0 +1,41 @@
...
+      languages = languages - [src_lang] unless languages.nil?
+      languages.each do |lang|
+        begin
+          response = AlegreClient::Request.get_mt(CONFIG['alegre_host'], { text: text, from: src_lang, to: lang }, CONFIG['alegre_token'])
```
caiosba: This code should not be here. It should be in `app/models/bot/alegre.rb`. The code here should call the bot and ask the bot to do this job. Something like `Bot::Alegre.default.get_mt_from_alegre`.

app/workers/machine_translation_worker.rb
```diff
@@ -0,0 +1,41 @@
...
+          mt_text = response['type'] == 'mt' ? response['data'] : nil
+        rescue
+          mt_text = nil
+        end
+        unless mt_text.nil?
+          lang_name = TwitterCldr::Shared::LanguageCodes.to_language(lang, :iso_639_1)
+          lang_name = lang_name.blank? ? lang : lang_name.downcase
```
caiosba: Until here should be the bot's responsibility. Agree with Karim, no need to store the localized language name... just create a helper on `Bot::Alegre` that we can call later to expose it on GraphQL (like we do for the source language).

app/workers/machine_translation_worker.rb
```diff
@@ -0,0 +1,41 @@
...
+          translations << { lang: lang, lang_name: lang_name, text: mt_text }
+        end
+      end unless languages.nil?
+      unless translations.blank?
+        # Delete old versions
+        mt_field = DynamicAnnotation::Field.joins(:annotation).where('annotations.annotation_type' => 'mt', 'annotations.annotated_type' => target.class.name, 'annotations.annotated_id' => target.id.to_s, field_type: 'json').first
+        mt_field.versions.destroy_all
+        mt = mt.load
+        User.current = author
```
caiosba: We should avoid setting the `User.current`, since it can cascade to other layers and cause undesired effects... what you need that for? If you can set the `mt.annotator = author` on hand, better.

caiosba:
Sawy, please notice that this PR introduces new Code Climate issues, please fix them before merging.

app/workers/machine_translation_worker.rb
melsawy: Paper trail add a new version only if `User.current` is set.

caiosba: In that case, please add `User.current = nil` after `mt.save!`.

app/workers/machine_translation_worker.rb
melsawy: May be the `src_lang` not  exist for any reason so I think I should handle a `nil` case to be in safe side

infojunkie: what will be the value of `src_lang` if `field` is nil ?

caiosba: Since this runs in background, why not call the bot to recognize the source language in case it is not present?

infojunkie: yes sounds like a good idea

caiosba: e.g.: `src_lang = field.nil? ? Alegre.default.get_language_from_alegre(text, target) : field.value`

melsawy: Done

app/workers/machine_translation_worker.rb
melsawy: Done

app/workers/machine_translation_worker.rb
melsawy: Done

PR #111, opened 2017-04-27
Feature/5651 support check client

danielafeitosa:


app/graph/types/project_media_type.rb
```diff
@@ -152,6 +152,14 @@
     }
   end
 
+  field :translations_count do
+    type types.Int
+
+    resolve ->(project_media, _args, _ctx) {
+      project_media.get_annotations('translation').count
+    }
+  end
```
caiosba: Could it be a generic GraphQL field that receives the annotation type as a parameter?

danielafeitosa: Could it be adding annotation_type as parameter in this field? https://github.com/meedan/check-api/blob/7a2c94679228dcf0f3295ecba46ac8a5f557f797/app/graph/types/project_media_type.rb#L86-L92

caiosba: Yes... when `annotation_type` is present, return `get_annotations(annotation_type).count`, otherwise keep the current behavior.

danielafeitosa: I added `annotations` to be returned on ProjectMedia following the same idea

app/graph/types/project_media_type.rb
```diff
@@ -85,9 +85,14 @@
 
   field :annotations_count do
```
caiosba: Please add a test to this field on `test/controllers/graphql_controller_test.rb`

app/graph/types/project_media_type.rb
```diff
@@ -85,9 +85,14 @@
...
     type types.Int
+    argument :annotation_type, types.String
 
-    resolve ->(project_media, _args, _ctx) {
-      project_media.get_versions_log_count
+    resolve ->(project_media, args, _ctx) {
```
infojunkie: There's a inconcistency here with 

```connection :annotations, -> { AnnotationType.connection_type } do
    argument :annotation_type, !types.String

    resolve ->(project_media, args, _ctx) {
      project_media.get_annotations(args['annotation_type'].split(',').map(&:strip))
    }
  end
``` 
which does NOT handle blank `annotation_type` BUT handles multiple types separated by commas.

caiosba: I don't think that field `annotations` should allow all annotations to be retrieved... but I think that `annotations_count` should handle multiple annotation types.

infojunkie: do we have a guarantee that `project_media.get_versions_log_count === project_media.get_annotations([all types]).count` ?

caiosba: No, and usually they won't be equal. For example, the `embed` annotation created by Pender, the first status that is created automatically, etc, are not counted in the `log_size`.

infojunkie: In that case, this implementation is not consistent with itself - maybe it would be better to replace the blank case with an explicit name like `log`.

danielafeitosa: The changes on `annotations_count` was added before the need to get multiple annotation types and I forgot to include this case. I'll include it now.
About the `blank` case, it was already there and probably is being used somewhere. We can create the `log` and find who uses it and rename there

app/graph/types/project_media_type.rb
```diff
@@ -83,10 +83,10 @@
     }
   end
 
-  field :annotations_count do
+  field :log_count do
     type types.Int
 
-    resolve ->(project_media, _args, _ctx) {
+    resolve ->(project_media, args, _ctx) {
```
infojunkie: `_args` not `args`

PR #115, opened 2017-05-15
Feature/5977 index account

melsawy:
- Index account
- Allow user to search by account username, title, description
- Re-index existing accounts

### DIFF
--- a/db/migrate/20170515161016_reindex_account_info.rb
+++ b/db/migrate/20170515161016_reindex_account_info.rb
@@ -3,7 +3,7 @@ class ReindexAccountInfo < ActiveRecord::Migration
     url = "http://#{CONFIG['elasticsearch_host']}:#{CONFIG['elasticsearch_port']}"
     client = Elasticsearch::Client.new url: url
     options = {
-      index: CheckElasticSearchModel.index_name,
+      index: CheckElasticSearchModel.get_index_name,
       type: 'media_search'
     }
     Account.find_each do |a|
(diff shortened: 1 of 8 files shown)


db/migrate/20170515161016_reindex_account_info.rb
```diff
@@ -0,0 +1,27 @@
+class ReindexAccountInfo < ActiveRecord::Migration
+  def change
+    url = "http://#{CONFIG['elasticsearch_host']}:#{CONFIG['elasticsearch_port']}"
+    client = Elasticsearch::Client.new url: url
+    options = {
+      index: CONFIG['elasticsearch_index'].blank? ? [Rails.application.engine_name, Rails.env, 'annotations'].join('_') : CONFIG['elasticsearch_index'],
```
infojunkie: There are already 2 places in the code that use this same line to get the ES index name. Please make a function in `CheckElasticSearchModel` that returns the index name and call it instead of this line.

db/migrate/20170515161016_reindex_account_info.rb
```diff
@@ -0,0 +1,27 @@
...
+      type: 'media_search'
+    }
+    Account.all.each do |a|
```
infojunkie: This will be a long migration. Was it tested on the full live data? Is there a way to avoid updating ES for each account separately.

caiosba: If you still need to loop through all accounts, use `find_each` instead of `all.each`: http://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-find_each

melsawy: Yes I tested it with the full live data.
I preferred to iterate on `Account` so I can update all medias related to same account with one query and don't think that I can avoid updating ES for each account separately

infojunkie:
Thanks Sawy lgtm

PR #140, opened 2017-07-06
Feature/update source model

melsawy:
- Add `team_id` to `Source`
 - Migrate existing sources by getting team via source.user
 - Add `source_id` to `User` 
- Migrate existing users by get the first source related to user


app/models/source.rb
```diff
@@ -4,10 +4,11 @@ class Source < ActiveRecord::Base
   has_many :project_sources
   has_many :projects , through: :project_sources
   belongs_to :user
```
caiosba: Sawy, pay attention that this `belongs_to` here will use the `user_id` in the source - so, `source.user` will return the creator, not the related user, which I think is the case, but just to clarify.

db/migrate/20170706201604_add_source_id_to_user.rb
```diff
@@ -0,0 +1,10 @@
+class AddSourceIdToUser < ActiveRecord::Migration
+  def change
+    add_reference :users, :source, index: true, foreign_key: true
+    User.find_each do |u|
+      s = Source.where(user_id: u.id).first
+      u.source_id = s.id
```
caiosba: I'm not sure if all users in the database have a source (should have)... please test with a dump from live

db/migrate/20170706201604_add_source_id_to_user.rb
```diff
@@ -0,0 +1,10 @@
...
+      u.save!
```
caiosba: I prefer to use `update_columns` in migrations instead of `save!`, because we usually don't want the callbacks to be called. Also, `update_columns` is faster. But if you have a reason to use `save!`, let me know plz.

db/migrate/20170706214243_add_team_to_sources.rb
```diff
@@ -0,0 +1,10 @@
+class AddTeamToSources < ActiveRecord::Migration
+  def change
+    add_reference :sources, :team, index: true, foreign_key: true
+    Source.find_each do |s|
+      t = s.user.teams.first
+      s.team = t
+      s.save!
```
caiosba: Same things as above - guard against users without team and try to use `update_columns` if possible.

PR #146, opened 2017-07-20
Feature/6111 index and search source information

melsawy:


app/models/project_source.rb
```diff
@@ -14,6 +14,8 @@ class ProjectSource < ActiveRecord::Base
   validates :source_id, uniqueness: { scope: :project_id }
   before_validation :set_account, on: :create
 
+  after_create :add_elasticsearch_data
```
caiosba: Sawy, is there a problem if we run this in background?

lib/check_search.rb
```diff
@@ -77,8 +74,9 @@ def should_hit_elasticsearch?
     !(@options['status'].blank? && @options['tags'].blank? && @options['keyword'].blank?)
   end
 
-  def medias_build_search_query
+  def medias_build_search_query(assocated_type = 'ProjectMedia')
```
caiosba: Typo: `assocated` should be `associated` :)

caiosba:
I just left two comments, Sawy... other than that looks good, thanks for the complete tests. Please also check the coverage and the Code Climate issue above.

app/models/project_source.rb
melsawy: Yes Caio always need to set parent document directory as we have an issue when create parent/child on background - we have a mantis ticket with this issue

lib/check_search.rb
melsawy: will fix it

app/models/project_source.rb
caiosba: Ok thanks

PR #154, opened 2017-08-09
Feature/6214 expose field value

amoedoamorim:


amoedoamorim:
Closing this because the codeclimate failure (Class has too many lines. [260/250]) requires a refactoring that won't be done at the moment.

PR #175, opened 2017-09-25
Feature/5961 user profile

melsawy:


app/models/ability.rb
```diff
@@ -158,7 +158,7 @@ def contributor_perms
     can :update, Embed
     can [:create, :update], ProjectSource, project: { team: { team_users: { team_id: @context_team.id }}}, source: { user_id: @user.id }
     can [:create, :update], Source do |obj|
-      obj.team_id == @context_team.id || obj.id == @user.source_id
+      (obj.team_id == @context_team.id && obj.user_id == @user.id) || (obj.team_id.nil? && obj.id == @user.source_id)
```
amoedoamorim: This looks like only the user who created the source is able to update it, is that right?

PR #187, opened 2017-10-12
Send global Slack notifications even for limited teams #6449

infojunkie:


lib/check_limits.rb
```diff
@@ -62,15 +62,19 @@ def can_use_checklist
...
   Bot::Slack.class_eval do
     alias_method :notify_slack_original, :notify_slack
 
     def notify_slack(model)
       p = self.get_project(model)
       t = self.get_team(model, p)
       unless t.nil?
-        self.notify_slack_original(model) unless t.get_limits_slack_integration == false
+        if t.get_limits_slack_integration == false
+          self.notify_super_admin(model, t, p)
```
caiosba: I think we'll need a test for this case

infojunkie: Done, thx https://github.com/meedan/check-api/commit/cc3ec056067c21df547e6a2d3b4615fd693dba6f

PR #193, opened 2017-10-24
Feature/6218 skipped integration tests 2

clarissacastella:


app/controllers/test_controller.rb
```diff
@@ -74,6 +93,28 @@ def new_source
     render_success 'project_source', ps
   end
 
+  def media_status
+    require 'sidekiq/testing'
+    Sidekiq::Testing.inline!
```
caiosba: If you really need Sidekiq here to run inline, please use it as a block, so it doesn't after other parts of the code. Like this:
```
Sidekiq::Testing.inline! do
   pm = ProjectMedia.find(params[:pm_id])
   s = pm.annotations.where(annotation_type: 'status').last.load
   s.status = params[:status]
   s.save!
   render_success 'project_media', pm
end
```
More at https://github.com/mperham/sidekiq/wiki/Testing (look for "Each of the above methods also accepts a block").

app/controllers/test_controller.rb
```diff
@@ -74,6 +93,28 @@ def new_source
...
+    pm = ProjectMedia.find(params[:pm_id])
+    s = pm.annotations.where(annotation_type: 'status').last.load
+    s.status = params[:status]
+    s.save!
+    render_success 'project_media', pm
+  end
+
+  def new_media_tag
+    require 'sidekiq/testing'
+    Sidekiq::Testing.inline!
```
caiosba: Same as comment above.

caiosba:
We need tests for the `TestController` methods two :)

They should be here: https://github.com/meedan/check-api/blob/develop/test/controllers/test_controller_test.rb

Check the examples there for reference.

clarissacastella:
Caio's comments addressed.
Please review again.

caiosba:
Looks good! If all tests pass and the coverage is 100%, feel free to merge.

PR #243, opened 2018-01-30
Feature/6607 required tasks

melsawy:


app/models/annotations/status.rb
```diff
@@ -124,4 +146,11 @@ def get_annotated_and_context
     end
     return annotated, context
   end
+
+  def can_verify_media
```
infojunkie: plz rename to `can_complete?`

app/models/annotations/status.rb
```diff
@@ -63,6 +79,12 @@ def self.possible_values(annotated, context = nil)
     statuses
   end
 
+  def self.can_resolved?(annotated)
```
infojunkie: plz rename to `is_resolved?`

PR #269, opened 2018-03-27
Ticket #6599: Translate Rails Admin strings with interpolation

danielafeitosa:


config/locales/en.yml
```diff
@@ -172,15 +172,16 @@ en:
         menu: Export project images
       delete_tasks:
         done: deleted the tasks
-        are_you_sure_you_want_to_delete_the_tasks_from_team: Are you sure you want to delete all tasks in team
-        all_the_following_tasks_will_be_deleted: '? The following tasks will be removed:'
+        are_you_sure_you_want_to_delete_the_tasks_from_team:
+          html: "Are you sure you want to delete all tasks in team <strong>%{team}</strong>? The following tasks will be removed:"
```
infojunkie: any reason why you're using a subkey? does the main `are_you_sure_you_want_to_delete_the_tasks_from_team` key collide with something?

PR #274, opened 2018-04-04
Fix/6708 rollback es items when fail to save

melsawy:
- Move Slack notification and ES index to `after_commit` callback
- Use https://github.com/grosser/test_after_commit to fire `after_commit` on testing mode.
- Move destroy ES items to background
- Fix testing


caiosba:
Looks good to me, Sawy, thanks. I just noticed that the `destroy_*` methods are very similar and there are four implementations of that... I think we can abstract in a single method. I'm talking about those lines:
```
return if self.disable_es_callbacks || RequestStore.store[:disable_es_callbacks]
options = {es_type: TagSearch, type: 'child'}
ElasticSearchWorker.perform_in(1.second, YAML::dump(self), YAML::dump(options), 'destroy')
```
This could be a single method that receives a `es_type` as argument.

PR #287, opened 2018-05-02
#6715 Get projects count for a team

amoedoamorim:


caiosba:
Missing a test for `projects_count`

PR #303, opened 2018-06-06
Fix/6942 search by custom status with hyphens

melsawy:


app/models/concerns/check_elastic_search_model.rb
```diff
@@ -46,19 +46,15 @@ def self.get_index_name
...
-      MediaSearch.migrate_es_data(source_index, target_index, mapping_keys)
-      sleep 1
-      MediaSearch.migrate_es_data(target_index, source_index, mapping_keys)
-      MediaSearch.index_name = source_index
-    rescue StandardError => e
-      Rails.logger.error "[ES MIGRATION] Could not start migation: #{e.message}"
-    end
+    # copy data to destination
+    MediaSearch.migrate_es_data(source_index, target_index)
+    sleep 20
```
infojunkie: is there a way to ask ES if the new index has finished processing?

lib/tasks/check_create_es_data_from_pg.rake
```diff
@@ -0,0 +1,27 @@
...
+  			obj.add_elasticsearch_data
+  			print '.'
+  		end
+  	end
+  	sleep 10
+  	# Add ES for annotations (child items)
+  	Annotation.find_each do |a|
+  		a = a.load
+      klass = a.class.name.downcase
+      if ['comment', 'tag', 'dynamic'].include?(klass)
```
infojunkie: can we use `a.respond_to("add_update_elasticsearch_#{klass}")` instead of hard-coding the types of the annotations?

melsawy: Sure

app/models/concerns/check_elastic_search_model.rb
melsawy: Yes there is a response for reindex api, we can parse it

PR #314, opened 2018-07-09
Feature/66981 refactor tag indexing

melsawy:


caiosba:
Sawy, the migration should also remove the `full_tag` field from the `data` field of the existing tags in the PG `annotations` table.

PR #319, opened 2018-07-19
Feature/6723 upgrade elasticsearch

melsawy:


config/initializers/location_search.rb
```diff
@@ -23,33 +24,26 @@ def lookup_type(type)
...
     geojson = JSON.parse(self.get_field_value(:response_geolocation))
     coordinates = geojson['geometry']['coordinates']
     indexable = geojson['properties']['name']
 
     if coordinates[0] != 0 || coordinates[1] != 0
+      # re-compute long value before sending to Elasticsearch
       location = {
         lat: coordinates[0],
-        lon: coordinates[1]
+        lon: ((coordinates[1].to_f + 180) % 360) - 180
```
infojunkie: what is this for?

db/migrate/20170720090144_add_es_mapping_for_account.rb
```diff
@@ -1,5 +1,6 @@
 class AddEsMappingForAccount < ActiveRecord::Migration
   def change
+    return
```
infojunkie: if this migration is no longer needed, we can delete it

lib/check_elastic_search.rb
```diff
@@ -1,133 +1,144 @@
...
+    # data.merge!(add_missing_fields(options))
+    fields = {'updated_at' => Time.now.utc}
+    options[:keys].each{|k| fields[k] = data[k] if !data[k].blank? }
+    client = MediaSearch.gateway.client
+    client.update index: CheckElasticSearchModel.get_index_alias, type: 'media_search', id: options[:doc_id], retry_on_conflict: 3,
+                  body: { doc: fields }
   end
 
-  def add_update_media_search_child(child, keys, data = {}, parent = nil)
+  # Keep this method until verify search feature.
```
infojunkie: what's that for?

lib/tasks/check_create_es_data_from_pg.rake
```diff
@@ -3,25 +3,75 @@ namespace :check do
...
+            doc.save!
+          rescue Exception => e
+            failed_items << {error: e, obj_id: obj.id, obj_class: obj.class.name}
+          end
+        end
+        print '.'
       end
-  	end
+    end
+    puts "Failed to index #{failed_items.size} items"
```
infojunkie: If zero failures plz print nothing

config/initializers/location_search.rb
melsawy: new versions of ES validate longitude value allowed range is  [-180..180]
I got an error `illegal longitude value`  while creating index from PG and used this formula to fix this error.

db/migrate/20170720090144_add_es_mapping_for_account.rb
melsawy: yes not needed, will delete it

lib/check_elastic_search.rb
melsawy: This method `add_missing_fields` was added to index all parent filed when using versions, now I use retry_on_conflict and just keep it until verify everything on QA then will remove it.
I did many tests locally and everything works well without this method.

lib/tasks/check_create_es_data_from_pg.rake
melsawy: Done

PR #333, opened 2018-10-17
Feature/6769 sahara import squashed

danielafeitosa:


caiosba:
There is a `import_spreadsheet` permission but it's not used anywhere

danielafeitosa:
> There is a `import_spreadsheet` permission but it's not used anywhere

The idea is to use it on `TeamImport` concern  :)

PR #359, opened 2019-01-21
Feature/6601 user can authin with multiple social accounts

melsawy:


app/controllers/api/v1/admin_controller.rb
```diff
@@ -20,8 +20,9 @@ def add_publisher_to_project
 
   # GET /api/admin/user/slack?uid=:uid
   def slack_user
-    user = User.where(provider: 'slack', uuid: params[:uid].to_s).last
-    user = { token: user.token } unless user.nil?
+    user = User.find_with_omniauth(params[:uid].to_s, 'slack')
+    slack_account = user.accounts.where(provider: 'slack').first unless user.nil?
```
infojunkie: this will not work in case of multiple Slack logins

app/mailers/registration_mailer.rb
```diff
@@ -10,9 +10,9 @@ def welcome_email(user, password=nil)
 
   def duplicate_email_detection(user, duplicate)
     @user = user
-    @duplicate_provider = duplicate.provider? ? duplicate.provider.camelcase : I18n.t(:mail_duplicate_email_exists_email)
-    @user_provider = user.provider? ? user.provider.camelcase : I18n.t(:mail_duplicate_email_exists_email)
-    @mail_body = (duplicate.provider? || user.provider?) ? :mail_duplicate_email_exists_body : :mail_duplicate_email_exists_body_both_emails
+    @duplicate_provider = duplicate.encrypted_password? ? I18n.t(:mail_duplicate_email_exists_email) : duplicate.get_social_accounts_for_login.first.provider
```
infojunkie: Does this logic allow a duplicate email for the SAME user?

app/controllers/api/v1/admin_controller.rb
caiosba: +1

app/controllers/api/v1/graphql_controller.rb
```diff
@@ -66,7 +66,7 @@ def format_error_message(e)
       private
 
       def authenticate_graphql_user
-        params[:query].to_s.match(/^((query )|(mutation[^\{]*{\s*(resetPassword|changePassword|resendConfirmation)))/).nil? ? authenticate_user! : authenticate_user
+        params[:query].to_s.match(/^((query )|(mutation[^\{]*{\s*(resetPassword|changePassword|userDisconnectLoginAccount|)))/).nil? ? authenticate_user! : authenticate_user
```
caiosba: Why you removed the `resendConfirmation` mutation from here?

app/graph/mutations/user_disconnect_login_account_mutation.rb
```diff
@@ -0,0 +1,21 @@
...
+  input_field :id, !types.Int
+
+  input_field :provider, !types.String
+
+  return_field :success, types.Boolean
+
+  return_field :user, UserType
+
+  resolve -> (_root, inputs, _ctx) {
+    user = User.where(id: inputs[:id]).last
```
caiosba: What prevents a user from disconnecting an account of another user? I think we should load from `User.current`... can we do that?

app/models/bot_user.rb
```diff
@@ -38,12 +38,11 @@ def confirmation_required?
 
   private
 
-  def set_uuid
-    self.uuid = ('check_bot_user_' + SecureRandom.hex(10))
-  end
+  # def set_uuid
+  #   self.uuid = ('check_bot_user_' + SecureRandom.hex(10))
+  # end
```
caiosba: Why this method is commented?

app/models/concerns/user_multi_auth_login.rb
```diff
@@ -0,0 +1,139 @@
...
+	  end
+
+	  def self.find_with_token(token)
+	    account = Account.where(token: token).last
+	    account.nil? ?  User.where(token: token).last : account.user
+	  end
+
+	  def get_social_accounts_for_login(provider = nil)
+	    s = self.source
+	    return nil if s.nil? || !ActiveRecord::Base.connection.column_exists?(:accounts, :provider)
```
caiosba: Why we need to check if a column exists in the table?

app/models/concerns/user_multi_auth_login.rb
```diff
@@ -0,0 +1,139 @@
...
+	    if provider.nil?
+	      a = s.accounts.where('provider IS NOT NULL')
+	    else
+	      a = s.accounts.where(provider: provider).first
```
caiosba: Does it mean that we don't support more than one account of the same provider for the same user?

app/models/concerns/user_multi_auth_login.rb
```diff
@@ -0,0 +1,139 @@
...
+	    providers = []
+	    accounts = self.get_social_accounts_for_login
+	    allow_disconnect =  (accounts.count == 1 && !self.encrypted_password?) ? false : true
+	    LOGINPROVIDERS.each do |p|
+	      a = accounts.select{|i| i.provider == p}.first
+	      if a.nil?
+	        providers << {key: p, connected: false}
+	      else
+	        info = a.omniauth_info.dig('info')
+	        if a.provider == 'slack'
```
caiosba: The code that is executed for the `if` block is the same as the `else` block, so why do we need this condition?

caiosba: Since the only difference between the code that is executed for the `if` block and the one executed for the `else` block is the `info` key, can we simplify this whole block?

db/migrate/20190107151222_add_omniauthinfo_to_accounts.rb
```diff
@@ -0,0 +1,17 @@
+class AddOmniauthinfoToAccounts < ActiveRecord::Migration
+  def change
+  	add_column :accounts, :omniauth_info, :text
+  	add_column :accounts, :uid, :string
+  	add_column :accounts, :provider, :string
```
caiosba: I think we should add an index for this `provider` column.

app/models/concerns/user_multi_auth_login.rb
infojunkie: We want to support more than one account of the same provider

db/migrate/20190107151222_add_omniauthinfo_to_accounts.rb
infojunkie: We should add an index for any column that will be used for search

app/models/account.rb
```diff
@@ -128,7 +130,7 @@ def set_omniauth_info_as_annotation
       em = Embed.new
       em.annotated = self
     end
-    em.embed_for_registration_account(self.user.omniauth_info)
+    em.embed_for_registration_account(self.omniauth_info)
```
infojunkie: `Embed` is a terrible name for what this represents. `Metadata` would be much more appropriate. Is it feasible to rename the annotation and associated functions? @caiosba and @melsawy

app/mailers/registration_mailer.rb
melsawy: No just need to catch a provider for duplicate email in case user tried to register with existing email to use it on email body

app/controllers/api/v1/graphql_controller.rb
melsawy: Removed by mistake will add it

app/graph/mutations/user_disconnect_login_account_mutation.rb
melsawy: Sure will depend on `User.current`

app/models/bot_user.rb
melsawy: No need to `uuid` as I removed it from `User` mode and use `uid` in `Account` and filled it for social account login only, will remove it from code

app/models/concerns/user_multi_auth_login.rb
melsawy: Old migration go through this method and raise an error as column not exists

app/models/bot_user.rb
caiosba: Yeah, in that case please add a migration to remove this column from the database too

app/models/concerns/user_multi_auth_login.rb
caiosba: It's better to check this in the migration then

app/models/concerns/user_multi_auth_login.rb
melsawy: Done

PR #361, opened 2019-01-24
Ticket #7161: Rescue and notify Errbit when can't delete team

danielafeitosa:


app/workers/team_deletion_worker.rb
```diff
@@ -0,0 +1,14 @@
+class TeamDeletionWorker
+  include Sidekiq::Worker
+
+  def perform(team_id, user_id)
+    team = Team.find_by_id(team_id)
+    return if team.nil?
+    begin
+      team.destroy
+    rescue StandardError => e
```
caiosba: If you want to capture an exception, shouldn't `destroy!` instead of `destroy`?

PR #364, opened 2019-01-30
Ticket #6601 - show an error message with user try to connect with ex…

melsawy:
…isting account


app/controllers/api/v1/omniauth_callbacks_controller.rb
```diff
@@ -27,6 +27,8 @@ def start_session_and_redirect
           user = User.from_omniauth(auth, current_api_user)
         rescue ActiveRecord::RecordInvalid => e
           session['check.error'] = e.message
+        rescue RuntimeError => e
+          session['check.warning'] = e.message;
```
caiosba: Please remove this `;`

PR #396, opened 2019-04-15
Fix/7489 recognize numbers in indian languages

melsawy:


infojunkie:
Thanks Sawy, great stuff. My only comment is that it would have been good to move the number recognition function to a specialized language module instead of putting it in the Smooch bot. Does not matter now, we can refactor it any time.

PR #461, opened 2019-11-04
Fix/7771 depend on media type for create listing

melsawy:


app/models/concerns/project_media_creators.rb
```diff
@@ -73,9 +73,8 @@ def set_quote_metadata
 
   protected
 
-  def create_video_or_image
-    extname = File.extname(self.file.path).delete('.')
-    m = VideoUploader.upload_extensions.include?(extname) ? UploadedVideo.new : UploadedImage.new
+  def create_video_or_image(media_type)
```
caiosba: Sawy, in order to guarantee backward compatibility, I think we should default `media_type` to `UploadedImage`

melsawy: Done

PR #471, opened 2019-11-21
7789 apply config at runtime

huslage:
This is work to have configurator run at run-time.


huslage:
Obsoleted by @hfikry92

PR #479, opened 2019-12-10
Adding gitlab environments to support config at runtime

huslage:


huslage:
This is not as important as I thought. The env vars we need to be set are in ECS configs, not in gitlab.

PR #480, opened 2019-12-10
471 hosam config runtime

hfikry92:
Please review


production/bin/start.sh
```diff
@@ -17,6 +17,19 @@ function config_replace() {
 	sed -i'.bak' -e ${CMD} ${FILE}
     fi
 }
+function getParsedGithubToken(){
```
infojunkie: As per https://mantis.meedan.com/view.php?id=7812, we no longer need `production/bin` since it was used for Linode deployments. So I suggest to not touch that folder for this present ticket.

production/bin_aws/start.sh
```diff
@@ -17,6 +17,20 @@ function config_replace() {
 	sed -i'.bak' -e ${CMD} ${FILE}
     fi
 }
+function getParsedGithubToken(){
```
infojunkie: Can you please document the purpose of this function?

production/bin_aws/start.sh
```diff
@@ -55,14 +69,14 @@ if [ -n "${PRIMARY}" ]; then
 fi
 
 # compile assets in the background, particularly the admin interface
-su ${DEPLOYUSER} -c "nice bundle exec rake assets:precompile" &
+#su ${DEPLOYUSER} -c "nice bundle exec rake assets:precompile" &
```
infojunkie: No need to comment out lines, better to remove them completely (along with their comments). Git history remains.

production/bin_aws/start.sh
infojunkie: Please use consistent spacing (e.g. between functions and between tokens) and consistent tabs.

production/bin/start.sh
hfikry92: Okay great to know :)

hfikry92:
Comments should be done. Still didn't eliminate the unneeded files from my work today.

huslage:
> Thanks. Another useful option here is `jq` which is a fantastic tool.

I didn't want to add another dependency into the mix. Python is already in the containers, so it was better to add this.

huslage:
@hfikry92 can you squash these commits please?

PR #485, opened 2019-12-17
Feature/7776 refactor archive annotation

danielafeitosa:
Add Perma.cc as option on keep bot
Create new archiver annotation "archiver" to replace the specific archivers (archive_org, archive_is, perma_cc)
Change the archivers field instances to be optional
Add a migration and task to replace the specific archivers annotation:
  * There will be one of type "archiver"
  * On DynamicAnnotation::Field, replaced the annotation with the new annotation type 'archiver'

### DIFF
--- a/app/models/bot/keep.rb
+++ b/app/models/bot/keep.rb
@@ -17,7 +17,8 @@ class Bot::Keep < BotUser
       'keep_backup',    # VideoVault
       'pender_archive', # Screenshot
       'archive_is',     # Archive.is
-      'archive_org'     # Archive.org
+      'archive_org',    # Archive.org
+      'perma_cc'        # Perma.cc
     ]
   end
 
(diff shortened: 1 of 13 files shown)


app/models/bot/keep.rb
```diff
@@ -17,7 +17,8 @@ def self.archiver_annotation_types
       'keep_backup',    # VideoVault
```
infojunkie: Is `keep_backup` still active on either Pender or Check API ?

app/models/bot/keep.rb
```diff
@@ -17,7 +17,8 @@ def self.archiver_annotation_types
...
       'pender_archive', # Screenshot
```
infojunkie: In a future refactoring, we would rename this to `screenshot`.

app/models/bot/keep.rb
danielafeitosa: It was removed from Pender. On CheckAPI it is not active now, there might be only the old annotations

infojunkie: OK thanks will file the a refactoring to review and remove Video Vault archives.

PR #506, opened 2020-01-30
Feature/7988 regex should search to match whole text or title

melsawy:


app/models/concerns/team_rules.rb
```diff
@@ -43,8 +43,20 @@ def get_smooch_message(pm)
...
-      obj.nil? && pm.report_type == 'claim' && !pm.text.match(/#{Regexp.new(value)}/).nil?
+    def title_matches_regexp(pm, obj, value)
+      return false unless obj.nil?
+      matches_regexp(pm.title, value)
+    end
+
+    def request_matches_regexp(pm, obj, value)
+      return false unless obj.nil?
+      smooch_message = get_smooch_message(pm)
+      return false if smooch_message.blank?
```
caiosba: Sawy, we should accept blank messages... for example, if there is a regular expression rule like `.*`, we want to match a request that is just a photo and no text, for example (this is similar to the keywords rule that matches zero words).

caiosba:
Looks good overall, I just added one comment.

app/models/concerns/team_rules.rb
melsawy: Done

PR #508, opened 2020-01-30
Adding support for honeycomb.io tracing.

huslage:


config/initializers/honeycomb.rb
```diff
@@ -0,0 +1,14 @@
...
+    sql.active_record
+    render_template.action_view
+    render_partial.action_view
+    render_collection.action_view
+    process_action.action_controller
+    send_file.action_controller
+    send_data.action_controller
+    deliver.action_mailer
+  ].freeze
+end
```
caiosba: please make sure it doesn't crash if the configs are not set... for example, you can wrap the whole block in a `if CONFIG['honeycomb_key'] && CONFIG['honeycomb_dataset']`

huslage: I tested this and it did not crash.

PR #519, opened 2020-02-03
remove asset compile from start.sh, add to dockerfile

huslage:


huslage:
Only effects the AWS environments.

PR #538, opened 2020-02-07
Sysops/8026 puma migrate

huslage:


config/environments/production.rb
```diff
@@ -54,17 +54,23 @@
...
+  # config/environments/production.rb
+  config.lograge.enabled = true
+  config.lograge.custom_options = lambda do |event|
+    options = event.payload.slice(:request_id, :user_id)
+    options[:params] = event.payload[:params].except("controller", "action")
+    options
+  end
+  config.lograge.formatter = Lograge::Formatters::Json.new
+  
+  # app/controllers/application_controller.rb
```
caiosba: as per the comment, this method should be added to `app/controllers/application_controller.rb` (there shouldn't be any method definition in `config/environments/*.rb`)... and probably this should be called as an `after_filter` for all requests, otherwise it's just a defined method that will never be called.

Dockerfile
```diff
@@ -6,19 +6,15 @@ ENV RAILS_ENV development
...
 # https://www.mikeperham.com/2018/04/25/taming-rails-memory-bloat/
 ENV MALLOC_ARENA_MAX 2
+ENV QT_QPA_PLATFORM offscreen
 
 RUN apt-get update -qq && apt-get install -y --no-install-recommends curl
 
 RUN curl -sL https://deb.nodesource.com/setup_12.x | bash -
 
-RUN apt-get update && apt-get install --no-install-recommends -y nodejs git build-essential libpq-dev sqlite3 libsqlite3-dev graphicsmagick \
+RUN apt-get update && apt-get install --no-install-recommends -y phantomjs nodejs git build-essential libpq-dev sqlite3 libsqlite3-dev graphicsmagick \
```
caiosba: so we are using `phantomjs` from the Debian package again? I thought it was crashing

config/environments/test.rb
```diff
@@ -56,17 +56,20 @@
...
+  config.lograge.formatter = Lograge::Formatters::Json.new
   
-  module ActiveSupport::TaggedLogging::Formatter
-    def call(severity, time, progname, data)
-      data = { msg: data.to_s } unless data.is_a?(Hash)
-      tags = current_tags
-      data[:tags] = tags if tags.present?
-      _call(severity, time, progname, data)
-    end
+  # app/controllers/application_controller.rb
```
caiosba: same as the comment above in `config/environments/production.rb`

config/initializers/sidekiq.rb
```diff
@@ -1,27 +1,6 @@
...
-    items = ctx.map {|c| c.split(' ') }.flatten
-    data[:sidekiq_context] = items if items.any?
-  end
-
-  # Replace default error handler
-  config.error_handlers << lambda do |ex, ctx|
-    Sidekiq::Logging.logger.warn(ex, job: ctx[:job]) # except job_str
-  end
-end
-
```
caiosba: why you removed this block? we need that... bad merge?

config/initializers/sidekiq.rb
```diff
@@ -1,27 +1,6 @@
 file = File.join(Rails.root, 'config', "sidekiq-#{Rails.env}.yml")
 file = File.join(Rails.root, 'config', 'sidekiq.yml') unless File.exist?(file)
 require File.join(Rails.root, 'lib', 'middleware_sidekiq_server_retry')
-require 'airbrake/sidekiq'
-
-Sidekiq::Logging.logger = Rails.logger.child
-
-Airbrake.add_filter(Airbrake::Sidekiq::RetryableJobsFilter.new)
```
caiosba: we need this line... this one should not be removed

config/initializers/sidekiq.rb
```diff
@@ -34,6 +13,9 @@
 
   Sidekiq.configure_server do |config|
     config.redis = redis_config
+    config.error_handlers << Proc.new do |e, context|
+      Airbrake.notify(e, context) if Airbrake.configured? && !e.is_a?(Elasticsearch::Transport::Transport::Errors::Conflict)
+    end
```
caiosba: bad merge? we don't need this block anymore

production/bin/start.sh
```diff
@@ -18,12 +18,13 @@ if [ ! -d "configurator" ]; then git clone https://${GITHUB_TOKEN}:x-oauth-basic
...
-puma="/app/current/tmp/puma-$RAILS_ENV.rb"
-
+puma="/app/current/tmp/puma-${RAILS_ENV}.rb"
 cp config/puma.rb $puma
-echo "pidfile '/app/current/tmp/pids/server-$RAILS_ENV.pid'" >> $puma
-echo "environment '$RAILS_ENV'" >> $puma
-echo "port $SERVER_PORT" >> $puma
-echo "workers 3" >> $puma
+cat << EOF >> $puma
+pidfile '/app/current/tmp/pids/server-${RAILS_ENV}.pid'
```
caiosba: I prefer to use `pwd` than having a part hard coded here

Dockerfile
huslage: it was crashing before I added that ENV var. Now it works fine.

config/initializers/sidekiq.rb
huslage: I added back this line and I get

`#17 7.850 NoMethodError: undefined method `child' for #<ActiveSupport::Logger:0x0000564df57ab1d0>`

caiosba: I mean, only this line: `Airbrake.add_filter(Airbrake::Sidekiq::RetryableJobsFilter.new)`, not the one above.

PR #550, opened 2020-02-27
7847 vframe alegre

infojunkie:
For @caiosba review plz.

Note that I've removing `AlegreClient` from the deps and I'd like to delete the `alegre_client` repo because it's an unncessary chore to keep the client in sync with changes to the API.


caiosba:
The general changes of:

Removing unused configuration keys
Replacing instance methods by class methods in Alegre Bot
Removing Alegre Client and replacing it by straight HTTP requests to the service
Look good! Other comments inline below.

caiosba:
The general changes of:

- Removing unused configuration keys
- Replacing instance methods by class methods in Alegre Bot
- Removing Alegre Client and replacing it by straight HTTP requests to the service

Look good! Other comments inline below.

caiosba:
The general changes of:

- Removing unused configuration keys
- Replacing instance methods by class methods in Alegre Bot
- Removing Alegre Client and replacing it by straight HTTP requests to the service

Look good! Other comments inline below.

app/models/bot/alegre.rb
```diff
@@ -96,32 +96,40 @@ def add_relationships(pm, pm_ids)
...
-      end
-    end
-    pm_ids = response.dig('results')&.collect{|r| r.dig('context', 'project_media_id')}
+  def self.get_image_similarities(pm)
+    return if pm.report_type != 'uploadedimage'
+
+    # Query for similar images.
+    similar = self.request_api('get', '/image/similarity', {
+      url: pm.media.file.file.public_url,
+      context: self.get_context(pm),
```
caiosba: If we send the full `context`, with both `team_id` and `project_media_id`, it means that on the Alegre side it will try to match the whole context and thus always return only the same project media, no?

caiosba: If we send the full `context`, with both `team_id` and `project_media_id`, it means that on the Alegre side it will try to match the whole context and thus always return only the same project media, no?

test/models/bot/alegre_test.rb
```diff
@@ -129,7 +122,7 @@ def setup
     pm2 = create_project_media project: p
     pm3 = create_project_media project: p
     create_relationship source_id: pm3.id, target_id: pm2.id
-    @bot.add_relationships(pm1, [pm2.id])
+    Bot::Alegre.add_relationships(pm1, [pm2.id])
     r = Relationship.last
     assert_equal pm1, r.target
     assert_equal pm3, r.source
```
caiosba: Some tests still need to be implemented... the following lines of `app/models/bot/alegre.rb` are not covered: 7-8, 55-57, 61, 130-131

app/models/bot/alegre.rb
caiosba: If we send the full `context`, with both `team_id` and `project_media_id`, it means that on the Alegre side it will try to match the whole context and thus always return only the same project media, no?

app/models/bot/alegre.rb
```diff
@@ -96,32 +96,40 @@ def add_relationships(pm, pm_ids)
...
+
+    # Add image to similarity database.
+    self.request_api('post', '/image/similarity', {
+      url: pm.media.file.file.public_url,
+      context: self.get_context(pm)
+    })
   end
+
+  def self.request_api(method, path, params = {})
+    uri = URI(CONFIG['alegre_host'] + path)
```
caiosba: Locally I had to change this line to `uri = URI(CONFIG['alegre_host'] + path + '/')` for it to work (e.g., I had to append a `/` to the end of the request URI). Without that, Alegre was always returning HTTP code 308, with a redirect that was not followed.

test/models/bot/alegre_test.rb
caiosba: Some tests still need to be implemented... the following lines of `app/models/bot/alegre.rb` are not covered: 7-8, 55-57, 61, 130-131

app/models/bot/alegre.rb
```diff
@@ -96,32 +96,40 @@ def add_relationships(pm, pm_ids)
...
-    end
-    pm_ids = response.dig('results')&.collect{|r| r.dig('context', 'project_media_id')}
+  def self.get_image_similarities(pm)
+    return if pm.report_type != 'uploadedimage'
+
+    # Query for similar images.
+    similar = self.request_api('get', '/image/similarity', {
+      url: pm.media.file.file.public_url,
+      context: self.get_context(pm),
+      threshold: 1
```
caiosba: Not sure if I'm doing something wrong, but with a `threshold` of 1, only identical images are returned.

app/models/bot/alegre.rb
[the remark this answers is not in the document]
infojunkie: God catch! Fixed, but how to test in Alegre bot that we're sending the right parameters to Alegre service? Right now, we mock Alegre service responses.

test/models/bot/alegre_test.rb
infojunkie: Thanks! Where can I view the uncovered lines?

app/models/bot/alegre.rb
infojunkie: Thanks, fixed.

app/models/bot/alegre.rb
infojunkie: As per tests in Alegre service, similar but not identical images don't necessarily differ by only 1 bit in their phash. I've increased this threshold to 5 for now, but eventually we'll expose this to the user for better control.

app/models/bot/alegre.rb
infojunkie: Thanks, fixed.

app/models/bot/alegre.rb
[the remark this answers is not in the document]
caiosba: We need to mock that behavior as well, I guess.

test/models/bot/alegre_test.rb
caiosba: On Code Climate we can see the uncovered files: https://codeclimate.com/repos/58bdc058359261025a0020fa/settings/test_reports/5e575d845d73f11ea80064a0. In order to see the lines, I had to run locally. We need a better way to do this in the parallel Travis build... for example, we can upload the HTML coverage report to S3 as well. There is a Code Climate browser extension that used to show the uncovered lines directly on GitHub, but it's not working at the moment (for me, at least) :(

infojunkie: Ideally, we need a way to display the uncovered lines in the build log. It's all about making it easier for devs to find and act on this info.

caiosba: Agree... but we have two parallel jobs. On the Travis matrix there is a way to run a third job after the two parallel ones are done... we can have the third job to only show the uncovered lines.

app/models/bot/alegre.rb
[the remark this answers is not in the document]
infojunkie: I am concerned about having to keep the mocks updated with the implementation. That's the main reason why I think `AlegreClient` is not practical (in addition to having to an expensive round-trip to GitHub when we update the client and use it in the Check API).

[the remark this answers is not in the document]
caiosba: Sure. Ideally we would have Contract Testing - we used mocked requests on Check API and then on Alegre side we make sure that those mocks are also valid. There is a gem for that: https://github.com/pact-foundation/pact-ruby.

PR #560, opened 2020-03-19
Feature/8065 click request message open conversation in slack

melsawy:


app/graph/types/project_type.rb
```diff
@@ -42,6 +42,12 @@
     }
   end
 
+  connection :smooch_user_fields, -> { DynamicAnnotationFieldType.connection_type } do
```
caiosba: Sawy, it's not a good idea to have GraphQL fields or connections that are specific for some annotation types or field names because it doesn't scale... we can't change the schema every time we want to expose an annotation. Ideally let's try to do something like we already have for project media annotations, e.g., a generic field/connection and we can pass the annotation type / field type as an argument: https://github.com/meedan/check-api/blob/develop/app/graph/mutations/graphql_crud_operations.rb#L367-L375

app/models/project.rb
```diff
@@ -227,6 +227,12 @@ def inactive
     team.inactive
   end
 
+  def smooch_user_fields
```
caiosba: Same thing as I said above: let's try to avoid specific methods for annotation types or field names... we can have a method similar to `ProjectMedia`'s `get_annotations`, e.g., a generic method where we can pass the field name and/or annotation type as parameters.

db/migrate/20171113215700_add_slack_message_channel_field.rb
```diff
@@ -5,6 +5,6 @@ class AddSlackMessageChannelField < ActiveRecord::Migration
   def change
     at = DynamicAnnotation::AnnotationType.where(annotation_type: 'slack_message').last
     ft = DynamicAnnotation::FieldType.where(field_type: 'text').last || create_field_type(field_type: 'text', label: 'Text')
-    create_field_instance annotation_type_object: at, name: 'slack_message_channel', label: 'Slack Message Channel', field_type_object: ft, optional: true
+    create_field_instance annotation_type_object: at, name: 'slack_message_channel', label: 'Slack Message Channel', field_type_object: ft, optional: false
```
caiosba: This mutation was already pushed to `develop`, so changing it won't have effect on existing instances. Please add a new mutation to change the `optional` value for this field instance.

app/graph/types/project_type.rb
melsawy: Removed this code and retrieve the slack channel in the requests logs by cache the value during field creating and get the value given request data if cache not exists based on our discussion

app/models/dynamic_annotation/field.rb
```diff
@@ -62,4 +64,10 @@ def set_field_type
   def set_json_value
     self.value_json = self.value if self.field_type == 'json' && self.respond_to?(:value_json)
   end
+
+  def set_author_slack_channel_url
+    a = self.annotation.load
+    value = a.get_field('smooch_user_data')&.value_json
+    Rails.cache.write("SmoochUserSlackChannelUrl:#{a.annotated_type}:#{a.annotated_id}:#{value['id']}", self.value) unless value.nil?
+  end
```
caiosba: Sawy, for this stuff that is specific for the Smooch Bot, we prefer to extend the class in the Smooch Bot class instead of adding custom logic to the core class... please add that logic to this place: https://github.com/meedan/check-api/blob/develop/app/models/bot/smooch.rb#L100

app/models/version.rb
```diff
@@ -266,6 +266,29 @@ def get_team_id
...
+        .joins("INNER JOIN annotations a ON a.annotation_type= dynamic_annotation_fields.annotation_type")
+        .where("a.annotated_type = ? AND a.annotated_id = ?", 'Project', obj.project_id).last
+        unless smooch_user_data.nil?
+          smooch_user = smooch_user_data.annotation.load
+          smooch_user.get_field_value('smooch_user_slack_channel_url')
+        end
+      end
+    end
+    slack_channel_url
+  end
```
caiosba: Same as above - please define `Version.class_eval` in `app/models/bot/smooch.rb` and add this logic there

app/models/version.rb
```diff
@@ -266,6 +266,29 @@ def get_team_id
...
+  def smooch_user_slack_channel_url
+    object_after = JSON.parse(self.object_after)
+    return unless object_after['field_name'] == 'smooch_data'
+    slack_channel_url = ''
+    data = JSON.parse(object_after['value'])
+    unless data.nil?
+      obj = self.associated
+      key = "SmoochUserSlackChannelUrl:Project:#{obj.project_id}:#{data['authorId']}"
+      slack_channel_url = Rails.cache.fetch(key) do
+        # Retrive URl
```
caiosba: "Retrieve URL"

db/migrate/20200309181011_add_smooch_slack_url_annotation_field.rb
```diff
@@ -0,0 +1,10 @@
+class AddSmoochSlackUrlAnnotationField < ActiveRecord::Migration
+	require 'sample_data'
```
caiosba: Please fix indentation

PR #564, opened 2020-03-27
Config updates to be reviewed

infojunkie:
@caiosba plz review updated example files (including comments)

### DIFF
--- a/config/config.yml.example
+++ b/config/config.yml.example
@@ -1,34 +1,178 @@
 development: &default
-  secret_token: somethingsecret
+  secret_token: 'somethingsecret'
+  app_name: 'Check'
   authorization_header: 'X-Check-Token'
-  twitter_consumer_key: test
-  twitter_consumer_secret: test
-  facebook_app_id: 123456
-  facebook_app_secret: test
-  slack_app_id: 123456
-  slack_app_secret: test
-  pender_url: http://localhost:3005
-  pender_url_private: http://localhost:3005
-  pender_key: test
-  elasticsearch_host: 127.0.0.1
-  elasticsearch_port: 9200
-  elasticsearch_log: false
-  elasticsearch_sync: false
+  checkdesk_base_url: 'http://localhost:3000'
+  checkdesk_base_url_private: 'http://localhost:3000'
   checkdesk_client: 'http://localhost:3333'
-  default_mail: root@localhost
-  checkdesk_base_url: http://localhost:3000
-  checkdesk_base_url_private: http://api:3000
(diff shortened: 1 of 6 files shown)


config/config.yml.example
```diff
@@ -1,107 +1,211 @@
...
-  pusher_cluster:
-  video_file_max_size:
-  uploaded_file_max_size:
-  image_min_width:
-  image_max_width:
-  image_min_height:
-  image_max_height:
+  elasticsearch_log: true
+  elasticsearch_sync: false
+  elasticsearch_cloud_id: 'qa:ZXUtd2VzdC0xLmF3cy5mb3VuZC5pbyQxZDkxYjk4YzZmZjQ0YjA0OGZhNDc2NzI4ZTQ2ZTAzMSRhMjE1ODRmZGVkOTM0MzFiYjkyOWU5OGQ3OWYyMzUxMQ=='
```
caiosba: I don't see this `elasticsearch_cloud_id` configuration being used anywhere... what's it used for?

config/config.yml.example
```diff
@@ -1,107 +1,211 @@
...
+  elasticsearch_user: 'elastic'
+  elasticsearch_password: 'HzzENVbibvxsl4LaykQqqOlW'
```
caiosba: This is the QA password, I don't think we need a password for the local instance, but please confirm with Aaron

config/config.yml.example
```diff
@@ -1,107 +1,211 @@
...
+  google_credentials_path: '/app/config/credentials.json'
+
+  # Smooch (Zendesk Sunshine Conversations) messaging platform gateway
+  #
+  # REQUIRED for Check Message
+  #
+  smooch_twitter_consumer_key: '<SMOOCH CONSUMER KEY>'
+  smooch_twitter_consumer_secret: '<SMOOCH CONSUMER SECRET>'
+  smooch_twitter_tier: 'sandbox'
+  smooch_twitter_env_name: 'checkmessage'
```
caiosba: Just to clarify more, those are the settings of a Twitter app used for Check Message, for the Twitter DMs integration

config/config.yml.example
infojunkie: @huslage any idea ^ ?

config/config.yml.example
infojunkie: are `sandbox` and `checkmessage` hardcoded values or they need to be set on the Twitter or Smooch side as well ?

caiosba: on Twitter side as well

PR #573, opened 2020-04-30
Feature/8195 drop terminal status and resolved required task flag

melsawy:
**Based on ticket #8195 I removed the following:**
1. Task status notification
2. Callback that resolve task when answered
3. Task status update to be unresolved in assignment.rb
4. Callbacks for re-open task when destroy answer
5. Required flag from task mutation
6. Callback for back status to active when add required task
7. Terminal status notification
8. Deadline feature (save & reset) and removed deadline field from ES
9. Terminal attribute from Admin UI for custom statuses
10. Assignment progress

**also I changed the logic for opened and completed tasks to be**
- opened tasks: Tasks that have no answers
- completed tasks: Tasks that have at least one answer

- Review add/update/delete team task against new changes


app/models/team.rb
```diff
@@ -314,11 +314,6 @@ def dynamic_search_fields_json_schema
         schema = Dynamic.send(method, self)
         [schema].flatten.each { |subschema| properties[subschema[:id] || type] = subschema }
       end
-      method = "field_sort_json_schema_type_#{type}"
-      if Dynamic.respond_to?(method)
-        sort = Dynamic.send(method, self)
-        properties[:sort][:properties][sort[:id]] = { type: 'array', title: sort[:label], items: { type: 'string', enum: [sort[:asc_label], sort[:desc_label]] } } if sort
-      end
```
caiosba, line 321: This should be present if we want to sort by a dynamic field... was it used only by the `deadline` field? If so, please comment that block of code instead of removing, and please add a comment to the top like "Uncomment to allow sorting by a dynamic field"

melsawy: Yes it only used by deadline field so I'll back it commented

db/migrate/20200430132539_remove_assignments_progress_cache_and_es_deadline_field.rb
```diff
@@ -0,0 +1,7 @@
+class RemoveAssignmentsProgressCacheAndEsDeadlineField < ActiveRecord::Migration
+  def change
+  	# calling reindex to remove `deadline` field
+    CheckElasticSearchModel.reindex_es_data
```
caiosba, line 4: How long will it take? If too long, please move to a rake task under `lib/tasks/migrate`.

db/migrate/20200430132539_remove_assignments_progress_cache_and_es_deadline_field.rb
```diff
@@ -0,0 +1,7 @@
...
+  	Rails.cache.delete_matched('cache-assignments-progress-*')
```
caiosba, line 5: Please fix the indentation :P

test/models/task_test.rb
```diff
@@ -59,17 +59,11 @@ def setup
...
-  end
-
-  test "should add response to task" do
-    t = create_task
-    assert_equal 'unresolved', t.reload.status
-    at = create_annotation_type annotation_type: 'response'
-    create_field_instance annotation_type_object: at, name: 'response_test'
-    t.response = { annotation_type: 'response', set_fields: { response_test: 'test' }.to_json }.to_json
+    # update status
+    t.status='resolved'
```
caiosba, line 66: We removed the logic related to task status change, but not the task status itself, right?

test/models/task_test.rb
```diff
@@ -59,17 +59,11 @@ def setup
...
     t = create_task status: nil
     assert_equal 'unresolved', t.reload.status
-  end
-
-  test "should add response to task" do
-    t = create_task
-    assert_equal 'unresolved', t.reload.status
-    at = create_annotation_type annotation_type: 'response'
-    create_field_instance annotation_type_object: at, name: 'response_test'
-    t.response = { annotation_type: 'response', set_fields: { response_test: 'test' }.to_json }.to_json
```
caiosba, line 72: I think we should keep this test... and at the end, assert that `t.responses.count > 0`, given that now in many parts of the code you replaced `t.status == 'resolved'` by `t.responses.count > 0`

caiosba: And before that, assert that `t.responses.count == 0`

db/migrate/20200430132539_remove_assignments_progress_cache_and_es_deadline_field.rb
infojunkie: Can't we write an ES query to remove the `deadline` field?

melsawy: no it'll not take long time and it's a zero downtime 
ES query will remove deadline data and keep the field mapping to remove the mapping we need to re-index es data

test/models/task_test.rb
melsawy: right

test/models/task_test.rb
melsawy: Ok

app/models/team_task.rb
```diff
@@ -47,34 +47,36 @@ def type=(value)
...
-  def update_teamwide_tasks_bg(options, projects, keep_resolved_tasks)
+  def update_teamwide_tasks_bg(options, projects, keep_completed_tasks)
     # get project medias for deleted projects
     handle_remove_projects(projects) unless projects.blank?
-    # update tasks with zero answer
-    update_tasks_with_zero_answer(options, keep_resolved_tasks)
-    # handle tasks with answers
-    update_tasks_with_answer if options[:required]
+    # collect updated fields with new values
+    colums = {}
```
infojunkie, line 60: typo: "columns"

melsawy: Fixed

test/models/task_test.rb
melsawy: Done

db/migrate/20200430132539_remove_assignments_progress_cache_and_es_deadline_field.rb
melsawy: Fixed

PR #583, opened 2020-05-26
Feature/8348 unify teamwide and custom tags

amoedoamorim:
• Remove teamwide attribute from TagText model, GraphQL layer and related logic
• Make sure that TagTextis not removed when there is no more related Tag
• Remove: used_tags and get_suggested_tags
• On GraphQL TeamType, remove custom_tags and teamwide_tags and replace them by a single tag_texts connection


app/models/team.rb
```diff
@@ -275,26 +275,10 @@ def custom_permissions(ability = nil)
...
-
   def permissions_info
     YAML.load(ERB.new(File.read("#{Rails.root}/config/permission_info.yml")).result)
   end
 
-  def used_tags
-    self.custom_tags.map(&:text)
-  end
-
-  def get_suggested_tags
```
caiosba, line 294: You can remove that from the `TeamType` too

app/models/team.rb
```diff
@@ -275,26 +275,10 @@ def custom_permissions(ability = nil)
...
-
-  def custom_tags
-    self.tag_texts.where(teamwide: false)
-  end
-
   def permissions_info
     YAML.load(ERB.new(File.read("#{Rails.root}/config/permission_info.yml")).result)
   end
 
-  def used_tags
```
caiosba, line 290: You can remove that from the `TeamType` too

db/schema.rb
```diff
@@ -132,9 +132,9 @@
     t.text     "description"
     t.boolean  "optional",        default: true
     t.text     "settings"
+    t.string   "default_value"
```
infojunkie, line 135: @caiosba any idea why the fields keep jumping up and down seemingly randomly?

db/schema.rb
```diff
@@ -444,9 +444,4 @@
   add_index "versions", ["item_type", "item_id", "whodunnit"], name: "index_versions_on_item_type_and_item_id_and_whodunnit", using: :btree
   add_index "versions", ["team_id"], name: "index_versions_on_team_id", using: :btree
 
-  add_foreign_key "accounts", "teams"
```
infojunkie, line 447: @caiosba any ida why these foreign keys keep popping in and out of existence seemingly randomly?

app/graph/mutations/tag_text_mutations.rb
```diff
@@ -1,12 +1,10 @@
 module TagTextMutations
   create_fields = {
     team_id: '!int',
-    teamwide: 'bool',
```
caiosba, line 4: Also, need to add a migration to remove the `teamwide` column from the database (`docker-compose exec api bundle exec rails g mutation RemoveTeamwideFromTagTexts`)

db/schema.rb
caiosba: no... PG version maybe? I get the same behavior as @amoedoamorim btw

app/models/team.rb
amoedoamorim: Missed this, thanks! Done

app/graph/mutations/tag_text_mutations.rb
amoedoamorim: Cool thanks! Is this a mutation generator? I'll run this and commit the file generated by it?

caiosba: Yeah, it's a generator... it will generate a new mutation, then you edit the mutation by adding this to the `change` method: `remove_column :tag_texts, :teamwide`

caiosba: And then you commit the migration file and the updated `db/schema.rb` (which is updated automatically after you run `docker-compose exec api bundle exec rake db:migrate`)

PR #591, opened 2020-06-17
Ticket #6771 - add order field to task and allow user to order tasks

melsawy:


app/graph/mutations/tasks_order_mutation.rb
```diff
@@ -0,0 +1,14 @@
+TasksOrderMutation = GraphQL::Relay::Mutation.define do
```
infojunkie, line 1: Why this and not the usual `set_fields` to set the order via the regular `UpdateTaskMutation` ? @caiosba

app/models/annotations/task.rb
```diff
@@ -28,6 +28,7 @@ def self.task_types
   field :suggestions_count, Integer
   field :pending_suggestions_count, Integer
   field :team_task_id, Integer
+  field :order, Integer
```
infojunkie, line 31: where is this field stored?

app/graph/mutations/tasks_order_mutation.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  input_field :tasks, !JsonStringType
+
+  return_field :success, types.Boolean
+
+  return_field :errors, JsonStringType
+
+  resolve -> (_root, inputs, _ctx) {
+    errors = Task.order_tasks(inputs[:tasks])
+    { success: true, errors: errors }
```
infojunkie, line 12: `success: !errors`

app/graph/mutations/tasks_order_mutation.rb
```diff
@@ -0,0 +1,14 @@
+TasksOrderMutation = GraphQL::Relay::Mutation.define do
+  name 'TasksOrder'
+
+  input_field :tasks, !JsonStringType
```
infojunkie, line 4: how is the order specified in this mutation?

app/graph/mutations/tasks_order_mutation.rb
infojunkie: oh, I see this was specified in the ticket - sorry

infojunkie:
I think a big piece missing here - and sorry if it was missing from the ticket - is to return the tasks ordered according to `order` in the `ProjectMediaType.tasks` call

app/models/annotations/task.rb
melsawy: In Data column

app/graph/mutations/tasks_order_mutation.rb
melsawy: I did a bulk sort so may be one of the ids have an issue like record not found, user has no ability to update, etc ... so I catch any errors while saving the item in errors key

PR #601, opened 2020-07-22
Feature/8404 new media type sound files

melsawy:


app/models/bot/smooch.rb
```diff
@@ -884,9 +892,17 @@ def self.save_media_message(message, type = 'image')
...
       hash = Digest::MD5.hexdigest(data)
-      filename = type == 'image' ? "#{hash}.jpeg" : "#{hash}.mp4"
+      filename =
+        if type == 'image'
+          "#{hash}.jpeg"
+        elsif type == 'video'
+          "#{hash}.mp4"
+        elsif type == 'audio'
+          "#{hash}.mp3"
+        end
```
caiosba, line 902: This can be simplified by just getting the extension based on the type... the name itself is the same for all.

app/models/bot/smooch.rb
```diff
@@ -766,6 +754,26 @@ def self.save_message(message_json, app_id, author = nil)
...
+           self.save_media_message(message, 'video')
+         when 'audio'
+           self.save_media_message(message, 'audio')
+         when 'file'
+           message['mediaType'] = self.detect_media_type(message)
+           m = message['mediaType'].to_s.match(/^(image|video|audio)\//)
+           m.nil? ? return : self.save_media_message(message, m[1])
+         else
+           return
+         end
```
caiosba, line 773: I suggest we create a `save_message(message)` and get rid of this whole `case..when` block... because this way we can just call `save_message(message)`. Also, in most cases the `case` value is the same as the `type` parameter for `save_media_message`.

PR #604, opened 2020-07-23
Feature/8183 team export

adamhooper:
export_team.rake: export a .sqlite3.lz4 file [#8183]

Rationale:

* Consumes half the space of a .zip full of .csvs
* Correctly copies NULLs, integers and dates.
* Allows querying! (With .zip of .csvs, we'd need to import into sqlite3 within Workbench.)
* lz4 is enough to upload (~60MB for a big Team) and fast enough to decompress (fraction of a second).

I used a new query architecture that lets us reuse WHERE clauses and explicitly lists the tables we copy. (It copies all a table's columns by default.) The query is customizable.

I redacted `token`s and `encrypted_password` from all users [#8505].


lib/tasks/data/export_team.rake
```diff
@@ -1,320 +1,459 @@
...
+        #
+        # DIRECT TO FILE: 30.8s, 349MB (63MB gzipped)
+        # MEMORY AND VACUUM: 26.5s, 327MB (54MB gzipped)
+        #
+        # So VACUUM INTO is 14% faster (or much more, if we ignore Rails init)
+        # and 6% more space-efficient (or 14%, after compression)
+        db.execute("VACUUM INTO '#{filename}'")
       end
+
+      # We compress for uploading to Workbench with LZ4 compression.
```
infojunkie, line 442: No need to mention Workbench here - my understanding is that .sqllite.lz4 is fairly agnostic, right?

adamhooper: I think it's a rare combination. Google finds plenty of filenames ending in ".sqlite3.gz" on the Web. It finds zero matches for ".sqlite3.lz4".

We'd be using `.sqlite3` if size didn't matter. (Size does matter, because Workbench limits the total size of uploaded files per step.) We'd be using `.sqlite3.gz` if the decompression only happened once. (Decompression happens every run, because Workbench wipes all data after every execution.)

If I were reading somebody else's code and it didn't explain that we're uploading to Workbench, I would think ".sqlite.lz4" is outlandish. I

adamhooper:
@infojunkie Is this pull request okay to merge? @huslage and I want to test it, and we can only deploy the `develop` branch.

infojunkie:
In the interest of getting this done, go ahead and merge.

PR #619, opened 2020-08-17
Feature/8412 fieldsets migration

melsawy:


db/migrate/20200813175255_set_team_fieldsets.rb
```diff
@@ -0,0 +1,53 @@
+class SetTeamFieldsets < ActiveRecord::Migration
+  def change
+    started = Time.now.to_i
+    RequestStore.store[:skip_notifications] = true
```
caiosba, line 4: Please set `skip_rules` too

db/migrate/20200813175255_set_team_fieldsets.rb
```diff
@@ -0,0 +1,53 @@
...
+        plural: 'tasks'
+      }.with_indifferent_access,
+      {
+        identifier: 'metadata',
+        singular: 'metadata',
+        plural: 'metadata'
+      }.with_indifferent_access
+    ]
+    # Set fieldsets setting to be the default for all teams
+    Team.find_each do |t|
```
caiosba, line 18: Please add a `print '.'` here to make debugging easier

db/migrate/20200813175255_set_team_fieldsets.rb
```diff
@@ -0,0 +1,53 @@
...
+      settings = t.settings || {}
+      settings[:fieldsets] = fieldsets
+      t.update_column(:settings, settings)
+    end
```
caiosba, line 22: Please add a `puts "Teams updated"`, or something like that here

db/migrate/20200813175255_set_team_fieldsets.rb
```diff
@@ -0,0 +1,53 @@
...
+        new_tasks = []
+        tasks.each do |t|
+          print "."
+          t.data["fieldset"] = 'tasks'
+          new_tasks <<  t
+        end
+        # Delete existing task before import new tasks
+        Task.where(id: tasks.map(&:id)).delete_all
+        # Import new tasks with old ids to keep versions
+        Task.import(new_tasks, recursive: false, validate: false)
```
caiosba, line 42: Did you confirm that `id`s are maintained?

melsawy: Yes Caio and revisions still there and appear on activity tab

db/migrate/20200813175255_set_team_fieldsets.rb
```diff
@@ -0,0 +1,53 @@
...
+      remove_index :annotations, name: "index_annotation_type_order"
+      Task.where(annotation_type: 'task').find_in_batches(:batch_size => 10000) do |tasks|
+        new_tasks = []
+        tasks.each do |t|
+          print "."
+          t.data["fieldset"] = 'tasks'
+          new_tasks <<  t
+        end
+        # Delete existing task before import new tasks
+        Task.where(id: tasks.map(&:id)).delete_all
```
caiosba, line 40: This operation looks risky if something goes wrong between deleting and importing... did you try to force an exception between those lines (for example, add a `raise` to line 41) and see if the deleted tasks are reverted?

db/migrate/20200813175255_set_team_fieldsets.rb
melsawy: Done

db/migrate/20200813175255_set_team_fieldsets.rb
melsawy: Done

db/migrate/20200813175255_set_team_fieldsets.rb
melsawy: Done

db/migrate/20200813175255_set_team_fieldsets.rb
```diff
@@ -0,0 +1,53 @@
...
+      }.with_indifferent_access
+    ]
+    # Set fieldsets setting to be the default for all teams
+    Team.find_each do |t|
+      settings = t.settings || {}
+      settings[:fieldsets] = fieldsets
+      t.update_column(:settings, settings)
+    end
+    # Set fieldset = "tasks" for all TeamTasks
+    TeamTask.update_all(fieldset: 'tasks')
```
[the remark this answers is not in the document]
melsawy: Done

db/migrate/20200813175255_set_team_fieldsets.rb
melsawy: Added a `raise` based on the return of `import` action

caiosba: Thanks! But the real question is: when `raise` runs and the whole migration is interrupted... the deleted tasks on line 44 are reverted? Or they are gone forever?

melsawy: They gone forever and print those tasks on console so we can take the action manually

caiosba: Hummm... still risky... let's discuss more on the call

melsawy: I can keep the old tasks and try the import with option `on_duplicate_key_update` but I think it'll take long time

caiosba: It's worth trying, at least to be sure of how long it takes.

melsawy: Yes I am trying now

melsawy: with `on_duplicate_key_update` options the migration took `62m` and I tested with/without indexed and the difference is `1m` to I ignore to delete/create indexes

caiosba: awesome Sawy, in that case, let's keep the `on_duplicate_key_update`

melsawy: OK

PR #627, opened 2020-09-03
Feature/8631 store all user messages from smooch bot

melsawy:


app/models/bot/smooch.rb
```diff
@@ -1035,19 +919,20 @@ def self.timeout_smooch_slack_human_conversation(uid, time)
...
-  def self.timeout_smooch_menu(uid, time)
+  def self.timeout_smooch_menu(time, message, app_id)
+    uid = message['authorId']
     stored_time = Rails.cache.read("smooch:last_message_from_user:#{uid}").to_i
     return if stored_time > time
     sm = CheckStateMachine.new(uid)
     unless sm.state.value == 'human_mode'
-      sm.reset
-      self.clear_user_bundled_messages(uid) unless Rails.env.test?
+      self.delay_for(1.seconds, { queue: 'smooch', retry: false }).bundle_messages(message['authorId'], message['_id'], app_id, 'timeout_requests')
```
caiosba, line 935: shouldn't we pass the `team` as the last attribute here? the `smooch` request annotation should have the `team` as the `annotated` for this case, right?

caiosba (changes requested):
Please check my inline comment in the commit.

app/models/bot/smooch.rb
melsawy: We already fetch the team in `save_message` so I used it in case I save a `smooch` for timeout requests
https://github.com/meedan/check-api/blob/feature/8631-store-all-user-messages-from-smooch-bot/app/models/concerns/smooch_messages.rb#L110

caiosba: Ok...

caiosba (approved):

PR #632, opened 2020-09-15
Feature/8579 slack settings for a list

melsawy:


app/models/project_media_project.rb
```diff
@@ -180,6 +181,16 @@ def self.bulk_update(input_ids, updates, team)
     { team: team, project: project, project_was: project_was, check_search_project_was: project_was&.check_search_project, ids: pm_graphql_ids }
   end
 
+  def slack_channel(event)
+    slack_events = self.project.setting(:slack_events)
```
caiosba, line 185: Will this always return an array? The line below assumes so.

app/models/project_media_project.rb
```diff
@@ -180,6 +181,16 @@ def self.bulk_update(input_ids, updates, team)
...
+    selected_event = slack_events.select{|i| i[:event] == event }.last unless slack_events.nil?
+    selected_event.blank? ? nil : selected_event[:slack_channel]
```
caiosba, line 187: It's good to guarantee that the hash can use strings instead of symbols for the key names (like `:event` and `:slack_channel`). An idea I like is to use the method `with_indifferent_access`.

caiosba (changes requested):

app/models/project_media_project.rb
melsawy: Done and used `with_indifferent_access`

app/models/project_media_project.rb
melsawy: Yes and I should validate the slack settings to be sure it's an array before saving

PR #636, opened 2020-09-25
Feature/8266 upgrade elasticsearch

melsawy:


lib/sample_data.rb
```diff
@@ -658,13 +658,18 @@ def get_es_id(obj)
...
-    { annotated: create_valid_media, context: create_project }.merge(options).each do |key, value|
-      m.send("#{key}=", value) if m.respond_to?("#{key}=")
+    ms = ElasticItem.new
+    pm = options[:project_media] || create_project_media
+    options[:id] = get_es_id(pm)
+    options[:annotated_id] = pm.id
+    options[:annotated_type] = pm.class.name
+    ms = ElasticItem.new
+    options.each do |key, value|
+      ms.attributes[key] = value
```
caiosba, line 668: Why we are instantiating `ElasticItem` twice?

caiosba (approved):
Sawy, apart from my comment in the PR, I have two other questions:
1) Did you test if the changes work with an existing index (e.g., can we find items that were indexed by the old version of the code)?
2) Please create a branch on Check Web with the same name as this branch to be sure that it still works from the client side

caiosba:
Sawy, apart from my comment in the PR, I have two other questions:

Did you test if the changes work with an existing index (e.g., can we find items that were indexed by the old version of the code)?
Please create a branch on Check Web with the same name as this branch to be sure that it still works from the client side

lib/sample_data.rb
melsawy: Fixed

melsawy:
Tested changes against existing data and run `check-web` tests against this branch both cases work well

PR #652, opened 2020-10-20
Update graphql_controller_4_test.rb

tamjid1971:


danielafeitosa:
Hi @tamjid1971!
Did you try to run this test? CheckAPI doesn't have a query to return books...
I'm afraid you created this PR in the wrong repository :disappointed:

PR #662, opened 2020-11-05
Bugfix 33 locale

sonoransun:
Set the locale for container runtimes. This resolves sysops ticket #33 .


sonoransun:
Just checked to confirm this in QA, and it turns out the language specific locales are not installed by default. Instead of installing language locales, which consume a non-trivial amount of space, I will use `C.UTF-8` which still supports UTF-8 character encoding, as requested.

The current error regarding the missing locales is shown as "bash: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)"

PR #664, opened 2020-11-07
Set locale to C.UTF-8 rather than en_US.UTF-8.

sonoransun:
Set UTF-8 capable locale. Avoid installing region specific locales to conserve storage.


sonoransun:
This has been verified in QA. If you connect to a Check API docker container, the following will be set:
```
checkdeploy@ip-10-0-11-88:/app/current$ env | grep LANG
LANGUAGE=C.UTF-8
LANG=C.UTF-8
```

We may want to set this locale for other containers.

caiosba:
Thanks @sonoransun ! I can finally type Portuguese accented words in Rails console :)

```
$ bundle exec rails c p
> x = "Maçã"
=> "Maçã"
```

> We may want to set this locale for other containers.

Yes, I agree!

PR #677, opened 2020-12-09
Feature/8878 hide blank in all items

amoedoamorim:


lib/check_search.rb
```diff
@@ -10,7 +10,7 @@ def initialize(options)
     @options['sort'] ||= (smooch_bot_installed ? 'last_seen' : 'recent_added')
     @options['sort_type'] ||= 'desc'
     # set show options
-    @options['show'] ||= MEDIA_TYPES
+    @options['show'] ||= (@options['projects'] ? MEDIA_TYPES : (MEDIA_TYPES - ['blank'])) # FIXME: Refactor CheckSearch to avoid special cases
```
caiosba, line 13: @amoedoamorim in order to avoid the CC issue and to keep it more organized, I suggest to put that logic in a separate method `default_types_to_show`, or something like that, and put the comment there too (and maybe detail more in the comment)

caiosba (changes requested):
Please see my inline comments.

lib/check_search.rb
amoedoamorim: Alright! Will fix that tomorrow morning

melsawy (approved):

PR #681, opened 2020-12-22
updates to remove deep duplication for teams

DGaffney:


app/graph/mutations/duplicate_team_mutation.rb
```diff
@@ -0,0 +1,19 @@
...
+  input_field :custom_name, types.String
+
+  return_field :team, TeamType
+
+  resolve -> (_root, inputs, ctx) {
+    _type_name, id = CheckGraphql.decode_id(inputs['team_id'])
+    user = User.current
+    ability = Ability.new(user)
+    team = GraphqlCrudOperations.load_if_can(Team, id, ctx)
+    raise I18n.t('team_clone.user_not_authorized') if ability.cannot?(:import_spreadsheet, team)
```
caiosba, line 15: It's better to create a new permission instead of reusing `import_spreadsheet`

app/models/concerns/team_duplication.rb
```diff
@@ -7,215 +7,109 @@ module TeamDuplication
...
-          self.copy_annotations
-          self.update_project_medias
-          self.update_project_media_projects
-          self.copy_relationships(@mapping[:"ProjectMedia"])
-          self.copy_versions(@mapping[:"Version"])
-          self.update_cloned_versions(@cloned_versions)
-          self.create_copy_version(@mapping[:ProjectMedia], user)
-          Version.set_callback(:create, :after, :increment_project_association_annotations_count)
-          team
+    def self.duplicate(t, custom_slug=nil, custom_name=nil)
```
caiosba, line 10: Please wrap `=` in spaces, just to keep the style (I know we should have a linter...)... so it would like this: `custom_slug = nil, custom_name = nil`

app/models/concerns/team_duplication.rb
```diff
@@ -7,215 +7,109 @@ module TeamDuplication
...
-        field.save(validate: false)
-        self.set_mapping(f, field)
+    def self.process_team_users(t, team, processed_user_ids)
+      t.team_users.each do |tu|
+        next if processed_user_ids.include?(tu.user_id)
+        new_tu = TeamUser.new(tu.attributes.select{|k,_| k!="id"})
+        new_tu.team = team
+        new_tu.save!
       end
     end
```
caiosba, line 84: If the idea of this method is to migrate the `TeamUser`s, we don't need to - only the `TeamBotInstallation`s should be duplicated. And the user that is performing the duplication should be the owner of the new team.

app/models/concerns/team_duplication.rb
```diff
@@ -7,215 +7,109 @@ module TeamDuplication
...
+          self.alter_copy_by_type(original, copy)
         end
-      rescue StandardError => e
-        self.log_error(e, t)
-        nil
+        processed_user_ids = self.process_team_bot_installations(t, team)
+        self.process_team_users(t, team, processed_user_ids)
+        team = self.update_team_rules(team)
+        Team.current = team
+        team.save(validate: false)
```
caiosba, line 32: Please replace this by `team.save!` - we want to raise an exception if the new team is not valid (for example, slug exists or contains invalid characters).

app/models/concerns/team_duplication.rb
```diff
@@ -7,215 +7,109 @@ module TeamDuplication
...
+        copy.team_id = @team_id if !@team_id.nil?
       end
-    end
-
-    def self.versions_log_mapping(original, copy)
-      if original.respond_to?(:get_versions_log)
-        original.get_versions_log.find_each do |log|
-          self.set_mapping(log, copy)
-        end
+      copy.save(validate: false)
```
caiosba, line 58: Please replace this by `copy.save!` - we want to raise an exception if the copy is not valid

app/models/project.rb
```diff
@@ -200,9 +200,13 @@ def search
     CheckSearch.new({ 'parent' => { 'type' => 'project', 'id' => self.id }, 'projects' => [self.id] }.to_json)
   end
 
-  def generate_token
+  def generate_token(force=false)
```
caiosba, line 203: Please add spaces around `=` :D

app/models/concerns/team_duplication.rb
```diff
@@ -7,215 +7,109 @@ module TeamDuplication
...
-      associations.each_pair do |field, method|
-        unless changes[field].blank?
-          changes[field].map! do |a|
-            c = @mapping.dig(log.send(method).to_sym, a)
-            c ? c.id : a
+    def self.store_clones(team)
+      @clones.each do |clone|
+        if !clone[:original].is_a?(Team)
+          if clone[:clone].respond_to?(:team_id) && clone[:clone].team_id.nil?
+            clone[:clone].team_id = team.id
```
caiosba, line 106: This line is missing a test.

caiosba (changes requested):

app/models/concerns/team_duplication.rb
DGaffney: After more review I've determined that this is actually not a line we need - I don't actually think we should see a nil id here? I'm going to run a test with this out altogether.

caiosba: You can also replace the condition `&& clone[:clone].team_id.nil?` by a `clone[:clone].team_id ||= team.id` in the line below.

caiosba (approved):

PR #690, opened 2021-01-12
Upload to Workbench TUS URLs

adamhooper:
Also, retry uploads on failure and skip uploads when the retries fail.

**TO DEPLOY**, be sure to change the CSV to match the new schema. Instead
of `workflow_id` and `step_id`, there is now just one column,
`workbench_step_url`. To migrate your CSV, apply a formula:

```workbench_step_url = "https://app.workbenchdata.com/api/v1/workflows/" + workflow_id + "/steps/" + step_id + "/files"```

For instance, now a valid CSV looks like:

```
team_slug,workbench_step_url,api_token
my-team-1,https://app.workbenchdata.com/api/v1/workflows/123456/steps/step-abcdef/files,SoMe-Tok3n
my-team-1,https://app.workbenchdata.com/api/v1/workflows/123456/steps/step-bcdefg/files,4n0th3r-70k_n
```

### DIFF
--- a/lib/tasks/data/workbench_upload.rb
+++ b/lib/tasks/data/workbench_upload.rb
@@ -1,13 +1,27 @@
 # :nocov:
-require 'aws-sdk-s3'
 require 'typhoeus'
 
+N_RETRIES = 3
+
 module WorkbenchUpload
-  def self.upload_file_to_workbench(workflow_id:, step_id:, api_token:, path:, filename:)
-    s3_config = get_s3_config_from_workbench(workflow_id, step_id, api_token)
-    puts "Uploading to s3://#{s3_config['bucket']}/#{s3_config['key']}"
-    upload_file_to_s3(path, s3_config)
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: lib/tasks/data/workbench_upload.rb [hunk 1]
```diff
@@ -1,69 +1,69 @@
 # :nocov:
-require 'aws-sdk-s3'
```
huslage, line 2: Is there a reason we are using Typhoeus directly instead of using the S3 SDK? 

Also, what about uploading directly to google instead of using MinIO?

adamhooper:
@huslage We're not uploading to S3 any more.

Workbench doesn't actually use S3: it uses GCS. And we couldn't emulate S3 correctly. So Workbench isn't pretending any more.

huslage:
Got it.

huslage:
LGTM

PR #719, opened 2021-02-22
CHECK-277 Add vectorization rake task

DGaffney:


app/models/bot/alegre.rb
```diff
@@ -36,6 +36,14 @@ def delete_analysis_from_similarity_index
...
   end  
 
+  def self.default_model
+    'xlm-r-bert-base-nli-stsb-mean-tokens'
+  end
+
+  def self.indian_model
+    'indian-sbert'
+  end
+
```
caiosba, line 46: What about ElasticSearch? Are we dropping it completely?

caiosba: Also, it could be just a constant instead of a method, right?

lib/tasks/set_language_model_for_alegre_team_bot_installation.rake
```diff
@@ -0,0 +1,13 @@
...
+  # bundle exec rake check:fix_untitled_sources['id:1&name:sname']
+  # bundle exec rake check:fix_untitled_sources to fix sources with name == 'Untitled'
+  desc "Set Alegre Team Bot Installation model in use for generating string similarities (usage: bundle exec rake check:set_language_model_for_alegre_team_bot_installation['team_id:1','model_name:indian-sbert'])"
+  task set_language_model_for_alegre_team_bot_installation: :environment do |_t, args|
+    team = Team.find(args.team_id)
+    tbi = team.team_bot_installations.select{|x| x.user.login == "alegre"}.first
+    settings = tbi.json_settings
+    settings["alegre_model_in_use"] = args.model_name
+    tbi.json_settings = settings
+    tbi.save!
```
caiosba, line 11: We have a `lib/check_settings.rb` library to abstract this kind of operation, so it can simply be:
```
tbi.set_alegre_model_in_use = args.model_name
tbi.save!
```

lib/tasks/set_language_model_for_alegre_team_bot_installation.rake
```diff
@@ -0,0 +1,13 @@
+namespace :check do
+  # bundle exec rake check:fix_untitled_sources['id:1&name:sname']
+  # bundle exec rake check:fix_untitled_sources to fix sources with name == 'Untitled'
```
caiosba, line 3: Looks like these comments have nothing to do with this rake task hehe

caiosba (changes requested):
Please see my comments in the commit.

app/models/bot/alegre.rb
DGaffney: We are dropping elasticsearch as a default - it can obviously be passed in - I'll change this to a constant.

PR #729, opened 2021-03-10
Meedan 327 add condition

DGaffney:


app/models/bot/alegre.rb
```diff
@@ -27,7 +27,9 @@ def self.delete_analysis_from_similarity_index(pm_id)
     private
 
     def can_be_sent_to_index?
-      ['content', 'title'].include?(self.field_name) && self.annotation.annotation_type == 'verification_status'
+      first = ['content', 'title'].include?(self.field_name) && self.annotation.annotation_type == 'verification_status'
+      second = self.annotation && self.annotation.annotated && self.annotation.annotated.team && Bot::Alegre.team_has_alegre_bot_installed?(self.annotation.annotated.team.id)
+      return first && second
```
caiosba, line 32: Can we simplify it please?
```
['content', 'title'].include?(self.field_name) && self.annotation.annotation_type == 'verification_status' && Bot::Alegre.team_has_alegre_bot_installed?(self.annotation&.annotated&.team&.id&.to_i)
```
That should be equivalent to the above :)

PR #733, opened 2021-03-18
CHECK-375: fix errbit error

melsawy:


DGaffney (commented):
These look good to me!

PR #771, opened 2021-05-05
Check-436: Run tests without configurator

danielevalverde:


app/models/concerns/team_import.rb
```diff
@@ -4,33 +4,33 @@
 module TeamImport
```
caiosba, line 4: Why you commented just some lines and not the whole file? Are the uncommented lines used for anything else that is not related to the `TeamImport` feature?

config/config.yml.example
```diff
@@ -68,20 +68,20 @@ development: &default
   # Generate a key with SecureRandom to enable two-factor authentication (2FA) `SecureRandom.hex(20)`
   # If blank the QRCode won't be generated and the 2FA can't be enabled
   #
-  two_factor_key: # '<Google two factor key>'
+  two_factor_key: "a2ab0425f2cfa4243110e343bc865653aea941b91819e754c9a82a0115266add" # '<Google two factor key>'
```
caiosba, line 71: This is just a random string not used anywhere else, right? :)

danielevalverde: Yes, it was generated using `SecureRandom`

PR #774, opened 2021-05-11
CHECK-77: Folders, collections and dynamic lists

caiosba:


caiosba:
@melsawy I appreaciated your review in the PR - it's also important for you to be aware of these changes to the API and its data model.

app/models/project_group.rb
```diff
@@ -0,0 +1,14 @@
+class ProjectGroup < ActiveRecord::Base
+  validates_presence_of :title, :team_id
+
```
melsawy, line 3: @caiosba I think we should also add a validation for uniqueness title per team

melsawy (approved):
LGTM Dr Caio, just added a minor validation for ProjectGroup

app/models/project_group.rb
caiosba: Thanks Sawy, done!

PR #784, opened 2021-05-19
CHECK-569 restrict matches after identification to matches that are of same media type

DGaffney:


caiosba:
@DGaffney please add a unit test for this case.

PR #797, opened 2021-06-01
run tests with GitHub Actions

dmou:
and delete travis job


caiosba:
@dmou are we able to see the output of the `docker-compose exec` commands? By reading the current output I'm not sure if all tests really passed or if they didn't even run :)

What I'm suspecting is that it was too fast.

dmou:
you're right, I was too hasty 😞
it worked locally 😝 will reopen this PR when it's actually done 👀

PR #808, opened 2021-06-21
CHECK-675: get similarity for too short text

melsawy:


app/models/bot/alegre.rb
```diff
@@ -119,49 +114,47 @@ def self.merge_suggested_and_confirmed(suggested_or_confirmed, confirmed, pm)
...
+      confirmed = self.get_items_with_similarity(type, pm, self.get_threshold_for_query(type, pm, true))
       self.merge_suggested_and_confirmed(suggested_or_confirmed, confirmed, pm)
     else
       {}
     end
   end
 
+  def self.get_items_with_similarity(type, pm, threshold)
+    if type == 'text'
+      self.get_merged_items_with_similar_text(pm, threshold)
```
DGaffney, line 147: I think we should probably also `reject_same_case` on these as well, no @caiosba ?

app/models/bot/alegre.rb
```diff
@@ -119,49 +114,47 @@ def self.merge_suggested_and_confirmed(suggested_or_confirmed, confirmed, pm)
...
+      self.reject_same_case(self.get_items_with_similar_media(self.media_file_url(pm), threshold, pm.team_id, "/#{type}/similarity/"), pm)
+    end
+  end
+
   def self.get_merged_items_with_similar_text(pm, threshold)
     by_title = self.get_items_with_similar_title(pm, threshold)
     by_description = self.get_items_with_similar_description(pm, threshold)
     Hash[(by_title.keys|by_description.keys).collect do |pmid|
-      [pmid, [by_title[pmid].to_f, by_description[pmid].to_f].sort.last]
+      [pmid, [by_title[pmid].to_f, by_description[pmid].to_f].max]
```
DGaffney, line 157: yep

app/models/bot/alegre.rb
```diff
@@ -293,46 +285,29 @@ def self.send_to_text_similarity_index(pm, field, text, doc_id, model=nil)
...
-
-  def self.send_to_video_similarity_index_package(pm)
-    {
-      doc_id: self.item_doc_id(pm, 'video'),
-      url: self.media_file_url(pm),
-      context: {
-        team_id: pm.team_id,
-        project_media_id: pm.id,
-        has_custom_id: true
+  def self.send_to_media_similarity_index(pm)
```
DGaffney, line 288: Good Call

app/models/bot/alegre.rb
caiosba: What you mean, @DGaffney ? Sorry, I didn't understand...

melsawy: me to didn't understand,  I just group methods to overcome CC and still using same existing code

app/models/bot/alegre.rb
```diff
@@ -364,23 +339,31 @@ def self.split_text(text)
     text.split(/\s/)
   end
 
-  def self.get_items_with_similar_title(pm, threshold, text_length_threshold=self.similarity_text_length_threshold)
-    self.split_text(pm.title.to_s).length > text_length_threshold ? self.get_merged_similar_items(pm, threshold, ['original_title', 'analysis_title'], pm.title) : {}
+  def self.get_items_with_similar_title(pm, threshold)
```
DGaffney, line 342: Is this part of this ticket to remove the idea of a `similarity_text_length_threshold`? If so, this is fine, otherwise, this is definitely breaking that old logic

melsawy: yes @DGaffney I removed  `similarity_text_length_threshold` condition as short text will be added as suggestion

app/models/bot/alegre.rb
DGaffney: We have a function `reject_same_case` that removes from the list of similar matches the ProjectMedia we are currently matching to other cases (e.g. don't return PM ID of 1 as a match for PM ID of 1). We use it on line 149 here, but not on these text cases - does that help clarify?

DGaffney (commented):
Looks good to me other than the things I noted, so long as tests pass

app/models/bot/alegre.rb
caiosba: Ah, thanks... @melsawy please confirm we're keeping that behavior.

melsawy: I see that we already using the behavior for `reject_sama_case` for text cases https://github.com/meedan/check-api/blob/feature/CHECK-675-add-too-short-text-as-suggestions/app/models/bot/alegre.rb#L410

PR #897, opened 2021-11-11
Feature/check 87 sensitive content

melsawy:


app/models/concerns/team_rules.rb
```diff
@@ -179,6 +180,17 @@ def add_tag(pm, value, _rule_id)
       tag.skip_check_ability = true
       CheckNotification::InfoMessages.send('tagged_by_rule', item_title: pm.title, tag: tag_text.text) if tag.save
     end
+
+    def add_warning_cover(pm, _value, _rule_id)
+      flag = pm.annotations('flag').last.load
+      unless flag.nil?
+        RequestStore.store[:skip_rules] = true
```
caiosba, line 187: We can also skip permission check here, Sawy

caiosba (approved):

app/models/concerns/team_rules.rb
melsawy: Done

PR #940, opened 2022-01-25
CHECK-1207 add model to Relationship

DGaffney:


app/models/bot/alegre.rb
```diff
@@ -493,6 +493,15 @@ def self.add_relationship(pm, pm_id_scores, parent_id)
     end
   end
 
+  def self.get_indexing_model(pm)
+    type = self.get_pm_type(pm)
+    if type == text
```
melsawy, line 498: @DGaffney I think it will be more simpler if you replace `if else` with conditional operator 
i.e type == text ? type : self.indexing_model_to_use(pm)

melsawy, line 498: also I think text should be a string 'text' right?

DGaffney: You're right!

DGaffney: Good catch!

PR #947, opened 2022-01-31
URL Annotation

dariusk:
This creates a new annotation type called `"url"`. The data structure that it expects to store is an array of objects, each containing two strings, `"url"`, and `"title"`. The `title` is optional and may be an empty string. The `title` property should always be there, and if "empty" should be an empty string and never `null`.

```json
[
  { "url": "http://example.com", "title": "Example Web Site" },
  { "url": "https://meedan.com", "title": "Meedan" },
]
```

The formatted output (aka `Task.first_response_value`) is the JSON string of the object above.

Validation checks to see if if each `url` in each object in the array is a valid http or https url. If one or more urls is invalid, it returns an error and specifies each invalid url in the error message.

Tests coming in a future PR.

### DIFF
--- a/app/models/annotations/task.rb
+++ b/app/models/annotations/task.rb
@@ -19,7 +19,7 @@ class Task < ApplicationRecord
 
   field :type
   def self.task_types
-    ['free_text', 'yes_no', 'single_choice', 'multiple_choice', 'geolocation', 'datetime', 'file_upload', 'number']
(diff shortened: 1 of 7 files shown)


caiosba (approved):

caiosba:
Thanks @dariusk ! @melsawy please review :)

melsawy (approved):

PR #960, opened 2022-02-07
Feature/1188 edit annotation fields

amoedoamorim:
Allow editing fields with existing answers


melsawy (commented):
Changes are simple and looks good to me but we still need to add a callback when options changed to delete existing answers with deleted options

PR #1002, opened 2022-03-06
CHECK-1537 - Prefer vector models to Elasticsearch results

computermacgyver:
I earlier committed code for this improvement already, but I think it is better to explicitly compare against the `Bot::Alegre::ELASTICSEARCH_MODEL` constant when ordering results. We don't have any vector model that alphabetically comes before "e", but we could in the future and that would break the previous fix without warning, which is bad. This PR fixes that.


caiosba (approved):

caiosba:
:lgtm:

PR #1082, opened 2022-05-11
Update newsletter cron conversion

dariusk:
The API now accepts any string with an explicit offset in the format `(+|-)hh:mm`.

All time zones provided clientside are guaranteed to work via the API -- for example, even though Ruby doesn't support PHT, the string that comes in will say `Pacific/Phillippines (GMT+08:00)`, which simply gets converted to `+08:00` so `DateTime` doesn't have to know anything about the 3-letter codes. This means we can have fewer unit tests. I now test for positive and negative offsets, and kept the tests for non-integer offsets, and offsets where the day changes.

CHECK-1139

### DIFF
--- a/app/models/concerns/smooch_newsletter.rb
+++ b/app/models/concerns/smooch_newsletter.rb
@@ -77,8 +77,13 @@ module SmoochNewsletter
 
     def newsletter_cron(newsletter)
       hour = newsletter['smooch_newsletter_time'].to_i
-      timezone = newsletter['smooch_newsletter_timezone'].to_s.upcase
-      # Mapping for timezones not supported by Ruby's DateTime
+      # If an offset is being passed, it's in the new format
+      if newsletter['smooch_newsletter_timezone'].match?(/\W\d\d:\d\d/)
+        timezone = newsletter['smooch_newsletter_timezone'].match(/\W\d\d:\d\d/)
+      else
+        timezone = newsletter['smooch_newsletter_timezone'].to_s.upcase
+      end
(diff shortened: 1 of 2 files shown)


app/models/concerns/smooch_newsletter.rb
```diff
@@ -77,8 +77,13 @@ def newsletter_content_changed?(newsletter, language, team_id)
 
     def newsletter_cron(newsletter)
       hour = newsletter['smooch_newsletter_time'].to_i
-      timezone = newsletter['smooch_newsletter_timezone'].to_s.upcase
-      # Mapping for timezones not supported by Ruby's DateTime
+      # If an offset is being passed, it's in the new format
+      if newsletter['smooch_newsletter_timezone'].match?(/\W\d\d:\d\d/)
+        timezone = newsletter['smooch_newsletter_timezone'].match(/\W\d\d:\d\d/)
+      else
```
dariusk, line 83: @caiosba I'm afraid that if I get rid of the conditional here and only look for the new format, there will be a situation where we try to parse a date with a null/undefined value for the timezone and there will be some kind of error thrown. This is mostly me not knowing how Ruby's `match` and `DateTime.parse` really work in failure cases.

caiosba: OK, makes sense, it doesn't hurt to have a fallback... better be safe.

caiosba: Let's just be sure that the `else` case is also covered in that unit test.

caiosba (approved):

PR #1084, opened 2022-05-13
CHECK-1848 Add new reindexer

DGaffney:


lib/tasks/migrate/20220512132847_reindex_alegre.rake
```diff
@@ -0,0 +1,79 @@
...
+    task update_alegre_stored_project_media: :environment do
+      started = Time.now.to_i
+      running_bucket = []
+      log_errors = []
+      temp_ids = []
+      team_total = BotUser.alegre_user.team_bot_installations.count
+      counter = 0
+      sent_cases = []
+      received_cases = []
+      BotUser.alegre_user.team_bot_installations.find_each do |tb|
```
caiosba, line 13: We can skip teams that don't have text similarity enabled: `next if !tb.team.get_master_similarity_enabled || !tb.team.get_text_similarity_enabled`

lib/tasks/migrate/20220512132847_reindex_alegre.rake
```diff
@@ -0,0 +1,79 @@
...
+                request_doc = Bot::Alegre.send_to_text_similarity_index_package(
+                  pm,
+                  field,
+                  field_value,
+                  Bot::Alegre.item_doc_id(pm, field)
+                ).merge(models: models)
+                request_doc.delete(:model)
+                running_bucket << request_doc
+              end
+            end
```
caiosba, line 38: Can we parallelize this block? One thread for each field? Something like:
```
threads = []
threads << Thread.new {
  # The current block above comes here
}
threads.map(&:join)
```
```

caiosba (approved):

PR #1120, opened 2022-06-10
Get tipline hardcoded content from transifex

amoedoamorim:
This commit adds a rake task for downloading the tipline translatables from a new project `check-tiplines` and new resource `hardcoded-bot-strings`.
Also modifies `smooch_strings.rb` to read from the downloaded content instead of hardcoded hash.

Fixes CHECK-1902


app/models/concerns/smooch_strings.rb
```diff
@@ -1,336 +1,15 @@
...
-          "te": "మీరు ప్రస్తుతం మా వార్తాలేఖకు సభ్యత్వం పొందలేదు.",
-          "ur": "آپ نے فی الحال ہمارے نیوز لیٹر کو سبسکرائب نہیں کیا ہے۔"
-        }
-      }[key.to_sym]
-      language = language.gsub(/[-_].*$/, '').to_sym
-      string ? (string[language] || string[:en]) : string
+      yaml = YAML.load(File.read(File.join(Rails.root, 'config', 'tipline_strings.yml')))
+      language = language.gsub(/[-_].*$/, '')
+      strings = yaml[language] || yaml['en']
+      strings[key]
```
caiosba, line 12: As per the previous code, please fallback to `key` when there is no `strings[key]`: `strings[key] || key`.

Also, please confirm if `YAML.load` loads the keys as strings (e.g., `'en'`) or symbol (e.g., `:en`).

lib/tasks/transifex.rake
```diff
@@ -49,6 +49,28 @@ namespace :transifex do
     end
   end
 
+  # Download translations from Transifex - tipline resource
+
+  task download_tipline: [:environment, :login] do
+    project = Transifex::Project.new('check-tiplines')
+    resource_slugs = project.resources.fetch.select{ |r| r['slug'] =~ /^hardcoded-bot-strings/ }.collect{ |r| r['slug'] }
```
caiosba, line 56: There is only one resource, right? So we could replace this `select` by a `find`.

caiosba (changes requested):
Please see my inline comments - great work overall!

lib/tasks/transifex.rake
```diff
@@ -49,6 +49,28 @@ namespace :transifex do
...
+    langs = project.languages.fetch.collect{ |l| l['language_code'] } # + ['en'] Is + ['en'] necessary if translation coordinator is set?
```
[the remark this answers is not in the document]
amoedoamorim: I just tested this and deleted the commented part. English will be downloaded if there's a translation team set on Transifex. Same behaviour as the other languages. I'm adding myself as Coordinator in each of them.

lib/tasks/transifex.rake
amoedoamorim: Even better, since the resource name is known we can just pass it to the `resource = project.resource(slug)` line

PR #1123, opened 2022-06-15
Downsample Honeycomb events by half. 

sonoransun:
This change reduces Honeycomb events by half as we're approaching monthly limits due to increased usage.


caiosba (approved):

hartsick:
@sonoransun Do you know if this downsampling is still needed? I'm replacing this file in #1223 and am thinking about how and what rate we will need to sample

sonoransun:
> 

It is no longer needed! We upgraded our plan earlier this month. Thanks for pointing this out, I'll clean up the deprecated branch!

PR #1152, opened 2022-07-22
Check 2147 support for m 1 mac dockerized development environment

hartsick:
Updates gems causing issue with M1 macs.

Also updates configuration and tests to account for change in Alegre port.

For more detail see commit messages and https://github.com/meedan/check/pull/69


huslage (approved):
LGTM. I think we might want to parameterize these external services into ENV vars instead of hardcoded values.

hartsick:
yeah, agreed!

PR #1159, opened 2022-07-29
Avoid using Nokogiri precompiled native gem

hartsick:
In the Docker container on M1 Macs, I was running into an issue when
installing Nokogiri that a required library, glibc 2.29, was not
available. This prevented running check-api on my machine, though the image built successfully.
Nokogiri's recommended approach for addressing this is to force
using the Ruby platform gem. For our version of Bundler (<2.0), it
is only possible to do this at the bundler config level (vs gem
level, which is supported in Bundler >=2.3.18).

https://nokogiri.org/tutorials/installing_nokogiri.html#how-can-i-avoid-using-a-precompiled-native-gem

CHECK-2147

For folks I've tagged for review, I'd love thoughts on:
* If there's a way we could address this at the container level (like making sure glibc is available)
* Any concerns about using ruby platform gem / checking for any problems building and running on non-M1 machine

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,9 +40,9 @@ RUN python -m pip install -U setuptools wheel
 WORKDIR /app
 COPY Gemfile /app/Gemfile
 COPY Gemfile.lock /app/Gemfile.lock
-RUN echo "gem: --no-rdoc --no-ri" > ~/.gemrc \
-    gem install bundler -v "< 2.0" \
-    && bundle install --jobs 20 --retry 5
+RUN echo "gem: --no-rdoc --no-ri" > ~/.gemrc && gem install bundler -v "< 2.0"
+RUN bundle config force_ruby_platform true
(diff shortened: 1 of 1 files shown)


dariusk:
Hmm, is this going to result in sitting through Nokogiri's legendary compile times every time we update the bundle?

caiosba (approved):

caiosba:
Changes look good... we just use Nokogiri in one place in Check API, which is to turn a report template into an actual report. So, if it seems to be too problematic (I mean, more than Nokogiri already is) we can also replace it by another gem that does what we need.

hartsick:
@dariusk I haven't really been able to tell the difference on my machine, but it's definitely something I was concerned about and would want to hear other folks' experience with. Looking at the last few CI runs with and without the changes there doesn't seem to be a meaningful and consistent difference in build times.

dariusk:
@hartsick Good to know!

PR #1174, opened 2022-08-13
Check-2248 Do not count suggested and then deleted suggestions as detached

computermacgyver:
Fix what counts as rejected or detached items
    
    When a suggestion is confirmed by a user and then deleted, it should not count as a "manually detached match"
    Such cases should could as "rejected suggestions". These cases are distinguished by whether the 'confirmed_by'
    field is NULL or not.


computermacgyver:
This has been manually merged. Closing the pull request.

PR #1223, opened 2022-09-28
Begin using Open Telemetry to send traces and metrics to Honeycomb

hartsick:
Honeycomb is encouraging new users to use the Open Telemetry instrumentation rather than the proprietary Beeline library for sending information to their service. This PR updates our code base to use that instrumentation to send to Honeycomb, following the pattern recently set in Pender. It preserves the existing behavior of adding attributes (for GraphQL queries) to traces and allowing sampling, though this is turned off by default.

Some things this PR does:
* Adds Open Telemetry gems and uses them to report traces to Honeycomb
* Tries to create a guide-rails type config for Open Telemetry. Most of the configuration for the library is done by environment variables, which can easily conflict. The OpenTelemetryConfig class attempts to help us with that, and allows us to unit test the config.
* Sets up a OpenTelemetryTestConfig that is used in test environments. By default, this dumps any spans reported, but it can optionally be configured to record, which preserves the spans for test purposes
    * This configuration also initializes the same instrumentation libraries as development and deployed environments, to try to catch any issues arising from that configuration in our automated tests.
* Allows for sampling of traces sent to Honeycomb
* Allows for configuration of Honeycomb log level in config. By default (in config.yml.example), log level is info in development/deployed envs and error in tests to reduce noise
* Creates a convenience wrapper class for us to put any methods we use with Open Telemetry gems. This is both for our convenience / to reduce possible misconfiguration, and to aid in ease of unit testing.

Steps taken to test locally can be found here: https://meedan.atlassian.net/browse/CHECK-2484?focusedCommentId=23601 

Some open questions / places I'd particularly appreciate feedback:
* Where should usage documentation go? I had added [docs to Pender](https://github.com/meedan/pender/blob/develop/README.md#observability), but since this configuration now spans across and will impact multiple services I'm thinking we move it into Confluence (I'm drafting a page to explain our Honeycomb setup; that seems a natural place for it)
* What seems confusing that we could clarify, either through docs or code streamlining?


lib/check_open_telemetry_config.rb
```diff
@@ -0,0 +1,83 @@
+require 'opentelemetry/sdk'
+require 'opentelemetry/exporter/otlp'
+require 'opentelemetry/instrumentation/all'
+
+module Check
```
hartsick, line 5: This logic is kind of complicated, but there's a lot of room for error in the way that Open Telemetry uses environment variables for configuration. I'm trying to provide some guiderails around this, and also do so in a way that lets us unit test and control the initialization.

test/lib/tracing_service_test.rb
```diff
@@ -0,0 +1,14 @@
+require 'test_helper'
+require 'minitest/autorun'
+require 'mocha/mini_test'
```
hartsick, line 3: this is impacted by #1249 , so just need to keep an eye on this depending on what's merged first

production/config.yml.dockerfile
```diff
@@ -158,12 +158,16 @@ development: &default
   #
   google_analytics_code: # '<GOOGLE ANAYTICS CODE>'
 
-  # Honeycomb analytics service https://www.honeycomb.io
+  # Open Telemetry configuration, for reporting to Honeycomb
```
hartsick, line 161: @caiosba could you help me understand what this file does and what its relationship to config.yml.example should be? I wasn't sure if we need to keep them in lockstep or what.

caiosba: I didn't remember that this file even existed... I don't think it's used in our live deployments anymore, but @sonoransun can confirm.

hartsick: I see it's being called in `production/Dockerfile`. but it seems to be [deleted in the same line it's created](https://github.com/meedan/check-api/blob/develop/production/Dockerfile#L68)? Also realize I'm not sure how that Dockerfile is being used

sonoransun: This is using a default config to generate the precompiled assets, E.g. it's all in service to this command: `bundle exec rake assets:precompile`. Note that these steps are done when building the image, but the final entrypoint command, `ENTRYPOINT ["/opt/bin/start.sh"]`, isn't invoked until runtime. At this point, proper QA or Live configs will be generated as expected.

I think it would be better to copy the default config into place, removing this config entirely. E.g. use this line instead: `RUN cp config/config.yml.example config/config.yml && \`

Thanks for catching these @hartsick , it's easy to overlook them otherwise!

hartsick: awesome! thanks for that context, @sonoransun ! I just wanted to make sure I wasn't missing any configs we needed to set - think I found this initially when cmd + f'ing

PR #1245, opened 2022-10-11
Fixes for "add item to imported report" feature

caiosba:
Still work in progress.

Fixes CHECK-2543.

### DIFF
--- a/app/models/project_media.rb
+++ b/app/models/project_media.rb
@@ -244,32 +244,36 @@ class ProjectMedia < ApplicationRecord
   end
 
   def replace_by(new_pm)
+    return if self.id == new_pm.id
     if self.team_id != new_pm.team_id
       raise I18n.t(:replace_by_media_in_the_same_team)
     elsif self.media.media_type != 'blank'
       raise I18n.t(:replace_blank_media_only)
     else
-      # Save the new item
-      new_pm.updated_at = Time.now
-      new_pm.skip_check_ability = true
-      new_pm.channel = { main: CheckChannels::ChannelCodes::FETCH }
-      new_pm.save(validate: false) # To skip channel validation
+      ProjectMedia.transaction do
+        # Save the new item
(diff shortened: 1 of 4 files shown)


app/models/project_media.rb
```diff
@@ -250,48 +250,51 @@ def replace_by(new_pm)
...
       new_pm.updated_at = Time.now
       new_pm.skip_check_ability = true
       new_pm.channel = { main: CheckChannels::ChannelCodes::FETCH }
       new_pm.save(validate: false) # To skip channel validation
-      new_pm.analysis = { title: analysis['title'], content: analysis['content'] }
-      # Point the claim
+
+      # Point the claim and consequently the fact-check
+      new_pm = ProjectMedia.find(new_pm.id)
       new_pm.claim_description = self.claim_description
```
melsawy, line 260: I think we can move this line before L256 and do one save instead of load item again and re-save it

app/models/project_media.rb
```diff
@@ -250,48 +250,51 @@ def replace_by(new_pm)
...
+      new_pm.save!
+
+      # All annotations from the old item should point to the new item
+      # Remove any status and report from the new item
+      Annotation.where(annotation_type: ['verification_status', 'report_design'], annotated_type: 'ProjectMedia', annotated_id: new_pm.id).delete_all
+      Annotation.where(annotated_type: 'ProjectMedia', annotated_id: self.id).update_all(annotated_id: new_pm.id)
+
+      # All versions from the old item should point to the new item
+      Version.from_partition(self.team_id).where(associated_type: 'ProjectMedia', associated_id: self.id).update_all(annotated_id: new_pm.id)
```
melsawy, line 269: Should update associated_id and for the actions related to ProjectMedia we should also update item_id

caiosba: Thanks Sawy, done.

PR #1249, opened 2022-10-13
Configure mocha to automatically teardown stubs after test run

hartsick:
In order to get mocha to tear down stubs, it has to be loaded last. I think that previously we were importing it before some of our minitest libraries, and also weren't using the minitest-specific version of the gem, which meant that we lost the auto-teardown functionality.

Changing the version of mocha and the load order means we no longer need to manually tear down mocks as part of test setup and teardown. (See https://github.com/freerange/mocha#installation)

See related Pender commit: https://github.com/meedan/pender/pull/281/commits/eac701330e9259731a36b9f15337da3e3e383eb6

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -18,6 +18,7 @@ group :test do
   gem 'rails-controller-testing'
   gem 'pact-consumer-minitest'
   gem 'rspec-rails', '4.0.1'
+  gem 'mocha', '~>1.14.0'
 end
 
 group :doc do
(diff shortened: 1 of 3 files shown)


hartsick:
One thing I noticed while doing this is that for some of our tests when we set an expectation on the mock but then unstub before the test finishes ([`test/models/project_media_test.rb:923`](https://github.com/meedan/check-api/blob/develop/test/models/project_media_test.rb#L923) and [`test/models/project_media_test.rb:1908`](https://github.com/meedan/check-api/blob/develop/test/models/project_media_test.rb#L1908) are examples) we don't actually make any expectations - the test passes regardless of what happens. If you remove the unstub at the end the tests are currently failing. I didn't update because I wasn't sure of the behavior we wanted to test.

I think this is another reason why we should stop unstubbing if we don't need to - and will need to dedicate a little time to updating tests when we do. Seems like a good "cleanup as you touch a file" task

PR #1274, opened 2022-10-31
CHECK-2383: Cluster with elasticsearch and meantokens

computermacgyver:


app/models/request.rb
```diff
@@ -40,8 +40,8 @@ def attach_to_similar_request!
     # First try to find an identical media
     similar_request_id = Request.where(media_id: media.id, feed_id: self.feed_id).where.not(id: self.id).order('id ASC').first
     if similar_request_id.nil?
-      if media.type == 'Claim' && ::Bot::Alegre.get_number_of_words(media.quote) > 3
-        params = { text: media.quote, threshold: threshold, context: context }
+      if media.type == 'Claim' && ::Bot::Alegre.get_number_of_words(media.quote) > 1
+        params = { text: media.quote, threshold: {::Bot::Alegre::ELASTICSEARCH_MODEL => 0.85, ::Bot::Alegre::MEAN_TOKENS_MODEL =>  0.9}, context: context }
```
computermacgyver, line 44: @ahmednasserswe Please change `threshold` to `per_model_threshold`
Ref: https://github.com/meedan/alegre/blob/ff7954c3d26db861cb4487c0555048d0b6ea1e67/app/main/lib/text_similarity.py#L48

This change will also be needed in the test file.

PR #1293, opened 2022-11-22
Add migration to delete unused Annotation Type for Archive.is

hartsick:
The rest of code was deleted in a previous commit.

### DIFF
--- a/db/migrate/20221122015259_remove_archive_is_annotation_type.rb
+++ b/db/migrate/20221122015259_remove_archive_is_annotation_type.rb
@@ -1,0 +1,9 @@
+class RemoveArchiveIsAnnotationType < ActiveRecord::Migration[5.2]
+  class DynamicAnnotationType < ActiveRecord::Base
+    self.table_name = :dynamic_annotation_annotation_types
+  end
+
+  def change
+    DynamicAnnotationType.where(annotation_type: 'archive_is').delete_all
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2022_11_22_013817) do
+ActiveRecord::Schema.define(version: 2022_11_22_015259) do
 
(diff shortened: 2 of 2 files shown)


db/migrate/20221122015259_remove_archive_is_annotation_type.rb
```diff
@@ -0,0 +1,9 @@
+class RemoveArchiveIsAnnotationType < ActiveRecord::Migration[5.2]
+  class DynamicAnnotationType < ActiveRecord::Base
+    self.table_name = :dynamic_annotation_annotation_types
+  end
```
caiosba, line 4: Is this needed if you use `DynamicAnnotation::AnnotationType` below, in the `change` method?

caiosba (changes requested):
See inline comment please.

db/migrate/20221122015259_remove_archive_is_annotation_type.rb
hartsick: @caiosba I prefer not to use application models in migrations, since model names and methods can change over time and break past migrations. I think using a temporary class is considered a migration best practice, since it ensures that we can always run the migration in the future even if the model changes significantly.

here's a blogpost that talks about the approach: https://jakeyesbeck.com/2021/04/10/avoid-models-in-migrations/

another option I've used with the same effect is doing data migrations as idempotent, automatically run on deploy rake tasks that are periodically deleted.

caiosba (approved):

db/migrate/20221122015259_remove_archive_is_annotation_type.rb
caiosba: Ah right, that makes sense and I agree it's a good practice!

When we have long-running tasks like that, we don't implement as migrations, we implement as rake tasks that are executed after a deployment: https://github.com/meedan/check-api/tree/develop/lib/tasks/migrate. But since this one is super fast, I think it's better to keep it as a migration but using your temporary class approach.

hartsick: okay great. I didn't realize we also had that practice, good to know!

PR #1294, opened 2022-11-23
Feature/2624 support 3 letter language code

amoedoamorim:
This commit fixes language names display in Tipline bots.
Bhojpuri will be displayed as its native name "भोजपुर" and support is added to display different names for dialects.


caiosba (approved):
Looks good! I just have a question on the last file, but if the test pass, we are good to go

PR #1296, opened 2022-11-26
Return an error on webhook controller if issue archiving

hartsick:
We were seeing that in a few cases, particularly in reporting metrics and in archiving (Perma CC), that Pender was trying to notify the webhook of a response before Check API had a chance to make all of the related models (link, project media, annotations) to save the information. Previously we were working around those issues and always returning a successful response, which meant that Pender didn't realize that it should retry, and archiving (which never retries automatically) was failing despite having a successful response from the archiving service.

To address this, I'm now raising exceptions in Bot::Keep that are caught by the controller and rendered as a 425 (too early) error. On the other side, Pender is now looking for errors when it notifies the webhook and will retry using Sidekiq.


app/models/bot/keep.rb
```diff
@@ -54,29 +66,32 @@ def self.webhook(request)
...
+        m.set_fields = { metadata_value: data.to_json }.to_json
+        m.save!
+
+        project_media = ProjectMedia.where(media_id: link.id)
+        raise ObjectNotReadyError.new('project_media') unless project_media.count > 0
+
+        project_media.find_each do |pm|
+          annotation = pm.annotations.where(annotation_type: 'archiver').last
+
+          unless annotation.nil? || !DynamicAnnotation::Field.where(field_name: "#{type}_response", annotation_id: annotation.id).exists?
```
hartsick, line 89: here's the part I'm wondering about. what's the best way to know we should throw an exception?

caiosba: @hartsick , looking at the data we have, I think that in most cases it will be just one `ProjectMedia` per `Media`/`Link`, so I'm inclined to simplify based on that. That said, I think we can raise an exception when `Link` is `nil` or when at least one of the `ProjectMedia` doesn't have the `archiver` annotation. What do you think?

caiosba:
Pull request looks great! Loved the idea of using HTTP code 425, can you please add it to https://github.com/meedan/check-api/blob/develop/lib/error_codes.rb so we can reuse it in other places?

PR #1297, opened 2022-11-28
CHECK-1900: Mandatory fact-check language

melsawy:
…ns in the report design file


db/migrate/20221129063218_adjust_report_design_schema.rb
```diff
@@ -0,0 +1,30 @@
...
+
+      # We need to do this in the migration, otherwise the app can be inconsistent
+
+      n = Dynamic.where(annotation_type: 'report_design').count
+      i = 0
+      Dynamic.where(annotation_type: 'report_design').find_each do |report|
+        i += 1
+        puts "[#{Time.now}] (#{i}/#{n}) Updating report with ID #{report.id}..."
+        data = report.data.with_indifferent_access
+        data[:options] = data[:options].first
```
caiosba, line 22: @melsawy we can't just get the first one, we need to be sure that it is not empty... I suggest we first check for the size, if there is only one, we get it... otherwise, we get the first one that is not empty (we can check for title and description)

db/migrate/20221129063218_adjust_report_design_schema.rb
```diff
@@ -0,0 +1,30 @@
...
+    at = DynamicAnnotation::AnnotationType.where(annotation_type: 'report_design').last
+    unless at.nil?
+      json_schema = at.json_schema.clone.with_indifferent_access
+      options_schema = json_schema[:properties][:options][:items]
+      json_schema[:properties][:options] = options_schema
+      
+      at.json_schema = json_schema
+      at.save!
+
+      # We need to do this in the migration, otherwise the app can be inconsistent
```
caiosba, line 14: In that case, let's schedule a deployment date/time where we can take the app down, @melsawy .

db/migrate/20221129063218_adjust_report_design_schema.rb
```diff
@@ -0,0 +1,30 @@
...
+
+      n = Dynamic.where(annotation_type: 'report_design').count
+      i = 0
+      Dynamic.where(annotation_type: 'report_design').find_each do |report|
+        i += 1
+        puts "[#{Time.now}] (#{i}/#{n}) Updating report with ID #{report.id}..."
+        data = report.data.with_indifferent_access
+        data[:options] = data[:options].first
+        report.data = data
+        report.save!
```
caiosba, line 24: @melsawy I suggest we add a `begin ... rescue` around this and output the failed IDs so we can look into them later, but not stop the migration if something fails.

Or we can make this migration idempotent, by ignoring the reports that are already in the right format.

db/migrate/20221129063218_adjust_report_design_schema.rb
melsawy: I'll move this migration to rake task as we have around `227920` items to migrate and run it immediately after deployment.
and either to take app down or expect some user can not save a fact check for a while

db/migrate/20221129063218_adjust_report_design_schema.rb
melsawy: Done

PR #1337, opened 2022-12-21
Simplify statistics rake task to only generate tipline data

hartsick:
To do:
- [x] Cache past data and only re-generate current month
- [ ] Remove passing of team slugs, instead generate for all Teams that have active tiplines that month
- [ ] Address what happens at month boundaries (ie need to regenerate previous month's data, when a month change happened overnight)

Questions / feedback sought:
* I don't love this caching approach where I am now modifying the data to put back in the same store, and relying on it as a historical data record. It feels brittle (storing an array of a bunch of hash objects) now that we're trying to use it more like a queryable data store. Any ideas?
* I decided to keep with caching the data on a per-team basis rather than on a per-team, per-month basis because though the latter was easier to recall data to update the cache it seemed more difficult to gather all relevant data for returning to the client (I don't think you can do eg give me all cached data under key prefix x)

Done so far:
* Only generate this month's data, otherwise fall back to cache
* Bring in Caio's updates to statistics generation
* Simplifies how we call the statistics rake task, by only passing in team slugs. Statistics are regenerated for each month start at the first month a project media was created for an integration
* Adds a simple sanity test for the rake task, so that we can move forward with refactoring with confidence. Since the get_statistics method is the most complicated and we don't need to change it, I extracted it to an external module so that we can mock the behavior for now
* Removes all day and week generated statistics for tiplines, and all feeds statistics. Previously these were being uploaded to S3 for manual processing by Pierre for partners. We are going to move these into Metabase, so are removing them to simplify the rake task and instead retaining only the statistics that are added to the cache to be displayed in the web app

CHECK-2657

[CHECK-2657]: https://meedan.atlassian.net/browse/CHECK-2657?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/check_statistics.rb
+++ b/lib/check_statistics.rb
@@ -1,0 +1,216 @@
+module CheckStatistics
+  class << self
+    def requests(slug, platform, start_date, end_date, language, type = nil)
+      relation = Annotation
+        .where(annotation_type: 'smooch')
+        .joins("INNER JOIN dynamic_annotation_fields fs ON fs.annotation_id = annotations.id AND fs.field_name = 'smooch_data'")
+        .where("fs.value_json->'source'->>'type' = ?", platform)
+        .where("fs.value_json->>'language' = ?", language)
+        .where('t.slug' => slug)
+        .where('annotations.created_at' => start_date..end_date)
+      unless type.nil?
+        relation = relation
+          .joins("INNER JOIN dynamic_annotation_fields fs2 ON fs2.annotation_id = annotations.id AND fs2.field_name = 'smooch_request_type'")
+          .where('fs2.value' => type.to_json)
+      end
+      relation
+    end
+
+    def reports_received(slug, platform, start_date, end_date, language)
+      DynamicAnnotation::Field
+        .where(field_name: 'smooch_report_received')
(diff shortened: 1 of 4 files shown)


caiosba:
> I don't love this caching approach where I am now modifying the data to put back in the same store, and relying on it as a 
> historical data record. It feels brittle (storing an array of a bunch of hash objects) now that we're trying to use it more like a 
> queryable data store. Any ideas?

Yeah, previously it made more sense because everything was changed at once, atomically, but now that it's incremental, it's weird and can lead to inconsistency. I'll give it some thought.

The rest looks great! Thanks for adding tests, super important.

caiosba:
When dealing with conflicts with `develop`, this is the only change you need to incorporate: https://github.com/meedan/check-api/blob/126c2cdc5ac7bed5c7b9bab8474b627d8b6dfdc5/lib/tasks/data/statistics.rake#L197-L203

hartsick:
Closing this in favor of working on statistics in Metabase for now (see https://meedan.atlassian.net/browse/CHECK-2657?focusedCommentId=25682 for more details)

PR #1343, opened 2023-01-04
Simplify statistics rake task to only generate tipline data

hartsick:
This PR simplifies our rake task for statistics generation, by:
1. Detecting what workspaces need to have statistics generated by looking which have historic tipline data, rather than passing in the workspace slugs. Data is then generated for each month (segmented by language and platform) since the first submitted tipline project media
2. Removing all tipline statistics (day and weekly data) that are not displayed within the application
3. Removing all feed statistics

For points 2 and 3, these statistics were being uploaded to S3 for manual processing for partners. We are going to move these into Metabase, so are removing them to simplify the rake task and instead retaining only the statistics that are added to the cache to be displayed in the web app

This also adds a simple sanity test for the rake task, so that we can move forward with refactoring with confidence. Since the get_statistics method is the most complicated and we don't need to change it, I extracted it to an external module so that we can mock the behavior for now. Methods for generating the statistics data have not been modified from the previous implementation.

CHECK-2657

[CHECK-2657]: https://meedan.atlassian.net/browse/CHECK-2657?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/check_statistics.rb
+++ b/lib/check_statistics.rb
@@ -1,0 +1,216 @@
+module CheckStatistics
+  class << self
+    def requests(slug, platform, start_date, end_date, language, type = nil)
+      relation = Annotation
+        .where(annotation_type: 'smooch')
+        .joins("INNER JOIN dynamic_annotation_fields fs ON fs.annotation_id = annotations.id AND fs.field_name = 'smooch_data'")
+        .where("fs.value_json->'source'->>'type' = ?", platform)
+        .where("fs.value_json->>'language' = ?", language)
+        .where('t.slug' => slug)
+        .where('annotations.created_at' => start_date..end_date)
+      unless type.nil?
+        relation = relation
+          .joins("INNER JOIN dynamic_annotation_fields fs2 ON fs2.annotation_id = annotations.id AND fs2.field_name = 'smooch_request_type'")
+          .where('fs2.value' => type.to_json)
+      end
+      relation
+    end
+
+    def reports_received(slug, platform, start_date, end_date, language)
+      DynamicAnnotation::Field
+        .where(field_name: 'smooch_report_received')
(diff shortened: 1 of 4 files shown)


lib/tasks/data/statistics.rake
```diff
@@ -1,402 +1,67 @@
...
-          'New newsletter subscriptions',
-          'Newsletter cancellations',
-          'Current subscribers'
-        ]
-        outfile.puts(header.join(','))
 
-        slugs.each do |slug|
-          team = Team.find_by_slug(slug)
-          team_rows = []
+      team_ids = Team.joins(:project_medias).where(project_medias: { user: BotUser.smooch_user }).distinct.pluck(:id)
```
hartsick, line 13: @caiosba sanity check (flagging from the Jira comment): do you think this makes sense as a way to detect which workspaces we should be generating statistics for? I took the logic from the previous guard clause on line 50, but inverted it. seems to make sense to me, since I think we want to regenerate historic data.

caiosba: Good catch, yes, that is important because then we can keep generating statistics for workspaces even after they uninstall the tipline bot.

lib/check_statistics.rb
```diff
@@ -0,0 +1,216 @@
...
+    def team_requests(slug, platform, start_date, end_date, language)
+      base = requests(slug, platform, start_date, end_date, language)
+      base.joins("INNER JOIN teams t ON annotations.annotated_type = 'Team' AND t.id = annotations.annotated_id")
+    end
+
+    def unique_requests_count(relation)
+      relation.group("fs.value_json #>> '{source,originalMessageId}'").count.size
+    end
+
+    def get_statistics(start_date, end_date, slug, platform, language)
```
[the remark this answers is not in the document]
hartsick: I agree with this, but am going to ignore since it exists currently and we intend to move away from generating statistics this way.

caiosba:
@hartsick , yes, you can ignore all coverage and Code Climate alerts, since this code is not used by the running application and we're moving away from it soon anyway.

PR #1354, opened 2023-01-11
Save generated statistics to Postgres instead of Rails cache

hartsick:
Previously we were generating statistics and storing them to the Rails cache. The entries could be very large - one row per language, platform, and month since a tipline was enabled, for each workspace. We were also generating statistics for each month each time we ran it, which took a long time.

This PR moves toward saving the data more incrementally as a single-purpose ActiveRecord model, and determines which months to generate based on what months have missing or partial data.

From the frontend's perspective, this change should not change any behavior except for two small things:
* The ID returned with each row of the report is different (numeric ID from database instead of generated ID with date, platform, etc)
* Month is in format of `1. Jun 2020` instead of `1. 2020-06-01` (most recent change) or `1. June 2020` (currently deployed behavior)

Some details:

**MonthlyTeamStatistic**
The intention of the MonthlyTeamStatistic model is to only be used for monthly statistic generation. We could change this in the future if we wanted—bounding to a month allowed me to set a validation on the time period. As written currently, we only have one MonthlyTeamStatistic per team, language, platform, and start date (assuming that we would want to only have say one object for Test Team on WhatsApp in English for July, but that the end date might change if we're midmonth and regenerating so isn't worth validating).

The model has the ability for us to specific methods for formatting, which we can then configure in the `FIELD_MAPPINGS` hash so that they are used in the `.formatted_hash` method, which is used for preparing the response to client.

**Approach to generating data**
I kept to the same approach we were using before, which looks at the first ProjectMedia for a tipline for a given team and then iterates through each month of data for each platform and language the team has enabled.

* If full data is present for the month (ie there is a MonthlyTeamStatistic with a start and end date matching start and end date of that month), it continues to next month
* If partial data is present for the month (ie there is a MonthlyTeamStatistic with a start date matching month start but end date not matching the end - representing partially calculated data, like when the task was run on a month boundary), it regenerates and updates the model
* If no data is present for the month (ie there is no MonthlyTeamStatistic - representing a new month), it generates and creates a model

**Note**: I'm a little worried about the performance of this since we are still iterating over a lot of data and touching the database once per iteration (vs assembling the array and saving the whole thing once to cache like before). Want to see if we can try and monitor in an environment with more data than I have locally. We could do a bulk insert, but would miss the ActiveRecord validations so wanted to avoid if we can.

**Changes to get_statistics**
* Returns a hash from CheckStatistics.get_statistics, so that we can more easily keep track of what values are being set instead of relying on positional data
* Removes generation of statistics that we were filtering out of the response to the front end (search, some message related fields)
* Changes the ID format for the returned data structure. It's still unique because it's a database index, but is no longer human readable in the same way

## How I checked locally

I ran the rake task on develop on a workspace with tipline enabled and some minimal data - `bundle exec rake check:data:statistics`. Entered rails console and checked report `Team.find(team_id).data_report` and checked Rails cache `Rails.cache.read("data:report:#{team_id}"). Then switched branches, rebuilt container and reran rake task, then in rails console checked return `Team.find(team_id).data_report` and then the cached values `MonthlyTeamStatistic.where(team_id: team_id)`.

* UI looks the same
* Output from team.data_report looks the same
* Confirmed that database is caching values (saves past months, and then doesn't update the non-current months)

**this branch result for team.data_report**
```
[
  {
    "ID"=> 1,
    "Org"=> "test",
    "Platform"=> "Telegram",
    "Language"=> "en",
    "Month"=> "1. Dec 2022",
    "Conversations"=> 1,
    "Average messages per day"=> 0,
    "Unique users"=> 1,
    "Returning users"=> 0,
    "Valid new requests"=> 0,
    "Published native reports"=> 0,
    "Published imported reports"=> 0,
    "Requests answered with a report"=> 0,
    "Reports sent to users"=> 0,
    "Unique users who received a report"=> 0,
    "Average (median) response time"=> "-",
    "Unique newsletters sent"=> 0,
    "New newsletter subscriptions"=> 0,
    "Newsletter cancellations"=> 0,
    "Current subscribers"=> 0
  },
  {
    "ID"=> 2
    "Org"=> "test",
    "Platform"=> "Telegram",
    "Language"=> "en",
    "Month"=> "2. Jan 2023",
    "Conversations"=> 0,
    "Average messages per day"=> 0,
    "Unique users"=> 0,
    "Returning users"=> 0,
    "Valid new requests"=> 0,
    "Published native reports"=> 0,
    "Published imported reports"=> 0,
    "Requests answered with a report"=> 0,
    "Reports sent to users"=> 0,
    "Unique users who received a report"=> 0,
    "Average (median) response time"=> "-",
    "Unique newsletters sent"=> 0,
    "New newsletter subscriptions"=> 0,
    "Newsletter cancellations"=> 0,
    "Current subscribers"=> 0
  }
]
```

**develop result**
```
[
  {
    "ID"=>"test-telegram-en-2022-12-01-2022-12-31",
    "Org"=>"test",
    "Platform"=>"Telegram",
    "Language"=>"en",
    "Month"=>"1. 2022-12-01",
    "Conversations"=>1,
    "Average messages per day"=>0,
    "Unique users"=>1,
    "Returning users"=>0,
    "Valid new requests"=>0,
    "Published native reports"=>0,
    "Published imported reports"=>0,
    "Requests answered with a report"=>0,
    "Reports sent to users"=>0,
    "Unique users who received a report"=>0,
    "Average (median) response time"=>"-",
    "Unique newsletters sent"=>0,
    "New newsletter subscriptions"=>0,
    "Newsletter cancellations"=>0,
    "Current subscribers"=>0
  },
  {
    "ID"=>"test-telegram-en-2023-01-01-2023-01-31",
    "Org"=>"test",
    "Platform"=>"Telegram",
    "Language"=>"en",
    "Month"=>"2. 2023-01-01",
    "Conversations"=>0,
    "Average messages per day"=>0,
    "Unique users"=>0,
    "Returning users"=>0,
    "Valid new requests"=>0,
    "Published native reports"=>0,
    "Published imported reports"=>0,
    "Requests answered with a report"=>0,
    "Reports sent to users"=>0,
    "Unique users who received a report"=>0,
    "Average (median) response time"=>"-",
    "Unique newsletters sent"=>0,
    "New newsletter subscriptions"=>0,
    "Newsletter cancellations"=>0,
    "Current subscribers"=>0
  }
]
```

**Times to run:**
* current branch: 1st run: real 23s, user 3.2s, sys 1.10s; 2nd run: real 23s, user 3.1s, sys 1.8s
* develop: 1st run: real 23.4s, user 3.6s, sys 1.2s; 2nd run: real 28.6s, user 3.3s, sys 1.3s

Honeycomb trace of running the current branch and then develop: https://ui.honeycomb.io/meedan/environments/dev/datasets/check-api/result/CuANxUVZErU
* [Example trace of task on this branch](https://ui.honeycomb.io/meedan/environments/dev/datasets/check-api/result/CuANxUVZErU/trace/ym4PEgb8JBT?fields[]=c_name&fields[]=c_service.name&fields[]=c_library.name&span=350b056433a981ee) - duration 1.1s with 130 spans
* [Example trace of task on develop](https://ui.honeycomb.io/meedan/environments/dev/datasets/check-api/result/CuANxUVZErU/trace/CVvq35DtVxP?fields[]=c_name&fields[]=c_service.name&fields[]=c_library.name&span=ac46f12b58137070) - duration 7.9s with 318 spans

Hopefully will see more improvement with larger dataset, but we'll see!

CHECK-2892

### DIFF
--- a/app/models/concerns/team_associations.rb
+++ b/app/models/concerns/team_associations.rb
@@ -17,6 +17,7 @@ module TeamAssociations
     has_many :project_groups, dependent: :destroy
     has_many :feed_teams, dependent: :destroy
     has_many :feeds, through: :feed_teams
+    has_many :monthly_team_statistics # No "dependent: :destroy" because we want to retain statistics
 
     has_annotations
   end
--- a/app/models/monthly_team_statistic.rb
+++ b/app/models/monthly_team_statistic.rb
@@ -1,0 +1,53 @@
+class MonthlyTeamStatistic < ApplicationRecord
+  belongs_to :team
+
+  validates_presence_of :start_date, :end_date, :team, :platform, :language
+
+  include ActionView::Helpers::DateHelper
+
+  # Mapping of attributes to human-readable descriptions
+  FIELD_MAPPINGS = {
+    id: "ID",
+    org: "Org", # model method
+    platform: "Platform",
+    language: "Language",
+    month: "Month", # model method
+    conversations: 'Conversations',
+    average_messages_per_day: 'Average messages per day',
+    unique_users: 'Unique users',
+    returning_users: 'Returning users',
+    valid_new_requests: 'Valid new requests',
+    published_native_reports: 'Published native reports',
+    published_imported_reports: 'Published imported reports',
+    requests_answered_with_report: 'Requests answered with a report',
+    reports_sent_to_users: 'Reports sent to users',
+    unique_users_who_received_report: 'Unique users who received a report',
+    formatted_median_response_time: 'Average (median) response time', # model method
+    unique_newsletters_sent: 'Unique newsletters sent',
+    new_newsletter_subscriptions: 'New newsletter subscriptions',
+    newsletter_cancellations: 'Newsletter cancellations',
+    current_subscribers: 'Current subscribers',
+  }.freeze
+
+  def formatted_hash
+    {}.tap do |formatted_hash|
+      FIELD_MAPPINGS.each do |attribute_name, human_readable_label|
+        formatted_hash[human_readable_label] = public_send(attribute_name.to_sym) || '-'
+      end
+    end
+  end
+
+  # Below methods must match a key in FIELD_MAPPINGS to be included in
+  # the .formatted_hash output
+  def org
+    team.name
+  end
+
+  def month
+    start_date&.strftime('%b %Y')
+  end
+
+  def formatted_median_response_time
+    distance_of_time_in_words(median_response_time) if median_response_time
+  end
+end
--- a/app/models/monthly_team_statistic.rb
+++ b/app/models/monthly_team_statistic.rb
@@ -2,6 +2,7 @@ class MonthlyTeamStatistic < ApplicationRecord
   belongs_to :team
 
   validates_presence_of :start_date, :end_date, :team, :platform, :language
+  validates :start_date, uniqueness: { scope: [:platform, :language, :team] }
 
   include ActionView::Helpers::DateHelper
 
@@ -9,7 +10,7 @@ class MonthlyTeamStatistic < ApplicationRecord
   FIELD_MAPPINGS = {
     id: "ID",
     org: "Org", # model method
-    platform: "Platform",
+    platform_name: "Platform", # model method
     language: "Language",
     month: "Month", # model method
     conversations: 'Conversations',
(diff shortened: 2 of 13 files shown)


caiosba (approved):

caiosba:
Looks great! Great work @hartsick !

PR #1370, opened 2023-01-19
Save tipline messages when messages sent and received

hartsick:
Specific questions I had noted inline as comments below, but all feedback and other questions welcome!

## To do:
- [x] Test locally with tipline setup for newsletter, report, and user message testcases
    - Tested with a local Telegram tipline and my personal phone
    - Was able to see a TiplineMessage created for each message delivered to my phone, with language ("en"), platform ("Telegram"), sent_at, uid, external_id, and payload all set as expected
    - Event was present for messages associated with a newsletter or report creation / update

## Done:

**Store messages to tipline in database - from user & when newsletter sent**

In order to move away from calculating conversations from the data stored
in annotations, we are beginning to save information in a new model, named
TiplineMessage. These are intended to be saved each time a message is sent
or received, and we are beginning to save them when a user sends a message
to a tipline (incoming) or a partner sends a newsletter to a user (outgoing).

To avoid potential race conditions, we add a validation that looks at uniqueness
both on the external ID from Smooch (which should be unique to the message) and
other indicators, such as the time it was sent, platform, language, uid, and
team.

We were mostly able to use the existing methods to pull data out of the
payload from smooch, but there was one instance where the payload for
outgoing (newsletter) trigger events differs from incoming (user messages)
events - outgoing events have a truncated message body, which requires
us to pull the platform information and timestamp from the payload.

Event is set from the message cache for important message types - report activity and newsletter sending, since we need to track these for statistics generation.

**Remove Rspec from Gemfile dependencies**

We do need Rspec as a dependency of Pact for our contract tests, but we aren't using it directly. Including it in the Gemfile makes running our scaffold generators (eg `rails g model Foo`) add rspec files instead of test unit test files, which are Minitest- compatible.

CV2-2647

### DIFF
--- a/app/models/concerns/smooch_messages.rb
+++ b/app/models/concerns/smooch_messages.rb
@@ -144,11 +144,20 @@ module SmoochMessages
       disabled ? 'disabled' : state
     end
 
+    # Used for incoming messages (e.g. message:appUser)
+    # where full message contents available
     def get_platform_from_message(message)
       type = message.dig('source', 'type')
       platform = type ? ::Bot::Smooch::SUPPORTED_INTEGRATION_NAMES[type].to_s : 'Unknown'
       RequestStore.store[:smooch_bot_platform] = platform
       platform
+    end
+
+    # Used for outgoing messages (e.g. message:delivery:channel) where
+    # message contents are truncated
+    def get_platform_from_payload(payload)
+      type = payload.dig('destination', 'type')
+      type ? ::Bot::Smooch::SUPPORTED_INTEGRATION_NAMES[type].to_s : 'Unknown'
     end
 
     def request_platform
--- a/app/models/concerns/smooch_messages.rb
+++ b/app/models/concerns/smooch_messages.rb
@@ -199,9 +199,10 @@ module SmoochMessages
     end
 
     def process_message(message, app_id, send_message = true, type = 'default_requests')
-      message['language'] = self.get_user_language(message)
+      uid = message['authorId']
+      message['language'] = self.get_user_language(uid)
 
-      return if !Rails.cache.read("smooch:banned:#{message['authorId']}").nil?
+      return if !Rails.cache.read("smooch:banned:#{uid}").nil?
 
       hash = self.message_hash(message)
       pm_id = Rails.cache.read("smooch:message:#{hash}")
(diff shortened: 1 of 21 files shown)


app/models/concerns/smooch_messages.rb
```diff
@@ -144,13 +144,22 @@ def send_message_if_disabled_and_return_state(uid, workflow, state)
...
     def get_platform_from_message(message)
       type = message.dig('source', 'type')
       platform = type ? ::Bot::Smooch::SUPPORTED_INTEGRATION_NAMES[type].to_s : 'Unknown'
       RequestStore.store[:smooch_bot_platform] = platform
       platform
     end
 
+    # Used for outgoing messages (e.g. message:delivery:channel) where
+    # message contents are truncated
+    def get_platform_from_payload(payload)
```
hartsick, line 158: Question for review: does it make sense to have this method here? I thought of modifying the method above to allow either a payload or message JSON to be passed, but thought that it might be confusing and also that we might not want to alter the request store for sent messages, since we aren't already.

app/models/tipline_message.rb
```diff
@@ -0,0 +1,38 @@
+class TiplineMessage < ApplicationRecord
+  enum direction: { direction_unset: 0, incoming: 1, outgoing: 2 } # default: direction_unset
```
hartsick, line 2: Question for review: I'm wondering here if this abstraction makes sense, or if we should just directly store the trigger (which we still have saved in the model under payload)

app/models/concerns/smooch_messages.rb
caiosba: Yes, I think it makes more sense to have it as a separate method.

app/models/tipline_message.rb
caiosba: I think it makes sense because it helps a lot with code readability. We usually do that using constants to make the stored number to its meaning (for example: https://github.com/meedan/check-api/blob/develop/lib/check_archived_flags.rb). But it's OK to use an `enum` too.

caiosba (approved):
I answered a couple of questions but it looks great! I didn't test, though... happy to help on that too. Let me know please when it's "ready".

hartsick:
@caiosba I'm waiting for CI to finish to see if tests pass, but have tested this locally successfully and it's ready for re-review!

caiosba (approved):
Looks great, fantastic work Christa! Today we don't have a use case to index the `event` column, but in the future we may want to do that.

PR #1421, opened 2023-03-01
Add rake task to migrate smooch_data from annotations to TiplineMessage

hartsick:
This rake task intends to take the smooch data we have stored for annotations and translate them into the new TiplineMessage format, so that we can more easily calculate historic statistics data.

We iterate through a few types of saved data:
* Messages sent to users (via project media and team requests)

In this case, we are previously only storing one annotation for an entire lifecycle of a message, from receipt of initial message to submission of claim. This means that multiple of what we now consider TiplineMessages are represented by one Annotation and have been collapsed into one external_id, which in the new world will be unique. To get around this, we make the following modifications: 1) split up the message text by the text separator (a newline), 2) create two TiplineMessages for each text in that cycle - one outgoing, and one incoming, 3) create unique external_ids that can be related back to the original annotation (by using the original message ID for first, then appending an index to subsquent, and appending "_sent" to the outgoing messages).

* Reports sent to users

For this, we take a similar approach to above without mimicing conversations. Since we now look to the "event" key on TiplineMessage to understand meaningful events, usually represented by template we sent, we hardcode this to "fact_check_report_annotation".

* Newsletters sent

This one is different than the other approaches. For this, we see every newsletter sent by an organization for each language. We then go through and calculate the number of newsletters sent by looking at the number of TiplineSubscriptions for the newsletter in that language at that time, and create one TiplineMessage for each subscription (representing a user and platform). To do this, we must again create a synthetic external_id.

--

Additional notes:
* For each of these approaches, we drop any records with duplicate external_ids at import, since we assume they are duplicates of what we already have. We have seen this in the reports and conversations data, since it seems like we just get multiple webhook events for the same messages.
* For all of these approaches, we specify some legacy data flags that attempt to give us ways of tracking taht this is historic (and differently formatted than the new) data, and of checking the original data it was made from (eg by storing smooch_data)
* Because this rake task will take awhile to run, I have added a cache that will help us skip completed teams on subsequent runs. Beacuse of compelxity within the individual calculations, I didn't attempt to cache on a more granular level
* To avoid duplicating TiplineMessages in the time between we've deployed the code that creates TipelineMessages when the webhook is called, and when this task is first run, the rake task looks at when the first non-legacy TiplineMessage was created and only imports data before that mark.
* I wanted to add automated tests to actually address the rake functionality, but setup was really difficult because our data model is hard! This should make it easier going forward!

CV2-2647

### DIFF
--- a/lib/tasks/migrate/20230224205948_add_legacy_smooch_data_to_tipline_message.rake
+++ b/lib/tasks/migrate/20230224205948_add_legacy_smooch_data_to_tipline_message.rake
@@ -1,0 +1,213 @@
+
+# A helper class for migrating smooch_data from existing tiplines
+# into the new TiplineMessage format
+class MigratedTiplineMessageHelper
+  class << self
+    def requests(team_id, cutoff_time)
+      relation = Annotation.where(annotation_type: 'smooch').joins("INNER JOIN dynamic_annotation_fields fs ON fs.annotation_id = annotations.id AND fs.field_name = 'smooch_data'")
+        .where('t.id' => team_id)
+        .where("annotations.created_at < ?", cutoff_time)
+    end
+
+    def project_media_requests(team_id, cutoff_time)
+      base = requests(team_id, cutoff_time)
+      base.joins("INNER JOIN project_medias pm ON pm.id = annotations.annotated_id AND annotations.annotated_type = 'ProjectMedia' INNER JOIN teams t ON t.id = pm.team_id")
+    end
+
+    def team_requests(team_id, cutoff_time)
+      base = requests(team_id, cutoff_time)
+      base.joins("INNER JOIN teams t ON annotations.annotated_type = 'Team' AND t.id = annotations.annotated_id")
+    end
+
+    def reports_received(team_id, cutoff_time)
+      DynamicAnnotation::Field
+        .where(field_name: 'smooch_report_received')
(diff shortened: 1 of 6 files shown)


lib/tasks/migrate/20230224205948_add_legacy_smooch_data_to_tipline_message.rake
```diff
@@ -0,0 +1,213 @@
...
+          Version.from_partition(team_id).
+            where(whodunnit: BotUser.smooch_user.id.to_s, item_id: tbi.id.to_s, item_type: ['TeamUser', 'TeamBotInstallation']).
+            where("created_at < ?", cutoff_time).each do |newsletter_version|
+              begin
+                YAML.load(JSON.parse(newsletter_version.object_after)['settings'])['smooch_workflows'].each do |smooch_workflow|
+                  language = smooch_workflow['smooch_workflow_language']
+                  sent_time = smooch_workflow['smooch_newsletter']['smooch_newsletter_last_sent_at']
+
+                  sent_newsletters_by_language[language] ||= []
+                  sent_newsletters_by_language[language] << sent_time.to_s
```
caiosba, line 182: Not every version will have a `smooch_newsletter_last_sent_at`, so let's add only if it's a valid timestamp.

caiosba (changes requested):
Just added one comment. Great work!

caiosba:
I tested it locally after the change I'm suggesting and it works well. But anyway, the real/right way to test this will be to generate statistics according to this new model and then compare the results.

PR #1431, opened 2023-03-10
[CV2-2675] Do not suggest to suggestions

skyemeedan:
https://meedan.atlassian.net/browse/CV2-2675
Alters the logic of how proposed similarity relationships are cached to deal with issues reported by users in which weak chains of similarity caused unrelated media to be suggested
- adds a test file that tries to enumerate the combinations of project media suggestions to cover the logc
- modifies the logic so that a suggestion to a suggestion will not be cached
- also implements the 'if a confirmed relationship is proposed to the child of a suggested relationship, break the (weaker) relationship with parent and form a new relationship cluster

### DIFF
--- a/app/models/bot/alegre.rb
+++ b/app/models/bot/alegre.rb
@@ -495,6 +495,7 @@ class Bot::Alegre < BotUser
 
   def self.return_prioritized_matches(pm_id_scores)
     if pm_id_scores.is_a?(Hash)
+      # make K negative so that we bias towards older IDs
       pm_id_scores.sort_by{|k,v| [Bot::Alegre::ELASTICSEARCH_MODEL != v[:model] ? 1 : 0, v[:score], -k]}.reverse
     elsif pm_id_scores.is_a?(Array)
       pm_id_scores.sort_by{|v| [Bot::Alegre::ELASTICSEARCH_MODEL != v[:model] ? 1 : 0, v[:score]]}.reverse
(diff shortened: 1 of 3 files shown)


app/models/bot/alegre.rb
```diff
@@ -505,46 +506,91 @@ def self.return_prioritized_matches(pm_id_scores)
...
+        pm_id_scores[parent_id][:relationship_type] = new_type if pm_id_scores[parent_id]
+      end
+
+      # (4) if the relationship to parent was only suggested, but new relationship is confirmed
+      if !parent_relationship.is_confirmed? & proposed_relationship_is_confirmed
         new_type = Relationship.confirmed_type
+        # break the old parent relationship involving proposed_id, make the proposed_id into a new parent
+        Rails.logger.info "[Alegre Bot] [ProjectMedia ##{pm.id}] [Relationships 3/6] [Relationships NOTE] removing suggested relationship pm_id, parent_id #{
+          [parent_relationship.source_id, parent_relationship.target_id].inspect}"
+        parent_relationship.destroy
```
caiosba, line 579: I suggest to use `destroy!` instead of `destroy` so we can get an Errbit exception if something goes wrong.

skyemeedan: for my education, what does the `!` do here?  ruby docs suggest it is kind of like a warning that this is modifying the state of the object?

computermacgyver:
> @caiosba the CI is failing with `The command "docker-compose exec -e PATTERN='controllers contract' api test/run-tests.sh" exited with 1.` I'm guessing this from the integration tests, probably because I really did change the behavior of an API function?

@skyemeedan There are two earlier unit tests failing. I suspect the behavior is now correct, and we'll need to update the tests to match, but obviously  need to check. 

The failures are [further up the output](https://app.travis-ci.com/github/meedan/check-api/jobs/598075194):
```
Failure:
Bot::Alegre3Test#test_resets_relationship_transitively [/app/test/models/bot/alegre_3_test.rb:351]:
...
Failure:
Bot::Alegre3Test#test_should_fail_to_add_relationships [/app/test/models/bot/alegre_3_test.rb:340]:
```

PR #1443, opened 2023-03-21
CV2-2893: fix CC

melsawy:


skyemeedan (approved):
thanks for cleaning up my edits!

PR #1449, opened 2023-03-27
push images built by commit to ecr

dmou:
This creates a Github Actions workflow to build and push arm64 and amd64 images for all non-deploy branch (ie. develop and master) commits to our private repository. This will not be triggered by forks from external contributors.

Pushed images will be tagged with both git sha (ie. `873f96246cebf3196ac4a566bd5157d61c94df0f`) and branch name (innermost sub-branch after last `/`).


huslage (commented):
Can we use OIDC instead of static AWS keys?

https://github.com/aws-actions/configure-aws-credentials v2 can do this

huslage (approved):
LGTM

PR #1451, opened 2023-03-27
Migrate Check API from Errbit to Sentry

hartsick:
Previously we used Errbit / Airbrake to track errors in our applications.
We now have access to Sentry, and are using it to track errors instead.
As a result, we need to replace the Airbrake SDK with the Sentry SDK.
Some additional details:

* Sentry is loaded in all environments, but only reports when DSN is
configured. This is expected to be set via SSM in deployed environments,
and only optionally configured (for testing) locally
* Similar to when we set context for Honeycomb, we attempt to set user
information (either logged in user ID or API key) when it is available
in the controller. This will help us understand and debug when errors are
only being reported for certain users / API keys. We use IDs to limit
the spread of PII for our users
* To maintain the previous behavior for reporting errors, we configure
Sidekiq to only report errors to Sentry when the job hits the max number
of retries. This will also suppress errors we want to track (since we aren't
manually saying which errors we expect to retry). We hope to improve
this in a future ticket by raising retryable errors and configuring a
death_handler for Sidekiq similar to how we do in Pender

Also raises custom errors for frequently seen issues. We may want to ignore these errors in Sentry, so we want to make sure that the errrors we raise are unique and help us understand what's going on.

Whereas Errbit surfaced the error message in the user interface, Sentry
relies more on error classes, including for permamently ignoring them.

These errors represent the following:
* A duplicate import for Fetch (a subset of ActiveRecord uniqueness errors). Should still work with any catches as before,
since the custom class inherits from the previous class - ActiveRecord::RecordNotUnique
* Different failure states for delivering messages to users


test/lib/check_notification_test.rb
```diff
@@ -15,14 +15,11 @@ class CheckNotificationTest < ActiveSupport::TestCase
   end
 
   test "should notify airbrake and not call Pusher when send notification with undefined info" do
-    Airbrake.stubs(:configured?).returns(true)
-    Airbrake.stubs(:notify).returns('notified airbrake')
+    CheckSentry.stubs(:notify).returns('notified airbrake')
```
caiosba, line 18: I know this doesn't have any impact, but just for consistency, can we replace airbrake by Sentry? :)

caiosba (approved):
I just added one cosmetic comment :)

test/lib/check_notification_test.rb
hartsick: Definitely! I just found a few other places where we still reference airbrake by name - going to change those, too, just to reduce confusion.

melsawy (approved):

PR #1456, opened 2023-03-29
[don't merge] Remove NodeJS and things relating to asset precompilation

hartsick:
We no longer use the asset pipeline for Check API, so let's try to remove the dependencies relating to it that we no longer need.

Also removes the asset precompilation step in the production Dockerfile, and the config that we only used for that purpose.

CV2-2723


caiosba (changes requested):
@melsawy I'm curious also on whether it affects the assets pipeline in general. For example, that error:

```
ActionView::Template::Error
The asset "graphiql/rails/application.css" is not present in the asset pipeline.
Sprockets::Rails::Helper::AssetNotFound
The asset "graphiql/rails/application.css" is not present in the asset pipeline.
```

Will it still happen?

Also, did you try if we are able to access:

- Swagger UI: http://localhost:3000/api
- Sidekiq UI: http://localhost:3000/sidekiq
- GraphiQL: http://localhost:3000/graphiql

melsawy:
@caiosba I fixed the one related to GraphiQL in another PR https://github.com/meedan/check-api/pull/1537
so please approve this one so I can merge

caiosba (approved):

PR #1462, opened 2023-04-03
Support common URL syntax for ElasticSearch connections

sonoransun:
In the past we've use parameterized configuration for ElasticSearch connections to explicitly set the username, password, port, and hostname for connections.

This change supports the common URL syntax used by other applications like Alegre and Fetch, making ElasticSearch configuration more consistent. Fallback to parameterized connections is supported if ELASTICSEARCH_URL is not set.


config/initializers/elasticsearch.rb
```diff
@@ -1,11 +1,18 @@
 require 'typhoeus'
 require 'typhoeus/adapters/faraday'
 
-user = CheckConfig.get('elasticsearch_user').to_s
-password = CheckConfig.get('elasticsearch_password').to_s
-host = CheckConfig.get('elasticsearch_host').to_s + ':' + CheckConfig.get('elasticsearch_port').to_s
-client = Elasticsearch::Client.new(log: (CheckConfig.get('elasticsearch_log').to_i == 1), user: user, password: password, host: host) do |f|
-  f.adapter :typhoeus
+unless ENV['ELASTICSEARCH_URL'] == ""
```
caiosba, line 4: Please use a stronger check: `unless ENV['ELASTICSEARCH_URL'].blank?`.

config/initializers/elasticsearch.rb
```diff
@@ -1,11 +1,18 @@
...
+  # NOTE: We may want to initialize explicitly with url: set for future compatibility.
+  client = Elasticsearch::Client.new(log: (CheckConfig.get('elasticsearch_log').to_i == 1)) do |f|
```
caiosba, line 6: So we don't need to explicitly pass the URL? When we don't pass any of the other parameters the lib is going to use `ENV['ELASTICSEARCH_URL']`?

config/initializers/elasticsearch.rb
sonoransun: Ah, good call. `.empty?` would not work right, but `.blank?` should do the trick!

Gemfile
```diff
@@ -58,9 +58,9 @@ gem 'devise_invitable', '2.0.5'
 gem 'omniauth-twitter'
 gem 'omniauth-facebook'
 gem 'omniauth-slack'
-gem 'elasticsearch', '~> 7.x'
-gem 'elasticsearch-api', '~> 7.x'
-gem 'elasticsearch-transport', '~> 7.x'
+gem 'elasticsearch', '7.10.1'
+gem 'elasticsearch-api', '7.10.1'
+gem 'elasticsearch-transport', '7.10.1'
```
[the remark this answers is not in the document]
sonoransun: Correct, these are the specific version pins we need for OpenSearch compatibility.

huslage:
@sonoransun PLease fix?

sonoransun:
@melsawy I looked into a unit test for ElasticSearch connection attempts, but all of the RSpec tests I could find related to query results rather than initialization.

One way around limitations of tests for initializers would be to have two distinct test invocations in CI, one using config with common URL syntax (and just one test?) and the rest as usual with default config and ENV.

Perhaps for now we can accept manual tests; in future library migration efforts we could then deprecate the parameterized initialization entirely.

melsawy:
> @melsawy I looked into a unit test for ElasticSearch connection attempts, but all of the RSpec tests I could find related to query results rather than initialization.
> 
> One way around limitations of tests for initializers would be to have two distinct test invocations in CI, one using config with common URL syntax (and just one test?) and the rest as usual with default config and ENV.
> 
> Perhaps for now we can accept manual tests; in future library migration efforts we could then deprecate the parameterized initialization entirely.



> @melsawy I looked into a unit test for ElasticSearch connection attempts, but all of the RSpec tests I could find related to query results rather than initialization.
> 
> One way around limitations of tests for initializers would be to have two distinct test invocations in CI, one using config with common URL syntax (and just one test?) and the rest as usual with default config and ENV.
> 
> Perhaps for now we can accept manual tests; in future library migration efforts we could then deprecate the parameterized initialization entirely.

yes @sonoransun you are right all my tests related to verify search result and I agreed that we should have test for ElasticSearch connection

melsawy (approved):

caiosba:
@sonoransun , you can merge if existing tests pass, and Sawy and I can take care of a test for that new type of connection.

sonoransun:
> @sonoransun , you can merge if existing tests pass, and Sawy and I can take care of a test for that new type of connection.

Ok, rebased against latest develop. We can address connection tests in initializers separately. Thank you!

PR #1477, opened 2023-04-12
Install bundler in production Dockerfile

hartsick:
Previously we weren't installing bundler in the production Dockerfile, which caused a problem building the image when we upgraded to Ruby 2.7.7. This adds bundler back into the Dockerfile, configures Bundler similarly to how we configure it in the development Dockerfile, and also formats the dependency list (installed via apt-get) so that it's easier to compare against the development Dockerfile--and adds git, which we didn't have installed in the prod Dockerfile, but were seeing warnings about at deploy time.

CV2-2669

### DIFF
--- a/production/Dockerfile
+++ b/production/Dockerfile
@@ -33,10 +33,22 @@ RUN apt-get update -qq && apt-get install -y curl
 #
 RUN curl -sL https://deb.nodesource.com/setup_12.x | bash -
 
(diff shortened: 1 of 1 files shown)


hartsick:
I'm going to merge this because there's nothing the tests would touch relating to the code changed anyway—just gotta see it deploy!

PR #1489, opened 2023-04-28
Remove code relating to benchmark, and fix deploys

hartsick:
With the Rails 6.0 upgrade, we ran into an issue where our production Dockerfile was requiring a SQLite version that was not matched by what was in our Gemfile.lock (1.13). Rather than get it to pass, I removed the workbench-related code, since the service is no longer used. That and asset precompilation, which we also no longer use and is removed in this commit, were the only places where we required SQLite.

This should fix deploys.

CV2-2669


sonoransun (commented):
💯

PR #1490, opened 2023-04-28
Enable select Rails 6 defaults

hartsick:
Now that we have seen the Rails 6.0 upgrade deployed and relatively stable, we move toward using the configuration files generated with `bin/rails app:update` on 6.0 and select defaults (there are two defaults that I did not enable in `config/initializers/new_framework_defaults_6_0.rb`, because it suggested waiting until you've seen Rails 6.0 stability in production for a bit).

For the most part this PR is auto-generated. When reviewing, I'd particularly like an eye on:
* any modifications or deletions that I may have made of configurations we've added or modified that we still need to keep
* any new configurations that are not commented out that we're not sure about
* any commented configurations we might want to consider using in the future (the content security policy is one that I'm curious about, for example)


config/boot.rb
```diff
@@ -1,6 +1,4 @@
 ENV['BUNDLE_GEMFILE'] ||= File.expand_path('../Gemfile', __dir__)
-# https://github.com/Shopify/bootsnap/issues/110
-ENV['BOOTSNAP_CACHE_DIR'] = File.expand_path("../tmp/cache#{ENV['TEST_ENV_NUMBER']}", __dir__) if ENV['RAILS_ENV'] == 'test'
```
caiosba, line 3: This was required to fix a common flakyness issue in Check API tests... it's not needed anymore because `bootsnap` was upgraded as per change above?

config/cable.yml
```diff
@@ -0,0 +1,10 @@
+development:
+  adapter: async
+
+test:
+  adapter: test
+
+production:
+  adapter: redis
+  url: <%= ENV.fetch("REDIS_URL") { "redis://localhost:6379/1" } %>
+  channel_prefix: check_production
```
caiosba, line 10: We don't use websockets, do we need this file?

config/environments/test.rb
```diff
@@ -21,6 +22,7 @@
   # Show full error reports and disable caching.
   config.consider_all_requests_local       = true
   config.action_controller.perform_caching = false
+  config.cache_store = :null_store
```
caiosba, line 25: I'm curious if this will affect tests that rely on cache (for example, tests for the cached fields)... but anyway, if tests pass, this is good to go!

config/initializers/new_framework_defaults_6_0.rb
```diff
@@ -0,0 +1,45 @@
...
+#
+# This option is not backwards compatible with earlier Rails versions.
+# It's best enabled when your entire app is migrated and stable on 6.0.
+# Rails.application.config.action_dispatch.use_cookies_with_metadata = true
+
+# Change the return value of `ActionDispatch::Response#content_type` to Content-Type header without modification.
+Rails.application.config.action_dispatch.return_only_media_type_on_content_type = false
+
+# Return false instead of self when enqueuing is aborted from a callback.
+Rails.application.config.active_job.return_false_on_aborted_enqueue = true
```
caiosba, line 23: Since we are using Sidekiq, I guess this line doesn't make any difference?

config/initializers/new_framework_defaults_6_0.rb
```diff
@@ -0,0 +1,45 @@
...
+
+# Send Active Storage analysis and purge jobs to dedicated queues.
+Rails.application.config.active_storage.queues.analysis = :active_storage_analysis
+Rails.application.config.active_storage.queues.purge    = :active_storage_purge
+
+# When assigning to a collection of attachments declared via `has_many_attached`, replace existing
+# attachments instead of appending. Use #attach to add new attachments without replacing existing ones.
+Rails.application.config.active_storage.replace_on_assign_to_many = true
```
caiosba, line 31: Since we're using `carrierwave` for uploads these lines don't make any difference, right?

config/storage.yml
```diff
@@ -0,0 +1,34 @@
...
+# microsoft:
+#   service: AzureStorage
+#   storage_account_name: your_account_name
+#   storage_access_key: <%= Rails.application.credentials.dig(:azure_storage, :storage_access_key) %>
+#   container: your_container_name
+
+# mirror:
+#   service: Mirror
+#   primary: local
+#   mirrors: [ amazon, google, microsoft ]
```
caiosba, line 34: Do we need this file since we're using `carrierwave` to handle uploads?

db/migrate/20230428213431_add_foreign_key_constraint_to_active_storage_attachments_for_blob_id.active_storage.rb
```diff
@@ -0,0 +1,10 @@
+# This migration comes from active_storage (originally 20180723000244)
+class AddForeignKeyConstraintToActiveStorageAttachmentsForBlobId < ActiveRecord::Migration[6.0]
+  def up
+    return if foreign_key_exists?(:active_storage_attachments, column: :blob_id)
+
+    if table_exists?(:active_storage_blobs)
+      add_foreign_key :active_storage_attachments, :active_storage_blobs, column: :blob_id
+    end
+  end
+end
```
caiosba, line 10: Do we need this migration since we're using `carrierwave` to handle uploads?

caiosba (commented):
I left a few questions.

config/boot.rb
hartsick: yeah, that's my understanding based on what's described in the issue in the comment!

config/cable.yml
hartsick: this is one of the things that came out of reading `rake app:update`. I don't believe we need it explicitly (the Rails 6.0 branch that I had already merged passed without it), but we do have actioncable included as one of our rails dependencies in the gemfile.lock so I erred toward including it as an example in future, which is same approach I took with other autogenerated config, partly to reduce the diff we have to manage in future. I think I'd like to keep the file, but comment it out to reduce confusion. what do you think?

config/initializers/new_framework_defaults_6_0.rb
hartsick: believe it shouldn't - I'm just trying to enable everything so that we can test it all before changing application.rb to use 6.0 defaults, which will include this

config/cable.yml
caiosba: Yes, add but comment sounds good to me! I prefer to comment than generating confusion in the future, specially that it references an environment variable.

config/initializers/new_framework_defaults_6_0.rb
hartsick: same comment as above - believe it shouldn't make a difference

config/initializers/new_framework_defaults_6_0.rb
caiosba: Sounds good

config/initializers/new_framework_defaults_6_0.rb
caiosba: Sounds good

config/storage.yml
caiosba: As above - let's the keep the default files but commented.

hartsick: same as what I said in a previous comment - I had taken the approach of including these auto-generated files 1) so that it reduces diff when we run app:update in future (based on my understanding of how it will diff), and 2) so we have examples if we want to move to eg activestorage in future. I think I have a slight preference toward keeping the file but keeping things that aren't needed commented out, if that doesn't break things.

caiosba: Agree!

db/migrate/20230428213431_add_foreign_key_constraint_to_active_storage_attachments_for_blob_id.active_storage.rb
caiosba: But let's keep it to avoid less conflicts in the future.

hartsick: it's a no-op, but I wanted to keep all autogenerated files from the update task for consistency

caiosba (approved):

config/cable.yml
hartsick: yeah that makes total sense! I'll update

melsawy (approved):

PR #1494, opened 2023-04-30
New tipline newsletter

caiosba:
This is a big pull request for all the backend work required for the new newsletters feature.

- Previous newsletter logic was deleted
- Added models `TiplineNewsletter` and `TiplineNewsletterDelivery` and related logic, permissions, GraphQL types and mutations
- Newsletters can be "static" (just static content) or pull content from an RSS feed, but for both cases, they can have up to 3 articles (plus introduction)
- Newsletters are unique for a given team and language
- Static newsletters can be sent at a specific scheduled date/time, and once they are sent, they are unscheduled ("disabled")
- RSS newsletters can be sent at different days (more than one) and a fixed time, and they will remain scheduled ("enabled") until the content hasn't changed between two deliveries
- Newsletters can have different header types, 3 of which expect an uploaded file: image, audio and video
- Logic (mostly validation and conversion) related to the different newsletter media header types (image, audio and video) are handled by specific concern files
- WhatsApp doesn't support audio headers, so the application converts audio to video with a cover image
- There is a Sidekiq worker (actually, refactored from previous version) that sends a newsletter and saves a `TiplineNewsletterDelivery` for it... instances of that model are used for the statistics calculation "number of newsletters sent"

- Added new settings fields for `Team`: `outgoing_urls_utm_code` and `shorten_outgoing_urls`
- Updated WhatsApp message template code to handle media header types
- Refactored the way URLs were "UTMized", by abstracting that to a `UrlRewriter` class, which handle UTM codes and URL shortening for all outgoing tipline URLs
- Refactored how RSS feeds are handled, by abstracting the logic to a `RssFeed` class


app/graph/mutations/tipline_newsletter_mutations.rb
```diff
@@ -0,0 +1,46 @@
...
+    footer: 'str'
+  })
+
+  # Schedule
+  fields.merge!({
+    send_every: 'json',
+    send_on: 'str',
+    timezone: 'str',
+    time: 'str'
+  })
```
melsawy, line 43: I think it's easier to define all fields at same time instead of using `.merge!` same as other mutations

app/models/team.rb
```diff
@@ -587,6 +595,30 @@ def get_feed(feed_id)
...
+    tbi = TeamBotInstallation.where(team_id: self.id, user_id: BotUser.smooch_user&.id.to_i).last
+    unless tbi.nil?
+      ['none', 'image', 'video', 'audio', 'link_preview'].each do |header_type|
+        mapped_header_type = {
+          'none' => 'none',
+          'image' => 'image',
+          'video' => 'video',
+          'audio' => 'video', # WhatsApp doesn't support audio header, so we convert it to video
+          'link_preview' => 'none'
+        }[header_type]
```
melsawy, line 610: You can move `mapped_header_type` array before `.each` and replace this block by `mapped_header_type[header_type]`

app/models/tipline_newsletter.rb
```diff
@@ -0,0 +1,247 @@
+class TiplineNewsletter < ApplicationRecord
+  SCHEDULE_DAYS = ['sunday', 'monday', 'tuesday', 'wednesday', 'thursday', 'friday', 'saturday']
+
+  include TiplineNewsletterImage
+  include TiplineNewsletterVideo
+  include TiplineNewsletterAudio
+
+  belongs_to :team
+  has_many :tipline_newsletter_deliveries
+  has_paper_trail on: [:create, :destroy], save_changes: true, ignore: [:updated_at, :created_at], versions: { class_name: 'Version' }
```
melsawy, line 10: You can add `if: proc { |_x| User.current.present? }`  condition to paper trail to make sure version created with `whodunnit` user

melsawy (changes requested):
Great work Caio, I added a few comments

app/models/tipline_newsletter.rb
caiosba: Yes, good idea - it also made me realize that I need to track changes for the `:update` event as well.

app/graph/mutations/tipline_newsletter_mutations.rb
caiosba: Hum... for big mutations, don't you think that it's more readable that way?

app/workers/tipline_newsletter_worker.rb
```diff
@@ -3,46 +3,71 @@ class TiplineNewsletterWorker
...
+        log team_id, language, "Newsletter sent to subscriber ##{ts.id}, response: (#{response.code}) #{response.body}"
+        count += 1
+      rescue StandardError => e
+        log team_id, language, "Could not send newsletter to subscriber ##{ts.id}: #{e.message}"
       end
     end
+    finish = Time.now
+
+    # Save a delivery event for this newsletter
+    TiplineNewsletterDelivery.create!({
```
hartsick, line 55: **question** this will raise an exception if creation fails for some reason, causing the job to get re-run from the beginning. is that intended and if so is this task idempotent?

lib/check_statistics.rb
```diff
@@ -55,7 +55,12 @@ def number_of_newsletters_sent(team_id, start_date, end_date, language)
           nil
         end
       end.reject{ |t| t.blank? }.collect{ |t| Time.parse(t.to_s) }.select{ |t| t >= start_date && t <= end_date }.collect{ |t| t.to_s }.uniq
-      times.size
+      old_count = times.size
+
+      newsletter = TiplineNewsletter.where(team_id: team_id, language: language).last
```
hartsick, line 60: **praise** soooo much simpler

**question** we should only have one month where we expect to tally the old + new (this one), right? other than that, we'd expect either old_count (if we're regenerating past months) or new_count (for anything after, say, june)?

lib/rss_feed.rb
```diff
@@ -0,0 +1,30 @@
+class RssFeed
+  require 'rss'
```
hartsick, line 2: **note** mostly a note to myself - rss is not included in the standard library for Ruby 3.0, and needs to be added to the Gemfile. I had fixed this in the previous location but will need to update. ref https://github.com/meedan/check-api/pull/1511

lib/rss_feed.rb
hartsick, line 1: **praise** I love this class

lib/url_rewriter.rb
```diff
@@ -0,0 +1,26 @@
...
+      uri.to_s
+    rescue
+      url
+    end
+  end
+
+  def self.shorten_and_utmize_urls(text, source = nil, owner = nil)
+    entities = Twitter::TwitterText::Extractor.extract_urls_with_indices(text, extract_url_without_protocol: true)
+    # Ruby 2.7 freezes the empty string from nil.to_s, which causes an error within the rewriter
+    Twitter::TwitterText::Rewriter.rewrite_entities(text || '', entities) do |entity, _codepoints|
```
hartsick, line 21: **note** Flagging for me to check this out in the Ruby 3.0 branch when it's on develop https://github.com/meedan/check-api/pull/1511

lib/url_rewriter.rb
hartsick, line 1: 🎉

test/lib/rss_feed_test.rb
hartsick, line 1: 🥹

test/workers/tipline_newsletter_worker_test.rb
```diff
@@ -18,44 +18,60 @@ def setup
   end
 
   test "should not crash if error happens when sending newsletter to some subscriber" do
-    Bot::Smooch.stubs(:send_final_messages_to_user).raises(StandardError)
+    Bot::Smooch.stubs(:send_message_to_user).raises(StandardError)
     assert_nothing_raised do
       TiplineNewsletterWorker.perform_async(@team.id, 'en')
     end
-    Bot::Smooch.unstub(:send_final_messages_to_user)
+    Bot::Smooch.unstub(:send_message_to_user)
```
hartsick, line 25: **info** we shouldn't need this line anymore since we're tearing down stubs at the end of every test. I've just been ripping out the lines in tests when I touch the test or file

hartsick (approved):
Caio, this looks so great!! Everything is looking a lot simpler. I left a few questions and comments, but I wouldn't consider any of them are blocking. The tests are 😗👌🏼

app/models/ability.rb
```diff
@@ -83,7 +83,7 @@ def editor_perms
     can [:cud], DynamicAnnotation::Field do |obj|
       obj.annotation.team&.id == @context_team.id
     end
-    can [:cud], [Account, Source], :team_id => @context_team.id
```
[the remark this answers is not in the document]
caiosba: `:cud` is a shortcut for `[:create, :update, :destroy]` :D

app/models/concerns/smooch_capi.rb
```diff
@@ -223,33 +223,32 @@ def capi_send_message_to_user(uid, text, extra, _force = false)
...
         # FIXME: Understand different errors that can be returned from Cloud API and have specific reports
         # response_body = self.safely_parse_response_body(response)
         e = Bot::Smooch::CapiMessageDeliveryError.new('Could not send message using WhatsApp Cloud API')
         CheckSentry.notify(e,
           uid: uid,
           type: payload.dig(:type),
           template_name: payload.dig(:template, :name),
-          template_language: payload.dig(:template, :language, :code)
+          template_language: payload.dig(:template, :language, :code),
+          error: response.body
```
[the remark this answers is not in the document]
caiosba: We don't know for CAPI yet because we're not using it on live yet... but I have left a comment on line 227 above to revisit it in the future once we know more about what can be returned:
```
FIXME: Understand different errors that can be returned from Cloud API and have specific reports
```

app/models/concerns/smooch_capi.rb
```diff
@@ -223,33 +223,32 @@ def capi_send_message_to_user(uid, text, extra, _force = false)
...
         )
       end
-      ret
+      response
     end
 
-    def capi_format_template_message(_namespace, template, _fallback, locale, image, placeholders)
+    def capi_format_template_message(_namespace, template, _fallback, locale, file_url, placeholders, file_type = 'image', preview_url = true)
```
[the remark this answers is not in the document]
caiosba: No, not used as a background job... the unused parameters situation is because I wanted to keep the same method signature for all different implementations:

- Smooch
- Cloud API
- On-premises API

But not all of them use all the parameters.

app/models/concerns/smooch_newsletter.rb
```diff
@@ -4,42 +4,6 @@ module SmoochNewsletter
   extend ActiveSupport::Concern
 
   module ClassMethods
-    TeamBotInstallation.class_eval do
```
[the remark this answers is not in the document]
caiosba: Hahaha

app/models/concerns/tipline_newsletter_audio.rb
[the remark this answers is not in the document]
caiosba: Thanks!

app/models/concerns/tipline_newsletter_image.rb
```diff
@@ -0,0 +1,49 @@
...
+  def should_convert_header_image?
+    # New file or new overlay text
+    self.header_type == 'image' && (self.new_file_uploaded? || self.previous_changes.keys.include?('header_overlay_text'))
+  end
+
+  def convert_header_file_image
+    return self.header_file_url if self.header_overlay_text.blank?
+
+    # Get the template and generate the HTML
+    FileUtils.mkdir_p(File.join(Rails.root, 'public', 'newsletter'))
```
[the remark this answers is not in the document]
caiosba: In some cases, to be used by external applications... for this case, just for simplicity: we know that permissions are set for that directory, there will be no conflicts with other files and no processing for assets pipeline.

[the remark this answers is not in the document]
caiosba: But also, we just do that with content that is OK to be public.

app/models/concerns/tipline_newsletter_image.rb
```diff
@@ -0,0 +1,49 @@
...
+    temp_url = CheckS3.public_url(path)
+
+    # Convert the HTML to PNG
+    uri = URI("#{CheckConfig.get('narcissus_url')}/?url=#{temp_url}&selector=%23frame")
+    request = Net::HTTP::Get.new(uri)
+    request['x-api-key'] = CheckConfig.get('narcissus_token')
+    response = Net::HTTP.start(uri.hostname, uri.port, use_ssl: uri.scheme == 'https') { |http| http.request(request) }
+    screenshot = JSON.parse(response.body)['url']
+    raise "Unexpected response from Narcissus for #{uri}: #{response.body}" unless screenshot =~ /^http/
+    FileUtils.rm_f temp
```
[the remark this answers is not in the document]
caiosba: Yes, good point!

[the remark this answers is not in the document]
caiosba: Done!

app/models/team.rb
```diff
@@ -587,6 +595,31 @@ def get_feed(feed_id)
...
+  def available_newsletter_header_types
+    header_type_mapping = {
+      'none' => 'none',
+      'image' => 'image',
+      'video' => 'video',
+      'audio' => 'video', # WhatsApp doesn't support audio header, so we convert it to video
+      'link_preview' => 'none'
+    }
+    available = []
+    tbi = TeamBotInstallation.where(team_id: self.id, user_id: BotUser.smooch_user&.id.to_i).last
```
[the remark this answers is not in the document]
caiosba: Yes, we want it as zero because then there will be no `TeamBotInstallation` and the job will exit anyway.

app/models/team.rb
```diff
@@ -587,6 +595,31 @@ def get_feed(feed_id)
...
+    unless tbi.nil?
+      ['none', 'image', 'video', 'audio', 'link_preview'].each do |header_type|
+        mapped_header_type = header_type_mapping[header_type]
+        if !tbi.send("get_smooch_template_name_for_newsletter_#{mapped_header_type}_no_articles").blank? &&
```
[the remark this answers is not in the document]
caiosba: Yes, that's a good point! And I think we should revisit that in many places of the application, because it's confusing and makes it hard to debug. In this case here, these methods are just shortcuts for settings keys... so, `object.get_foo_bar` is the same as `object.settings['foo_bar']`. I could replace by the latter for simplicity?

app/models/tipline_newsletter.rb
```diff
@@ -0,0 +1,250 @@
...
+    TiplineSubscription.where(team_id: self.team_id, language: self.language).count
+  end
+
+  def last_scheduled_by
+    User.find_by_id(self.last_scheduled_by_id)
+  end
+
+  def whatsapp_template_name
+    number = ['no', 'one', 'two', 'three'][self.articles.size]
+    type = {
```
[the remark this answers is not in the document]
caiosba: Yes, good idea!

app/models/tipline_newsletter.rb
```diff
@@ -0,0 +1,250 @@
...
+  end
+
+  def body
+    self.articles.join("\n\n")
+  end
+
+  def format_as_template_message
+    date = I18n.l(Time.now.to_date, locale: self.language.to_s.tr('_', '-'), format: :long)
+    file_url = ['image', 'audio', 'video'].include?(self.header_type) ? self.header_media_url : nil
+    file_type = { 'image' => 'image', 'video' => 'video', 'audio' => 'video' }[self.header_type]
```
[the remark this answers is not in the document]
caiosba: Yes! I changed it to use the same constant that I implemented based on your other two comments related to that.

app/models/tipline_newsletter.rb
```diff
@@ -0,0 +1,250 @@
...
+    now = Time.now.to_i
+    input = File.open(File.join(Rails.root, 'tmp', "newsletter-#{type}-input-#{self.id}-#{now}"), 'wb')
+    input.puts(URI(self.header_file_url).open.read)
+    input.close
+    output_path = File.join(Rails.root, 'tmp', "newsletter-#{type}-output-#{self.id}-#{now}.mp4")
+    video = FFMPEG::Movie.new(input.path)
+    video.transcode(output_path, options)
+    path = "newsletter/video/newsletter-#{type}-#{self.id}-#{now}"
+    CheckS3.write(path, 'video/mp4', File.read(output_path))
+    FileUtils.rm_f input.path
```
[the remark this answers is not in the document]
caiosba: Yes, good point! Done!

app/workers/tipline_newsletter_worker.rb
caiosba: The task is configured to not be retried (some users might have received the newsletter already, and we don't want to resend)... I just changed to `create!` and to notify Sentry if this event can't be saved.

lib/check_statistics.rb
caiosba: Correct! We can clean it up later.

test/workers/tipline_newsletter_worker_test.rb
caiosba: Yes! Sometimes I still forget hehe

db/schema.rb
```diff
@@ -265,7 +265,7 @@
     t.datetime "updated_at", null: false
   end
 
-  create_table "dynamic_annotation_fields", id: :serial, force: :cascade do |t|
+  create_table "dynamic_annotation_fields", id: :integer, default: -> { "nextval('new_dynamic_annotation_fields_id_seq'::regclass)" }, force: :cascade do |t|
```
hartsick, line 268: @caiosba I just rebased my branch off of develop and am running into an issue on this line when I try to use `bin/rails db:test:prepare`. I had missed it during review because I had looked at the migrations but not schema changes. do you know why this line changed, or what `nextval('new_dynamic_annotation_fields_id_seq'::regclass)` references?

caiosba: @hartsick I don't know! I didn't change it manually... I wonder if it was generated by some older version I was running. Feel free to overwrite it with a newer version.

hartsick: gotcha! I wonder if it's an old change hanging around in your local database? I now see that function is referenced in `[doc/structure.sql](https://github.com/meedan/check-api/tree/d0dc4cab212c62194c66d64d171d577b1293412d/doc/structure.sql)`. I'll see if I can change it back

caiosba: ah yes, probably old data I have here...

hartsick: it looks like some indexes below also got removed on this table - annotation_id, annotation, and value. that was also unintentional, right? just checking before I re-add

caiosba: > it looks like some indexes below also got removed on this table - annotation_id, annotation, and value. that was also unintentional, right? just checking before I re-add

Yes, it was all unintentional.

PR #1496, opened 2023-05-01
CV2-3052 add another patch for slash-delimited models

DGaffney:


melsawy (approved):
Looks good to me  but please fix code climate issues before merge to develop branch

PR #1498, opened 2023-05-02
AutoTagger bot

computermacgyver:
Implements an autotagger bot per CV2-3062.

Settings for the bot are in the corresponding check-web branch: https://github.com/meedan/check-web/tree/CV2-3062-tagger


app/models/bot/tagger.rb
```diff
@@ -0,0 +1,72 @@
+require 'json'
+
+class Bot::Tagger < BotUser
+  check_settings
+  class Error < ::StandardError
+  end
+
+  def self.get_tag_text(tag_id,auto_tag_prefix,ignore_autotags)
+    tag=TagText.find(tag_id).text
```
caiosba, line 9: This will crash (and we'll get a Sentry alert) if there is no `TagText` with this ID. Is this the intended behavior? We can simply ignore it:

```
tag = TagText.find_by_id(tag_id)
return nil if tag.nil?
tag = tag.text
```

caiosba: `Model.find(id)` -> raises an exception if object doesn't exist
`Model.find_by_id(id)` -> returns nil if object doesn't exist

app/models/bot/tagger.rb
```diff
@@ -0,0 +1,72 @@
...
+    if ignore_autotags &&  tag[0]==auto_tag_prefix
```
caiosba, line 10: There are two spaces after `&&`

app/models/bot/tagger.rb
```diff
@@ -0,0 +1,72 @@
...
+    pm = nil
+    begin
+      settings=JSON.parse(body["settings"])
+      auto_tag_prefix=settings["auto_tag_prefix"]
+      threshold=settings["threshold"]/100.0
+      ignore_autotags=settings["ignore_autotags"]
+      pm = ProjectMedia.where(id: body.dig(:data, :dbid)).last
+      if body.dig(:event) == 'create_project_media' && !pm.nil?
+        Rails.logger.info("[AutoTagger Bot] [ProjectMedia ##{pm.id}] This item was just created, processing...") 
+        #results=Bot::Alegre.get_merged_similar_items(pm, [{ value: threshold }], Bot::Alegre::ALL_TEXT_SIMILARITY_FIELDS, pm.title)
```
caiosba, line 34: Do we need to keep this comment?

app/models/bot/tagger.rb
```diff
@@ -0,0 +1,72 @@
...
+        results=Bot::Alegre.get_items_with_similar_text(pm, Bot::Alegre::ALL_TEXT_SIMILARITY_FIELDS, 
+          [{ value: threshold }], pm.title, [Bot::Alegre.matching_model_to_use(pm.team_id)].flatten.reject{|m| m==Bot::Alegre::ELASTICSEARCH_MODEL})
+        #get_items_with_similar_text(pm, fields, threshold, text, models = nil, team_ids = [pm&.team_id])
```
caiosba, line 37: Do we need to keep this comment?

app/models/bot/tagger.rb
```diff
@@ -0,0 +1,72 @@
...
+        Rails.logger.info("[AutoTagger Bot] [ProjectMedia ##{pm.id}] #{results.length} nearest neighbors #{results.keys()}")
+        Rails.logger.info("[AutoTagger Bot] [ProjectMedia ##{pm.id}] Results: #{results}")
+        tag_counts=results.map{|nn_pm,_| ProjectMedia.find(nn_pm).get_annotations('tag')}.flatten.map{|t| self.get_tag_text(t[:data][:tag],auto_tag_prefix,ignore_autotags)}.group_by(&:itself).transform_values(&:count)
+        tag_counts=tag_counts.reject{|t|t==nil}
+        tag_counts=tag_counts.sort_by{|k,v| v}
```
caiosba, line 42: This could be a one liner: `tag_counts = tag_counts.reject{ |t| t.blank? }.sort_by{ |k, v| v }`

app/models/bot/tagger.rb
```diff
@@ -0,0 +1,72 @@
...
+          Rails.logger.info("[AutoTagger Bot] [ProjectMedia ##{pm.id}] No most common tag")
+        end
+        handled = true
+      end
+    rescue StandardError => e
+      Rails.logger.error("[AutoTagger Bot] Exception for event `#{body['event']}`: #{e.message}")
+      self.notify_error(e, { bot: self.name, body: body }, RequestStore[:request])
+    end
+
+    #self.unarchive_if_archived(pm)
```
caiosba, line 67: I guess we can remove this line?

caiosba (changes requested):
I added a few comments, but looks good overall! It's amazing how much you can do without even knowing Ruby/Rails, Scott :)

One more thing that I didn't want to comment in many lines... since we don't have a linter (our bad, we need to fix it), please be sure that there is one space between operators and operands... for example, `A < B` instead of `A<B`.

We also need some unit tests - but we don't need to use your time for that, Scott... someone else can do it.

app/models/bot/tagger.rb
computermacgyver: Incorporated into the next if

```
    tag=TagText.find_by_id(tag_id).text
    if tag.nil? || (ignore_autotags && tag[0]==auto_tag_prefix)
      tag=nil
```

app/models/bot/tagger.rb
```diff
@@ -0,0 +1,72 @@
...
+          max_count=tag_counts.last[1]
+          if max_count<settings["minimum_count"]
+            Rails.logger.info("[AutoTagger Bot] [ProjectMedia ##{pm.id}] Max count #{max_count} is less than minimum required to apply a tag")
+            return false
+          end
+          most_common_tags=tag_counts.reject{|k,v| v<max_count}
+          Rails.logger.info("[AutoTagger Bot] [ProjectMedia ##{pm.id}] Most common tags #{most_common_tags}")
+          most_common_tags.each do |tag|
+            #TODO: Set the user to Bot.get_user('tagger')
+            Tag.create(annotated:pm,tag: auto_tag_prefix+tag[0])
```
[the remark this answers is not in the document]
computermacgyver: Done also reminded me I wanted to set the `annotator` property. Thanks.

app/models/bot/tagger.rb
caiosba: That method takes the item out of the trash, just to make it visible again.

computermacgyver: Thanks. I thought that was the case. I just wasn't 100% sure why Alegre invokes it. In any case, I think it's fine to leave for now and it doesn't apply to this PR. Thanks!

app/models/bot/tagger.rb
```diff
@@ -0,0 +1,65 @@
...
+
+class Bot::Tagger < BotUser
+  check_settings
+  class Error < ::StandardError
+  end
+
+  def self.get_tag_text(tag_id,auto_tag_prefix,ignore_autotags)
+    tag=TagText.find_by_id(tag_id).text
+    if tag.nil? || (ignore_autotags && tag[0]==auto_tag_prefix)
+      tag=nil
```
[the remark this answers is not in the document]
caiosba: Ah, it's because in Ruby the last line is like the `return`. So `tag = nil` could be simply rewritten as `nil`.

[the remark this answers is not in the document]
computermacgyver: ah, that is true. I didn't think of it that way. I was thinking it was complaining about 
```if tag.nil?
      tag = nil
   ...
```
as being useless, which it would be if the if statement didn't have the second part after the boolean or. I'll update. 

Codeclimate wins this time, but watch out :eye:  :joy:

hartsick:
@computermacgyver hi Scott! is this ready for re-review? wanted to check since it's been hanging out for a little bit

computermacgyver:
I think this is good to go except tests. The one outstanding comment from @caiosba is

> We also need some unit tests - but we don't need to use your time for that, Scott... someone else can do it.

I have a note on ML call to discuss, but if you're aware of anyone with capacity to take on the tests that would be great. Otherwise, I'll make then a side project for me for next week.

caiosba:
@computermacgyver , after @hartsick 's review, we can schedule a pairing session to work on some unit tests... I think that 4 unit tests are enough to cover all cases.

db/migrate/20230429100703_create_auto_tagger_bot.rb
```diff
@@ -0,0 +1,37 @@
...
+    RequestStore.store[:skip_notifications] = true
+
+    meedan_team = Team.where(slug: 'meedan').last || Team.new(name: 'Meedan', slug: 'meedan')
+    meedan_team.skip_notifications = true
+    meedan_team.skip_clear_cache = true
+    meedan_team.skip_check_ability = true
+    meedan_team.save!
+
+    Team.current = meedan_team
+    tb = BotUser.new
```
hartsick, line 12: **thought:** I'm not sure how often we make botusers for the application, but given the amount of setup here it seems like it would be nice to add a constructor with any needed validations if we expect to do it again.

db/migrate/20230429100703_create_auto_tagger_bot.rb
```diff
@@ -0,0 +1,37 @@
+class CreateAutoTaggerBot < ActiveRecord::Migration[5.2]
```
hartsick, line 1: **issue** I think it's most appropriate for things in `db/migrate` as being modifications to the database schema. here are a few reasons for that:
1. at least in theory (I know this would be difficult for us in practice), we should be able to squash our migrations into one at any point, since they'll take longer and longer to run over time
2. our models can change over time, breaking past migrations when they involve them. it's a little weird to go back and change history by modifying migration files.
3. this isn't really an issue since we shouldn't be relying on database state for tests, but just wanted to point out since it can be confusing - data in migrations that is required for the application to function won't be available in tests because tests load the schema directly and therefore don't run migrations

here is a thoughtbot article that talks about some of the stuff above at more length: https://thoughtbot.com/blog/data-migrations-in-rails

**suggestion / question** for that reason, I think it would be good for us to move this to somewhere outside of migrations. I think that `lib/tasks/migrate` is intended to be for this sort of data task, but I'm not totally sure since it's not tied specifically to a database migration. @caiosba do you have any thoughts here?

**thought** also, for development, I think we should also set up all required data to run the application (like bots) outside of migrations. seeds are good for this for required things, and the thoughtbot article above suggests adding a supplemental task that runs any needed data tasks for development

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2023_04_20_165704) do
+ActiveRecord::Schema.define(version: 2023_04_29_100703) do
```
hartsick, line 13: **suggestion** just flagging I don't think we need to update the schema here if we take approach noted above

app/models/bot/tagger.rb
```diff
@@ -0,0 +1,65 @@
+require 'json'
+
+class Bot::Tagger < BotUser
+  check_settings
+  class Error < ::StandardError
```
hartsick, line 5: **question** we create a custom error here, but don't seem to use it. is that intended?

app/models/bot/tagger.rb
```diff
@@ -0,0 +1,65 @@
...
+          Rails.logger.info("[AutoTagger Bot] [ProjectMedia ##{pm.id}] Most common tags #{most_common_tags}")
+          most_common_tags.each do |tag|
+            Tag.create!(annotated:pm, annotator: BotUser.get_user('tagger'), tag: auto_tag_prefix+tag[0])
+          end
+        else
+          Rails.logger.info("[AutoTagger Bot] [ProjectMedia ##{pm.id}] No most common tag")
+        end
+        handled = true
+      end
+    rescue StandardError => e
```
hartsick, line 57: **suggestion** if we wanted to use the custom error for this class (which I think would be helpful, since otherwise it may be hard to differentiate from similar exceptions in other classes since we're just catching all errors that could arise in the block) then you can use it by doing something like:
```
error = Error.new(e)
Rails.logger.error("[AutoTagger Bot] Exception for event `#{body['event']}`: #{error.class} - #{error.message}")
CheckSentry.notify(error, bot: self.name, body: body)
```

app/models/bot/tagger.rb
```diff
@@ -0,0 +1,65 @@
...
+    handled = false
+    pm = nil
+    begin
+      settings=JSON.parse(body["settings"])
+      auto_tag_prefix=settings["auto_tag_prefix"]
+      threshold=settings["threshold"]/100.0
+      ignore_autotags=settings["ignore_autotags"]
+      pm = ProjectMedia.where(id: body.dig(:data, :dbid)).last
+      if body.dig(:event) == 'create_project_media' && !pm.nil?
+        Rails.logger.info("[AutoTagger Bot] [ProjectMedia ##{pm.id}] This item was just created, processing...")
```
hartsick, line 33: **suggestion** I think a private method here to make sure logging is consistent  and to reduce visual noise might be helpful - something like

```
def log(message, pm_id = nil, level = :info)
    prefix = "[AutoTagger Bot] "
    prefix += "[ProjectMedia ##{pm_id}] " if pm_id
    Rails.logger.log(level, "#{prefix} #{message}")
end
```

app/models/bot/tagger.rb
```diff
@@ -0,0 +1,65 @@
...
+        results=Bot::Alegre.get_items_with_similar_text(pm, Bot::Alegre::ALL_TEXT_SIMILARITY_FIELDS,
+          [{ value: threshold }], pm.title, [Bot::Alegre.matching_model_to_use(pm.team_id)].flatten.reject{|m| m==Bot::Alegre::ELASTICSEARCH_MODEL})
+        Rails.logger.info("[AutoTagger Bot] [ProjectMedia ##{pm.id}] #{results.length} nearest neighbors #{results.keys()}")
+        Rails.logger.info("[AutoTagger Bot] [ProjectMedia ##{pm.id}] Results: #{results}")
+        tag_counts=results.map{|nn_pm,_| ProjectMedia.find(nn_pm).get_annotations('tag')}.flatten.map{|t| self.get_tag_text(t[:data][:tag],auto_tag_prefix,ignore_autotags)}.group_by(&:itself).transform_values(&:count)
```
hartsick, line 38: **issue** this line has a lot of complexity - can it be broken down? a comment would also be helpful, I think

app/models/bot/tagger.rb
```diff
@@ -0,0 +1,65 @@
...
+        tag_counts=tag_counts.reject{|t|t==nil}.sort_by{|_k,v| v}
+        Rails.logger.info("[AutoTagger Bot] [ProjectMedia ##{pm.id}] Tag distribution #{tag_counts}")
+        if tag_counts.length > 0
+          max_count=tag_counts.last[1]
```
hartsick, line 42: **question** I'm a little confused by this - is tag_counts an array of arrays? if it's not (which I think it isn't, based on line 39) then `tag_counts.last` should give you the last item of the array, and `tag_counts.last[1]` might give you nil since you're accessing a non-existent index

app/models/bot/tagger.rb
```diff
@@ -0,0 +1,65 @@
...
+          if max_count<settings["minimum_count"]
```
hartsick, line 43: **question** do we know that minimum_count comes back as an integer from settings? I'm not familiar with how we handle types

app/models/bot/tagger.rb
```diff
@@ -0,0 +1,65 @@
+require 'json'
+
+class Bot::Tagger < BotUser
+  check_settings
+  class Error < ::StandardError
+  end
+
+  def self.get_tag_text(tag_id,auto_tag_prefix,ignore_autotags)
+    tag=TagText.find_by_id(tag_id).text
+    if tag.nil? || (ignore_autotags && tag[0]==auto_tag_prefix)
```
hartsick, line 10: **question** is the intention here to return nil if the tag can't be found? line 9 will raise an exception, skipping this and sending us to the rescue StandardError, if the tag isn't found. if you'd prefer to return nil without exception you can do `tag = TagText.find_by_id(tag_id)&.text` - the important part here being the safe operator (& before method call)

app/models/bot/tagger.rb
```diff
@@ -0,0 +1,65 @@
...
+    pm = nil
+    begin
+      settings=JSON.parse(body["settings"])
+      auto_tag_prefix=settings["auto_tag_prefix"]
+      threshold=settings["threshold"]/100.0
+      ignore_autotags=settings["ignore_autotags"]
+      pm = ProjectMedia.where(id: body.dig(:data, :dbid)).last
+      if body.dig(:event) == 'create_project_media' && !pm.nil?
+        Rails.logger.info("[AutoTagger Bot] [ProjectMedia ##{pm.id}] This item was just created, processing...")
+        results=Bot::Alegre.get_items_with_similar_text(pm, Bot::Alegre::ALL_TEXT_SIMILARITY_FIELDS,
```
hartsick, line 34: **issue** this line has a lot of complexity - can it be broken down? a comment would also be helpful, I think

app/models/bot_user.rb
```diff
@@ -315,7 +315,11 @@ def self.notify_bots(event, team_id, object_class, object_id, target_bot, reques
     return if object.nil? || team.nil?
     team.team_bot_installations.each do |team_bot_installation|
       bot = team_bot_installation.bot_user
-      bot.notify_about_event(event, object, team, team_bot_installation) if bot.subscribed_to?(event) && (target_bot.blank? || bot.id == target_bot.id)
+      begin
+        bot.notify_about_event(event, object, team, team_bot_installation) if bot.subscribed_to?(event) && (target_bot.blank? || bot.id == target_bot.id)
+      rescue StandardError => e
```
hartsick, line 320: **question** what is this rescue intending to catch here? I'm curious because right now we seem to be catching all of the errors arising from `run` in the auto-tagger bot above and not re-raising, so shouldn't be introducing any new errors at this level, I think

hartsick (changes requested):
I think the core of this makes sense to me! I left a bunch of notes and questions using [conventional comments](https://conventionalcomments.org/). The ones marked as "issue" are something that I think we should take care of before merging (and why I marked "request changes", but I'm open to being overruled on that - know this has been sitting out for a bit); others are food for thought. I looked more for Ruby and Rails implementation / maintainability than I did for logic when reviewing.

I agree with Caio that it's important to get unit tests around this. I think in the future it would be great to do them to drive development, to see if we can't break down some of the logic and make sure we're capturing edge cases both in implementation and in tests. Dev/research team pairing could be great for this, I think!

Also, I think it would be great for us to get a linter on the project, especially since we have folks writing across different languages. There are spacing inconsistencies and some Ruby-specific style stuff I don't think it makes sense for a human to standardize on their own - better to have a robot do it.

db/migrate/20230429100703_create_auto_tagger_bot.rb
caiosba: Yes, recently we've been putting those changes to data (not schema) in `lib/tasks/migrate`, but we don't run them automatically - we need to remember to run them as post-deployment tasks, manually. I still think it makes sense to put those in `lib/tasks/migrate`... we can improve the process later.

app/models/bot_user.rb
computermacgyver: This change is unrelated to Tagger. Tagger does catch all errors as you note, but other Bots do not. In particular, Alegre does not and an exception in Alegre was causing Tagger not to run in some instances. The goal of the change here is to ensure we run all bots that should run regardless of whether they all succeed.

app/models/bot/tagger.rb
computermacgyver: Was copied from AlegreBot without thinking :sweat: '.  Based on the comment below, we'll use this in the rescue blog of `run`. Thanks

app/models/bot/tagger.rb
computermacgyver: Agreed. Implemented.

app/models/bot/tagger.rb
computermacgyver: Done. Thanks.

app/models/bot/tagger.rb
computermacgyver: `tag_counts` is an array of arrays. For example, `[['nature', 1], ['sport', 2]]` . Thus `.last[1]` gives the count of the most numerous tag.  I agree the reject of line 39 seems strange in that case. I'll double check.

app/models/bot/tagger.rb
computermacgyver: Comments added.

db/migrate/20230429100703_create_auto_tagger_bot.rb
computermacgyver: Re-implemented this as a rake in lib/tasks/migrate

app/models/bot/tagger.rb
```diff
@@ -0,0 +1,82 @@
...
+    handled = false
+    pm = nil
+    begin
+      settings=JSON.parse(body[:settings])
+      auto_tag_prefix=settings["auto_tag_prefix"]
+      threshold=settings["threshold"]/100.0
+      ignore_autotags=settings["ignore_autotags"]
+      pm = ProjectMedia.where(id: body.dig(:data, :dbid)).last
+      if body.dig(:event) == 'create_project_media' && !pm.nil?
+        self.log("This item was just created, processing...", pm_id = pm.id, level = Logger::INFO)
```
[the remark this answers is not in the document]
computermacgyver: I don't understand why this would be flagged.

test/models/bot/tagger_test.rb
```diff
@@ -0,0 +1,117 @@
...
+
+  pm1 = create_project_media quote: "rain", team: @team
+  Tag.create!(annotated:pm1, annotator: BotUser.get_user('tagger'), tag: "nature")
+  pm2 = create_project_media quote: "snow", team: @team
+  Tag.create!(annotated:pm2, annotator: BotUser.get_user('tagger'), tag: @auto_tag_prefix+"nature")
+  pm3 = create_project_media quote: "soccer", team: @team
+  Tag.create!(annotated:pm3, annotator: BotUser.get_user('tagger'), tag: "sport")
+
+  #Stub alegre to return pm1, pm2, and pm3 for any request
+  Bot::Alegre.stubs(:get_items_with_similar_text).returns({pm1.id=>pm1,pm2.id=>pm2,pm3.id=>pm3})
```
computermacgyver, line 63: What's best practice here @caiosba ? Should I stub the AlegreBot method I use or the go all the way down to the :request_api method where AlegreBot would actually call the Alegre service

caiosba: @computermacgyver best practice is to stub the highest level you're not testing... so you don't need to go all way down to the low-level `request_api` method.

app/models/bot/tagger.rb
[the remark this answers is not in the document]
computermacgyver: Resolved. Arguments are passed by position, not name and hence `pm_id` is a useless throwaway variable (not the name of the argument)

app/models/bot/tagger.rb
computermacgyver: In all my testing `mimimum_count` is always stored as an integer.

test/models/bot/tagger_test.rb
```diff
@@ -0,0 +1,117 @@
...
+    # Get tag text ignoring autotags
+    assert_nil Bot::Tagger.get_tag_text(tag1[:data][:tag], @auto_tag_prefix, true)
+    assert_equal "test", Bot::Tagger.get_tag_text(tag2[:data][:tag], @auto_tag_prefix, true)
+
+    # Get tag text for invalid id
+    # Should return nil
+    assert_nil Bot::Tagger.get_tag_text(-999, @auto_tag_prefix, false)
+  end
+
+  test "should tag item" do
```
hartsick, line 53: **praise:** it's so great that you added comprehensive tests around this!
**suggestion** in future I would suggest breaking this into more granular tests so that we get more meaningful failures if it fails midway through, and we can see all that fails without having to rerun. I think the comments you added to mark different "shoulds" would make great test divisions since they imply different expectations for different setups.

lib/tasks/migrate/202306040000_add_tagger_bot_user.rake
hartsick, line 1: great move! just a reminder that I *think* we'll need to run a task manually for this on QA and Live after deployment, though I'm blanking on how. @caiosba could you help remind me? I was trying to find documentation, but couldn't locate it

app/models/bot/tagger.rb
```diff
@@ -0,0 +1,82 @@
...
+          [{ value: threshold }], pm.title, [Bot::Alegre.matching_model_to_use(pm.team_id)].flatten.reject{|m| m==Bot::Alegre::ELASTICSEARCH_MODEL})
+        self.log("#{results.length} nearest neighbors #{results.keys()}", pm.id, Logger::INFO)
+        self.log("Results: #{results}", pm.id, Logger::INFO)
+
+        # For each nearest neighbor, get the tags.
+        tag_counts=results.map{|nn_pm,_| ProjectMedia.find(nn_pm).get_annotations('tag')}.flatten
+        # Transform from tag objects to strings and counts
+        tag_counts=tag_counts.map{|t| self.get_tag_text(t[:data][:tag],auto_tag_prefix,ignore_autotags)}.group_by(&:itself).transform_values(&:count)
+        # Reject any nil tags
+        tag_counts=tag_counts.reject{|k,_v|k==nil}.sort_by{|_k,v| v}
```
hartsick, line 54: **note:** here's a fun ruby shortcut - instead of`.reject{|k,_v|k==nil}` you can do `.compact`

hartsick (approved):
looks good!! nice work, scott

lib/tasks/migrate/202306040000_add_tagger_bot_user.rake
caiosba: Yes, we run it manually... we write those in the tickets as post-deployment tasks.

app/models/bot/tagger.rb
computermacgyver: Nice. I'll check that out. Thank you.

test/models/bot/tagger_test.rb
computermacgyver: Yeah, I was trying to balance between 
* more granular tests and 
* not having repeated setup code / not having too many tests that could slow down CI given this is quite a small, periphery bot at least for now
At the time I thought it made sense to group the three cases here that all test one function with the same setup code, but I can see the value of breaking them apart too. Thank you!

lib/tasks/migrate/202306040000_add_tagger_bot_user.rake
computermacgyver: Yes, I needed to run it manually as a post deployment task in my testing. :pray: 
(`rake check:migrate:create_tagger_user`)

app/models/bot/tagger.rb
computermacgyver: Looks like I'd have to put `compact` before the `group_by` as it doesn't remove key-value pairs
```
irb(main):015:0> fruits=["apple","apple","banana",nil,"raspberry","raspberry"]
=> ["apple", "apple", "banana", nil, "raspberry", "raspberry"]
irb(main):016:0> counts=fruits.group_by(&:itself).transform_values(&:count)
=> {"apple"=>2, "banana"=>1, nil=>1, "raspberry"=>2}
irb(main):017:0> counts.reject{|k,_v|k==nil}
=> {"apple"=>2, "banana"=>1, "raspberry"=>2}  #AS CURRENTLY IMPLEMENTED
irb(main):018:0> counts.compact
=> {"apple"=>2, "banana"=>1, nil=>1, "raspberry"=>2} # DOESN'T WORK
irb(main):019:0> fruits.compact.group_by(&:itself).transform_values(&:count)
=> {"apple"=>2, "banana"=>1, "raspberry"=>2} #WORKS
```

PR #1499, opened 2023-05-02
Upgrade to Rails 6.1

hartsick:
This commit upgrades us to using Rails 6.1. Enabling the Rails 6.1 default configs will come in another PR: #1504.

Here are a few things done in these commits; for more detail see the commit messages below:
* Update Rails and dependencies to 6.1+
* Move classes required for app to an auto-loaded path `app/lib`, from `lib`. The intention is that this folder can be for any lib-like classes that our application needs to run. Any lib classes that our web application doesn't need (eg used for rake task, other non-core functionality) can be kept in `lib`.
* Move to new errors API, which prevents us from modifying the errors hash directly.
* Hardcode the cache_key in cluster.rb - it seems that Rails has begun using the inflection configuration for cache key, which requires us to hardcode since our cache key is project_media_counter (vs now autogenerated projects_media_counter)
* Update Rails.application.class.parent_name  -> Rails.application.class.module_parent_name, to reflect that Rails class parent is now a module
* Make sure our tests assert_raise as the first level when there are multiple assertion blocks. The reverse that we had previously causes tests to fail, because the exception is raised without being caught by assertion
* Run app:update and add autogenerated files


caiosba (approved):
Thanks, pretty straightforward! Just curious: Why in some places single quotes were replaced by double quotes?

hartsick:
> Thanks, pretty straightforward! Just curious: Why in some places single quotes were replaced by double quotes?

@caiosba That's what the autogenerated files output, and I decided to accept the changes since otherwise we were going to have to deal with ignoring them in the diff each time we generate them for upgrade. It was a lot of noise. I think rubocop may complain about it, though (and assume that's why they were single quoted to begin with), so we might want to explicitly ignore them.

caiosba:
> > Thanks, pretty straightforward! Just curious: Why in some places single quotes were replaced by double quotes?
> 
> @caiosba That's what the autogenerated files output, and I decided to accept the changes since otherwise we were going to have to deal with ignoring them in the diff each time we generate them for upgrade. It was a lot of noise. I think rubocop may complain about it, though (and assume that's why they were single quoted to begin with), so we might want to explicitly ignore them.

Gotcha! OK, let's keep it... Rubocop didn't complain (because this rule is not turned on I think).

config/environments/production.rb
```diff
@@ -1,3 +1,5 @@
+require "active_support/core_ext/integer/time"
+
 Rails.application.configure do
```
hartsick, line 3: Make sure to doublecheck this file after #1506 is merged, since needed edits are made there

PR #1501, opened 2023-05-02
Enable Rails 5.2 application defaults

hartsick:
Previously we had generated the file to migrate to 5.2 defaults, but had not enabled any of them. They seem to be no-ops for the most part for us, so I think we should be safe to enable them in full since we've been running stable on 5.2 for the past few years.

This also lets us move to Rails 6.0 and 6.1 defaults in near future.

CV2-2669

---

In reviewing this pull request, I'd love an eye on if we think any of the configurations in the now deleted 5.2 file will impact Check API. Even though the file is deleted, the configurations they specify will be enabled automatically with the following line: `config.load_defaults 5.2`. As far as I could tell, it seemed like these shouldn't impact our application—I wrote notes below about each configuration.

Integration test branch: https://github.com/meedan/check-web/pull/1427


config/initializers/new_framework_defaults_5_2.rb
```diff
@@ -1,38 +0,0 @@
...
-# Existing cookies will be converted on read then written with the new scheme.
-# Rails.application.config.action_dispatch.use_authenticated_cookie_encryption = true
-
-# Use AES-256-GCM authenticated encryption as default cipher for encrypting messages
-# instead of AES-256-CBC, when use_authenticated_message_encryption is set to true.
-# Rails.application.config.active_support.use_authenticated_message_encryption = true
-
-# Add default protection from forgery to ActionController::Base instead of in
-# ApplicationController.
-# Rails.application.config.action_controller.default_protect_from_forgery = true
```
hartsick, line 28: This shouldn't impact us since we configure the forgery protection in ApplicationController, and then use that as the base for each of our controllers in the application.

config/initializers/new_framework_defaults_5_2.rb
```diff
@@ -1,38 +0,0 @@
...
-
-# Store boolean values are in sqlite3 databases as 1 and 0 instead of 't' and
-# 'f' after migrating old data.
-# Rails.application.config.active_record.sqlite3.represent_boolean_as_integer = true
```
hartsick, line 32: no-op, we no longer use sqlite

config/initializers/new_framework_defaults_5_2.rb
```diff
@@ -1,38 +0,0 @@
...
-
-# Use SHA-1 instead of MD5 to generate non-sensitive digests, such as the ETag header.
-# Rails.application.config.active_support.use_sha1_digests = true
-
-# Make `form_with` generate id attributes for any generated HTML tags.
-# Rails.application.config.action_view.form_with_generates_ids = true
```
hartsick, line 38: no-op, we don't use any form_withs

config/initializers/new_framework_defaults_5_2.rb
```diff
@@ -1,38 +0,0 @@
...
-# Add default protection from forgery to ActionController::Base instead of in
-# ApplicationController.
-# Rails.application.config.action_controller.default_protect_from_forgery = true
-
-# Store boolean values are in sqlite3 databases as 1 and 0 instead of 't' and
-# 'f' after migrating old data.
-# Rails.application.config.active_record.sqlite3.represent_boolean_as_integer = true
-
-# Use SHA-1 instead of MD5 to generate non-sensitive digests, such as the ETag header.
-# Rails.application.config.active_support.use_sha1_digests = true
```
hartsick, line 35: honestly not sure of impact of this; figure tests will tell us

config/initializers/new_framework_defaults_5_2.rb
```diff
@@ -1,38 +0,0 @@
...
-#
-# This option is not backwards compatible with earlier Rails versions.
-# It's best enabled when your entire app is migrated and stable on 5.2.
-#
-# Existing cookies will be converted on read then written with the new scheme.
-# Rails.application.config.action_dispatch.use_authenticated_cookie_encryption = true
-
-# Use AES-256-GCM authenticated encryption as default cipher for encrypting messages
-# instead of AES-256-CBC, when use_authenticated_message_encryption is set to true.
-# Rails.application.config.active_support.use_authenticated_message_encryption = true
```
hartsick, line 24: not sure of impact of this; figure tests will tell us

config/initializers/new_framework_defaults_5_2.rb
```diff
@@ -1,38 +0,0 @@
...
-# Rails.application.config.active_record.cache_versioning = true
-
-# Use AES-256-GCM authenticated encryption for encrypted cookies.
-# Also, embed cookie expiry in signed or encrypted cookies for increased security.
-#
-# This option is not backwards compatible with earlier Rails versions.
-# It's best enabled when your entire app is migrated and stable on 5.2.
-#
-# Existing cookies will be converted on read then written with the new scheme.
-# Rails.application.config.action_dispatch.use_authenticated_cookie_encryption = true
```
hartsick, line 20: not sure of impact of this—I think worst case it might log users out of sessions on deploy, or if we have to roll it back it will expire their cookies and cause them to re-log in.

config/initializers/new_framework_defaults_5_2.rb
```diff
@@ -1,38 +0,0 @@
...
-#
-# This file contains migration options to ease your Rails 5.2 upgrade.
-#
-# Once upgraded flip defaults one by one to migrate to the new default.
-#
-# Read the Guide for Upgrading Ruby on Rails for more info on each option.
-
-# Make Active Record use stable #cache_key alongside new #cache_version method.
-# This is needed for recyclable cache keys.
-# Rails.application.config.active_record.cache_versioning = true
```
hartsick, line 11: don't think we'll see an impact of this

caiosba (approved):

PR #1504, opened 2023-05-03
Enable the rollback-able Rails 6.1 defaults

hartsick:
We don't expect these to have a noticeable impact, but enabling them will allow us to move to using Rails 6.1 defaults as a whole in the future.

CV2-2669


hartsick:
Note to self: this branch should be updated with the latest settings from #1499

caiosba (approved):

PR #1506, opened 2023-05-04
Update production.rb for use in deployed environments

hartsick:
Previously we were loading a Base64 encoded string of a previous production.rb for Live and QA configuration in deployed environments, meaning what we had committed to config/environments/production.rb was being ignored.

This commit updates our committed production.rb with what was currently on QA / live, and refactors shared config to application.rb and extracts an environment variable for differing config across live & QA.

Lastly, it removes the step in our deployment where we try to create the production.rb file from SSM. Instead, it will use what's in repo.

CV2-3055

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -78,7 +78,7 @@ module Check
         enable_starttls_auto: true
       }
     end
-    config.action_mailer.default_url_options = { host: ENV.fetch('smtp_default_url_host', cfg['smtp_default_url_host']) }
+    config.action_mailer.default_url_options = { host: ENV['smtp_default_url_host'] || cfg['smtp_default_url_host'] }
 
     # CORS config
     allowed_origins = ENV['allowed_origins'] || cfg['allowed_origins']
--- a/config/application.rb
+++ b/config/application.rb
@@ -40,6 +40,7 @@ module Check
 
     cfg = YAML.load_file("#{Rails.root}/config/config.yml")[Rails.env]
 
+    # i18n
     config.i18n.fallbacks = ['en']
     config.i18n.default_locale = 'en'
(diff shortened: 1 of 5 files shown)


config/application.rb
```diff
@@ -48,6 +49,18 @@ class Application < Rails::Application
...
 
+    # Cache configuration
+    sidekiq_config_for_env = File.join(Rails.root, 'config', "sidekiq-#{Rails.env}.yml")
+    sidekiq_config = File.exist?(sidekiq_config_for_env) ? sidekiq_config_for_env : File.join(Rails.root, 'config', "sidekiq.yml")
+
+    if File.exist?(sidekiq_config) && !Rails.env.test?
+      require 'sidekiq/middleware/i18n'
+      redis_config = YAML.load_file(sidekiq_config)
+      redis_url = { host: redis_config[:redis_host], port: redis_config[:redis_port], db: redis_config[:redis_database], namespace: "cache_checkapi_#{Rails.env}" }
+      config.cache_store = :redis_cache_store, redis_url
```
caiosba, line 60: Since we have it here, do we still need it at https://github.com/meedan/check-api/blob/CV2-3055-use-config-environments-production-rb-from-file-not-from-ssm-in-check-api/config/initializers/03_sidekiq.rb#L28 ?

It's a bit awkward that the Redis configuration lives in the Sidekiq configuration file while we use Redis for other stuff... I know it's not in the scope of this ticket, but reviewing this made me realize that we may need to refactor this in a separate ticket (isolate Redis configuration from Sidekiq and use it for both caching and Sidekiq).

caiosba (approved):
I just have one question about one single line, but overall looks great!

PR #1507, opened 2023-05-07
CV2-3056: get user language based on supported language

melsawy:
Replace `language = self.cached_user_language(uid)` with a method `self.get_user_language(uid, {}, nil, team_id)` which verify that cached value exists in team language otherwise return the default team language as sometimes the cached value deleted from team languages

### DIFF
--- a/app/models/concerns/smooch_language.rb
+++ b/app/models/concerns/smooch_language.rb
@@ -4,10 +4,9 @@ module SmoochLanguage
   extend ActiveSupport::Concern
 
   module ClassMethods
-    def get_user_language(uid, message = {}, state = nil, team_id = nil)
-      team_id ||= self.config['team_id']
-      team = Team.find_by_id(team_id)
-      default_language = team&.default_language
+    def get_user_language(uid, message = {}, state = nil)
+      team = Team.find(self.config['team_id'])
+      default_language = team.default_language
       supported_languages = self.get_supported_languages
       guessed_language = nil
       if state == 'waiting_for_message'
--- a/app/models/concerns/smooch_language.rb
+++ b/app/models/concerns/smooch_language.rb
@@ -4,9 +4,10 @@ module SmoochLanguage
   extend ActiveSupport::Concern
 
   module ClassMethods
-    def get_user_language(uid, message = {}, state = nil)
-      team = Team.find(self.config['team_id'])
-      default_language = team.default_language
+    def get_user_language(uid, message = {}, state = nil, team_id = nil)
+      team_id ||= self.config['team_id']
+      team = Team.find_by_id(team_id)
+      default_language = team&.default_language
       supported_languages = self.get_supported_languages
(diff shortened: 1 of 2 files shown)


caiosba:
@melsawy please add a description to the PR, otherwise when someone is assigned for review, it's easier to know what's being reviewed.

app/models/concerns/smooch_language.rb
```diff
@@ -4,9 +4,10 @@ module SmoochLanguage
   extend ActiveSupport::Concern
 
   module ClassMethods
-    def get_user_language(uid, message = {}, state = nil)
-      team = Team.find(self.config['team_id'])
-      default_language = team.default_language
+    def get_user_language(uid, message = {}, state = nil, team_id = nil)
+      team_id ||= self.config['team_id']
+      team = Team.find_by_id(team_id)
```
caiosba, line 9: Why we need receive `team_id` as a parameter? I thought that in this context we should always have a `team_id` in the config.

app/models/concerns/smooch_language.rb
```diff
@@ -4,9 +4,10 @@ module SmoochLanguage
...
+      default_language = team&.default_language
```
caiosba, line 10: I think it's better to raise an exception here if `team` is `nil`, because really it should never be.

app/models/concerns/smooch_language.rb
caiosba, line 9: I think we should keep `Team.find()` here and raise an exception if there is no team with that ID.

app/models/concerns/smooch_search.rb
```diff
@@ -213,7 +213,7 @@ def search_by_keywords_for_similar_published_fact_checks(words, after, team_ids,
     def send_search_results_to_user(uid, results, team_id)
       team = Team.find(team_id)
       redis = Redis.new(REDIS_CONFIG)
-      language = self.cached_user_language(uid)
+      language = self.get_user_language(uid, {}, nil, team_id)
```
caiosba, line 216: As noted above, maybe we don't need to pass a `team_id` here.

caiosba (changes requested):

app/models/concerns/smooch_language.rb
melsawy: Removed it and depend on `team_id` in the config

app/models/concerns/smooch_language.rb
melsawy: Done

app/models/concerns/smooch_language.rb
melsawy: Done

app/models/concerns/smooch_search.rb
melsawy: Done

caiosba (approved):
Way cleaner, thanks!

PR #1508, opened 2023-05-07
CV2-3086: check file exists

melsawy:
Check file exists before open the file to avoid trigger an error `No such file or directory`

### DIFF
--- a/app/models/request.rb
+++ b/app/models/request.rb
@@ -154,7 +154,7 @@ class Request < ApplicationRecord
         media_type = "Uploaded#{type.camelize}"
         File.atomic_write(filepath) { |file| file.write(data) }
         media = Media.where(type: media_type, file: filename).last
-        if media.nil?
+        if media.nil? && File.exist?(filepath)
           media = Media.new(type: media_type)
           File.open(filepath) do |f2|
             media.file = f2


caiosba:
@melsawy please add a description to the PR, otherwise when someone is assigned for review, it's easier to know what's being reviewed.

hartsick:
Looks like a straightforward fix! A couple of questions / thoughts that came up as I was reviewing:
1. Do we know why the temp file isn't there sometimes? I assume it's just a periodic request and/or write failure that we can't do much about, but curious if you have more context. Is this something worth retrying?
2. Does it make sense to monitor this failure in any way?
3. Is there a test that could be written to catch this case in the future?

caiosba (changes requested):
Line 155 writes atomically to that file... can we dig more into why it wouldn't exist two lines below? Are we raising an exception and aborting execution if that line fails?

melsawy:
> 1. Do we know why the temp file isn't there sometimes? I assume it's just a periodic request and/or write failure that we can't do much about, but curious if you have more context. Is this something worth retrying?
I think it's a write failure as this error occurs only one time

> 2. Does it make sense to monitor this failure in any way?
I think yes but @caiosba can confirm

> 3. Is there a test that could be written to catch this case in the future?
No

at the end this failure occurs only one time so I think we should keep an eye on this one in future and do more investigation if got same error

melsawy:
> Line 155 writes atomically to that file... can we dig more into why it wouldn't exist two lines below? Are we raising an exception and aborting execution if that line fails?

It'll automatically raise an exception if that line fail and this atomically written happens to a tmp directory so may be deleted during deployment for example (as I mention it only one occurs)

caiosba:
Yes, this likely happened during a deployment...

caiosba (approved):

PR #1509, opened 2023-05-08
CV2-3002: Add support for OpenAI ada embeddings

computermacgyver:
add constant for new OpenAI ada model
specify the ada model as a valid relationship model type

The goal is to add a new vector model for text similarity to Alegre, Check-API, and Check-Web. All repos have a CV2-3002 branch.

References: CV2-3002


hartsick (approved):

caiosba (approved):
Yes, that's all we need in order to expose a new vector model.

PR #1510, opened 2023-05-08
Remove unused sinatra gem and file server bin

hartsick:
We no longer need the file server file since it was only needed for memebuster, and as a result we no longer need Sinatra.

Should close the related security update.


caiosba (approved):
Safe to go if automated tests pass!

PR #1511, opened 2023-05-08
Ruby 3.0 Upgrade

hartsick:
This predates the PR template!

This PR updates Ruby to 3.0 now that GraphQL has been upgraded. The main tasks were:
* Upgrading a few gems
* Making passing hash as keyword args explicit:
```
# previous: 
hash = {foo: 'bar'}; Thing.new(hash)

# now:
hash = {foo: 'bar'}; Thing.new(**hash)
```
* Removing suppression of ruby warnings from CI output, since we now expect warnings to be relevant
* Updating Bundler to 2+ to address some warnings


Gemfile
```diff
@@ -135,6 +135,8 @@ gem 'geocoder', '1.6.4'
 gem 'parallel'
 gem 'fx'
 gem 'unicode-display_width', '1.6.0'
+gem 'addressable'
+gem 'rss'
```
caiosba, line 139: Question: Is this the exact same library we were using which is now not part of the core anymore? I'm asking mostly because the RSS feature is critical for tiplines and I don't remember if in the tests it's mocked.

caiosba (approved):
It's amazing that after the GraphQL upgrade the changes here are pretty straightforward! I just added one question, but it's good to go as long as both Check API tests and integration tests are passing.

Gemfile
hartsick: Oh interesting. I think that's the case but will look into to confirm - this was a set of changes made three months ago, before I started upgrading GraphQL, so memory is a little rusty and I don't see a related note in the commit history. Tests are passing so I think it should be okay, but don't want to include things we don't need.

hartsick: Yeah, rss went from being a [default gem in 2.7.0](https://stdgems.org/2.7.0/) to being a [bundled gem in 3.0.+](https://stdgems.org/3.0.6/), which means we have to explicitly declare it

caiosba: Great, thanks!

PR #1512, opened 2023-05-08
Fixing a couple of relevant Rubocop issues related to suppressed exceptions

caiosba:
(1) Validate if an annotation type if available by using `Object.const_defined?` instead of trying to `constantize` and `rescue NameError`

(2) Fallback to an empty string if there is no picture for a Pender structure

### DIFF
--- a/app/models/dynamic_annotation/annotation_type.rb
+++ b/app/models/dynamic_annotation/annotation_type.rb
@@ -14,11 +14,8 @@ class DynamicAnnotation::AnnotationType < ApplicationRecord
   private
 
   def annotation_type_is_available
-    begin
-      self.annotation_type.camelize.constantize
+    if Object.const_defined?(self.annotation_type.camelize)
       errors.add(:annotation_type, 'is not available')
-    rescue NameError
-      # Not defined
     end
   end
 end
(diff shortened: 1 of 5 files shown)


app/models/dynamic_annotation/annotation_type.rb
```diff
@@ -14,11 +14,8 @@ def json_schema_enabled?
   private
 
   def annotation_type_is_available
-    begin
-      self.annotation_type.camelize.constantize
+    if !self.annotation_type.blank? && Object.const_defined?(self.annotation_type.camelize)
```
hartsick, line 17: I'm a little confused by this - I'm reading it as "if we have the annotation type, and the annotation type class is defined, add an error" - is that understanding correct? If so, is that what's intended? it reads like we want to do an unless, or some other logic

hartsick (approved):
I left one question on something I was confused by. If that behavior is correct, looks good to me!

app/models/dynamic_annotation/annotation_type.rb
caiosba: Yes, that's the intent. We have core annotation types that are defined by models in the code base. So, `Tag` and `Comment` can't be custom annotation types because there is a class for them, they are already core annotations.

PR #1515, opened 2023-05-15
Fix/cv2 3113 3085 sentry issues

melsawy:
- Use `&.` before map method to avoid an error related to nil value (CV2-3113)
 - Minimize data for Rails.logger.info and use `team_id` instead of Team object to overcome Encoding::CompatibilityError error between data in `team` object and data in `data` object and not sure if this fix will handle encoding error or not (CV2-3085)

### DIFF
--- a/app/controllers/api/v1/graphql_controller.rb
+++ b/app/controllers/api/v1/graphql_controller.rb
@@ -22,7 +22,7 @@ module Api
 
       def batch
         parse_graphql_result do |context|
-          queries = params[:_json].map do |param|
+          queries = params[:_json]&.map do |param|
             {
               query: param[:query],
               variables: prepare_query_variables(param[:variables]),
@@ -35,7 +35,7 @@ module Api
               id: result.query.context[:id],
               payload: result.to_h
             }
-          end
+          end unless queries.nil?
           results
         end
       end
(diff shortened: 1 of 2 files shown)


app/controllers/api/v1/graphql_controller.rb
```diff
@@ -35,7 +35,7 @@ def batch
               id: result.query.context[:id],
               payload: result.to_h
             }
-          end
+          end unless queries.nil?
```
caiosba, line 38: I didn't understand this `unless`... aren't we defining `queries` for the first time in this block?

melsawy: `queries` is defined but if `params[:_json]` is nil then `queries` will be nil and expected that `RelayOnRailsSchema.multiplex(queries)` with raise an error so I added this condition

caiosba: Ah OK, the Git diff confused me... so basically we're handling the case where the batch endpoint is called but no queries are passed to it?

caiosba (approved):

app/controllers/api/v1/graphql_controller.rb
melsawy: yes

PR #1518, opened 2023-05-17
CV2-2910: fix SyntaxError related to version metadata

melsawy:
User `ActiveRecord::Base.connection.quote_string` for metadata json field instead of multiple use inside metadata array


hartsick:
@melsawy good fix! would you add a unit test to capture the issue we saw and prevent against future regressions? hoping it's straightforward since there don't seem to be many dependencies inside the method.

melsawy:
> @melsawy good fix! would you add a unit test to capture the issue we saw and prevent against future regressions? hoping it's straightforward since there don't seem to be many dependencies inside the method.

Done

hartsick (approved):
Great! Thanks for adding the test!

PR #1519, opened 2023-05-18
CV2-3139: add de locales for invited mail template

melsawy:
Add German translation for invitation mail

### DIFF
--- a/config/locales/de.yml
+++ b/config/locales/de.yml
@@ -1,2 +1,24 @@
 ---
-de: {}
+de:
+  devise:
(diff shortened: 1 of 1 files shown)


caiosba (approved):
You just copied from Devise and adapted the variables for our case, right, @melsawy ?

melsawy:
> You just copied from Devise and adapted the variables for our case, right, @melsawy ?
y 
I copied from our `en.yml` and did a google translation as  already have our own copy which different than device one

PR #1523, opened 2023-05-18
Fixes for newsletter feature in production environment

caiosba:
* Load classes `UrlRewriter` and `RssFeed`
* Fix rake task extension


config/initializers/plugins.rb
```diff
@@ -1,2 +1,2 @@
-# Initialize the plugin by just calling the classes here
-CcDeville && Bot::Keep && Workflow::Workflow.workflows && CheckS3 && Bot::Fetch && Bot::Smooch && Bot::Slack && Bot::Alegre && CheckChannels
+# Load classes on boot, in production, that otherwise wouldn't be auto-loaded by default
+CcDeville && Bot::Keep && Workflow::Workflow.workflows && CheckS3 && Bot::Fetch && Bot::Smooch && Bot::Slack && Bot::Alegre && CheckChannels && RssFeed && UrlRewriter
```
hartsick, line 2: **thought, non-blocking** if the application needs these classes, I think it makes sense to move these to an autoloaded path (eg `app/lib` instead of `lib`; sorry for missing that on previous review). what do you think?

though also - many of these on the list before the diff are already in the autoload path, so I'm wondering if I'm misunderstanding

hartsick (approved):
good to merge! left one comment about how we use the plugins file, but non-blocking

config/initializers/plugins.rb
caiosba: Yes, we need to revisit that, this is not the right way of doing it. Some paths that I expected to auto-load don't actually do... for example, `app/models/bot`. I was postponing this until the Rails 6 upgrade so we could do it properly using Zeitwerk.

hartsick: ticket: https://meedan.atlassian.net/browse/CV2-3153

PR #1526, opened 2023-05-23
CV2-3142: handle delete cached fields for relationship-bulk_destroy

melsawy:
For `Relationship.bulk_destroy` there is a method that delete item cached fields `delete_cached_field` to enforce  create cached field with a new values 
I added all fields that related to `model: Relationship`


caiosba (approved):
LGTM

PR #1532, opened 2023-06-05
Update rails-i18n revision on Gemfile.lock

amoedoamorim:
Updated rails-i18n reference to include "bn" locale date format fix done in our fork. Tested manually on my local dev env.

A couple of patch updates sneaked in when I did bundle install but should be irrelevant
:)

Fixes CV2-3183


hartsick (approved):
🎉

PR #1539, opened 2023-06-08
Remove leading and trailing spaces from WhatsApp template placeholders.

caiosba:
This way, formatting won't be messed up, like bold, for example.

Fixes CV2-3230.

### DIFF
--- a/app/models/tipline_newsletter.rb
+++ b/app/models/tipline_newsletter.rb
@@ -62,7 +62,7 @@ class TiplineNewsletter < ApplicationRecord
     timezone = self.timezone
 
     # If an offset is being passed, then it's in the new format... we used to support timezone names
-    if self.timezone.match?(/\W\d\d:\d\d/)
+    if self.timezone.to_s.match?(/\W\d\d:\d\d/)
       timezone = self.timezone.match(/\W\d\d:\d\d/)
     else
       timezone = self.timezone.to_s.upcase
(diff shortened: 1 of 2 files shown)


app/models/tipline_newsletter.rb
```diff
@@ -62,7 +62,7 @@ def parsed_timezone
     timezone = self.timezone
 
     # If an offset is being passed, then it's in the new format... we used to support timezone names
-    if self.timezone.match?(/\W\d\d:\d\d/)
+    if self.timezone.to_s.match?(/\W\d\d:\d\d/)
```
caiosba, line 65: Unrelated to this PR really, but just for robustness.

app/models/tipline_newsletter.rb
```diff
@@ -273,7 +273,9 @@ def validate_header_file(max_size, allowed_types, message)
 
   def not_scheduled_for_the_past
     if self.content_type == 'static' && self.scheduled_time.past?
-      errors.add(:send_on, I18n.t(:send_on_must_be_in_the_future))
+      field = :send_on
+      field = :time if self.scheduled_time.strftime('%Y-%m-%d') == Time.now.utc.strftime('%Y-%m-%d')
+      errors.add(field, I18n.t(:send_on_must_be_in_the_future))
```
caiosba, line 278: Unrelated to this PR, but just for consistency, mark the problematic field as "time" instead of "day" if a datetime is in the past but the day is the same as today.

melsawy (approved):

PR #1543, opened 2023-06-13
New get_language_detection field

amoedoamorim:
This adds a field get_language_detection to the graphql Team type and and language_detection field to the mutation and a setter to the Team model.
And changes logic in smooch_languages.rb to only have Alegre autodetect message if this new setting is enabled.

### DIFF
--- a/app/graph/mutations/team_mutations.rb
+++ b/app/graph/mutations/team_mutations.rb
@@ -16,6 +16,7 @@ module TeamMutations
     slack_notifications: 'str',
     language: 'str',
     languages: 'json',
+    language_detection: 'bool',
     list_columns: 'json',
     tipline_inbox_filters: 'str',
     suggested_matches_filters: 'str',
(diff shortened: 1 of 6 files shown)


caiosba (approved):
Looks good so far!

amoedoamorim:
How do we make this `language_detection` field initialize as `true`?

caiosba:
> How do we make this `language_detection` field initialize as `true`?

Hum... in that case, it's better to use the opposite (`no_language_detection`)

amoedoamorim:
> > How do we make this `language_detection` field initialize as `true`?
> 
> Hum... in that case, it's better to use the opposite (`no_language_detection`)

Not sure I'm following here. Is this ruby magic or do you actually mean I should invert the implementation logic?

caiosba:
> > > How do we make this `language_detection` field initialize as `true`?
> > 
> > 
> > Hum... in that case, it's better to use the opposite (`no_language_detection`)
> 
> Not sure I'm following here. Is this ruby magic or do you actually mean I should invert the implementation logic?

Invert the logic, in a way that is not weird but that we don't need to set it as `true` for existing and new teams.

caiosba (approved):

PR #1544, opened 2023-06-15
Refactor/cv2 3193 shared feed back-end

melsawy:
Refactor Feed & FeedTeam models by doing the following:

- Feed model: add/remove fields

1. `saved_search_id`: reference to list (one-to-many: a SavedSearch has many Feed)
2. `user_id`: reference to user (user who created the feed)
3. `team_id`: reference to team (team who created the feed)
4. `description`: of type text
5. `tags`: array format of strings, i.e ['tag_a', 'tag_b']
6. `licenses`: array format of int, i.e [1, 2]
7. remove `filters` column 
 
- Feed model:  validations

1. Validate that list belongs to same feed teams
2. Validate that title and licenses are present
3. Validate licences in allowed values ({ 1 => 'academic', 2 => 'commercial', 3 => 'open_source' })

- Feed model:  permissions

1. Allow admin to create/update/destroy feed

- Feed model: graphql(types & mutations)

1. Expose all new fields to FeedType
2. Expose a mutations to create/update/destroy Feed

- FeedTeam model: add/remove fields

1. `saved_search_id`: reference to list (one-to-many: a SavedSearch has many Feed)
2. remove `filters` column 

- FeedTeam model:  validations

1. Validate that list belongs to same feed teams
2. Validate that team and feed are present

- Feed model: graphql(types & mutations)

1. Expose `saved_search` fields to FeedType
2. Expose `saved_search_id`  to updateFeedTeam mutation

- SavedSearch type: new fields

1. is_part_of_feeds: Boolean to determine either the list is a part of feed or not
2. feeds: a list of `FeedType` to get all feeds related to the list

- Add unit tests to cover all above cases.
- Fix existing tests.


app/graph/types/feed_type.rb
```diff
@@ -5,13 +5,46 @@
...
+  field :team_id, types.Int
+
+  field :user do
+    type -> { UserType }
+
+    resolve -> (feed, _args, ctx) {
+      RecordLoader.for(User).load(feed.user_id).then do |user|
+        ability = ctx[:ability] || Ability.new
+        user if ability.can?(:read, user)
+      end
```
caiosba, line 29: Suggestion: We don't need `RecordLoader` here because we're not going to load feed users in bulks (e.g., no risk of a `N + 1` queries problem).

app/graph/types/feed_type.rb
```diff
@@ -5,13 +5,46 @@
...
+    }
+  end
+
+  field :team do
+    type -> { TeamType }
+
+    resolve -> (feed, _args, _ctx) {
+      RecordLoader.for(Team).load(feed.team_id)
```
caiosba, line 37: Suggestion:  No need for `RecordLoader` here.

app/graph/types/feed_type.rb
```diff
@@ -5,13 +5,46 @@
...
+    }
+  end
+
+  field :saved_search do
+    type -> { SavedSearchType }
+
+    resolve -> (feed, _args, _ctx) {
+      RecordLoader.for(SavedSearch).load(feed.saved_search_id)
```
caiosba, line 45: Suggestion: No need for `RecordLoader` here.

app/graph/types/mutation_type.rb
```diff
@@ -128,6 +128,10 @@
   field :updateFactCheck, field: FactCheckMutations::Update.field
   field :destroyFactCheck, field: FactCheckMutations::Destroy.field
 
+  field :createFeed, field: FeedMutations::Create.field
+  field :updateFeed, field: FeedMutations::Update.field
+  field :destroyFeed, field: FeedMutations::Destroy.field
```
caiosba, line 133: Suggestion: We won't use `destroyFeed` for now, so no need to expose it.

app/models/feed.rb
```diff
@@ -125,4 +142,29 @@ def self.notify_subscribers(pm, title, summary, url)
...
+  def licenses_in_allowed_values
+    unless (self.licenses - LICENSES.keys).empty?
+      errors.add(:licenses, I18n.t(:"errors.messages.invalid_feed_licenses_value"))
+    end
+  end
+
+  def saved_search_belongs_to_feed_teams
+    unless saved_search_id.blank?
+      team_ids = self.feed_teams.map(&:team_id)
+      team_ids << self.team_id
```
caiosba, line 162: Suggestion: Implement a method `team_ids` in `Feed` that contains this logic... I saw this being used elsewhere.

app/models/ability.rb
```diff
@@ -59,6 +59,12 @@ def admin_perms
     can [:update, :destroy], TeamUser, team_id: @context_team.id
     can :duplicate, Team, :id => @context_team.id
     can :set_privacy, Project, :team_id => @context_team.id
+    can :cud, Feed do |obj|
+      team_ids = obj.feed_teams.map(&:team_id)
+      team_ids << obj.team_id
```
caiosba, line 64: Suggestion: Implement a method `team_ids` in `Feed` that contains this logic... I saw this being used elsewhere.

config/locales/en.yml
```diff
@@ -108,6 +108,8 @@ en:
       invalid_project_media_archived_value: Sorry, archived value not supported
       invalid_fact_check_language_value: Sorry, language value not supported
       fact_check_empty_title_and_summary: Sorry, you should fill title or summary
+      invalid_feed_licenses_value: Sorry, licenses value not supported
+      invalid_feed_saved_search_value: Sorry, list should belong to one of feeds team
```
caiosba, line 112: Fix: You're calling these strings with `errors.add(:licenses [...])`, so the error message will be:

```
Licenses Sorry, licenses value not supported
```

This happens because it concatenates the first and second arguments passed to `errors.add`. In that case, the messages could be:

```
value not supported
should belong to a workspace that is part of this feed
```

Let's use "workspace" instead of "team" for user-facing messages.

db/migrate/20230614045753_add_fields_to_feed.rb
```diff
@@ -0,0 +1,12 @@
+class AddFieldsToFeed < ActiveRecord::Migration[6.1]
+  def change
+    add_reference :feeds, :saved_search, index: true
+    add_reference :feeds, :user, index: true
+    add_reference :feeds, :team
```
caiosba, line 5: Suggestion: Add an index here too.

db/migrate/20230614045753_add_fields_to_feed.rb
```diff
@@ -0,0 +1,12 @@
...
+    add_column :feeds, :description, :text
+    add_column :feeds, :tags, :string
```
caiosba, line 7: Question: Why `string` and not `jsonb`? Or, why not `array`, like `licenses`?

melsawy: `string` as I implemented this field same as `send_every` https://github.com/meedan/check-api/blob/develop/db/migrate/20230516204030_create_tipline_newsletters.rb#L16

app/graph/types/feed_type.rb
melsawy: Done

app/graph/types/feed_type.rb
melsawy: Done

app/graph/types/feed_type.rb
melsawy: Done

app/graph/types/mutation_type.rb
melsawy: Done

app/models/feed.rb
melsawy: Done and add a method `get_team_ids`

app/models/ability.rb
melsawy: Done

config/locales/en.yml
melsawy: Done

db/migrate/20230614045753_add_fields_to_feed.rb
melsawy: Dne

app/graph/types/feed_type.rb
```diff
@@ -5,13 +5,46 @@
...
+  field :user do
+    type -> { UserType }
+
+    resolve -> (feed, _args, ctx) {
+      feed.user.then do |user|
+        ability = ctx[:ability] || Ability.new
+        user if ability.can?(:read, user)
+      end
+    }
+  end
```
caiosba, line 31: I think that in most cases there will be no permission to load a full user object... what if it we just return the user name in a string field `user_name` with no permission check?

app/models/ability.rb
```diff
@@ -59,6 +59,10 @@ def admin_perms
     can [:update, :destroy], TeamUser, team_id: @context_team.id
     can :duplicate, Team, :id => @context_team.id
     can :set_privacy, Project, :team_id => @context_team.id
+    can :cud, Feed do |obj|
+      obj.get_team_ids.include?(@context_team.id)
+    end
+    can [:create, :destroy], FeedTeam, team_id: @context_team.id
```
caiosba, line 65: What about `:update`?

app/models/saved_search.rb
```diff
@@ -5,4 +5,6 @@ class SavedSearch < ApplicationRecord
   validates :title, uniqueness: { scope: :team_id }, unless: proc { |ss| ss.is_being_copied }
 
   belongs_to :team, optional: true
+  has_many :feeds
+  has_many :feed_teams
```
caiosba, line 9: I think we need a `dependent: :nullify` here

config/locales/en.yml
```diff
@@ -108,6 +108,8 @@ en:
       invalid_project_media_archived_value: Sorry, archived value not supported
       invalid_fact_check_language_value: Sorry, language value not supported
       fact_check_empty_title_and_summary: Sorry, you should fill title or summary
+      invalid_feed_saved_search_value: value not supported
+        should belong to a workspace that is part of this feed
```
caiosba, line 112: Key missing for this second translation

caiosba (changes requested):

config/locales/en.yml
melsawy: I removed the key as I replaced the validation with `validates_inclusion_of`

app/graph/types/feed_type.rb
melsawy: Based on Figma design there is no need to get user data so I removed this field

app/models/saved_search.rb
melsawy: Done

caiosba (approved):

PR #1551, opened 2023-06-29
cv2-3077: Validation errors should not be sent to Sentry, only to the client

vasconsaurus:
## Description

We were sending validation errors to Sentry, e.g: 'Project can't be blank'. We don't need to send those to Sentry, only to the client.

References: [cv2-3077](https://meedan.atlassian.net/browse/CV2-3077?atlOrigin=eyJpIjoiMzQ0MzVlZjhjNWZkNDJlYzg0ZDRmYzY2Y2JkMTU0YzIiLCJwIjoiaiJ9)

## How has this been tested?

No, since we don't send errors from the development environment. I'll test this on QA.

### DIFF
--- a/app/controllers/api/v1/graphql_controller.rb
+++ b/app/controllers/api/v1/graphql_controller.rb
@@ -53,7 +53,7 @@ module Api
         # Mutations are not batched, so we can return errors in the root
         rescue ActiveRecord::RecordInvalid, RuntimeError, NameError, GraphQL::Batch::NestedError => e
           @output = parse_json_exception(e)
-          CheckSentry.notify(e) unless ActiveRecord::RecordInvalid
+          CheckSentry.notify(e) unless e.is_a?(ActiveRecord::RecordInvalid)
           render json: @output, status: 400
         rescue ActiveRecord::StaleObjectError, ActiveRecord::RecordNotUnique => e
           @output = format_error_message(e)
--- a/app/controllers/api/v1/graphql_controller.rb
+++ b/app/controllers/api/v1/graphql_controller.rb
@@ -53,7 +53,7 @@ module Api
         # Mutations are not batched, so we can return errors in the root
         rescue ActiveRecord::RecordInvalid, RuntimeError, NameError, GraphQL::Batch::NestedError => e
           @output = parse_json_exception(e)
-          CheckSentry.notify(e)
+          CheckSentry.notify(e) unless ActiveRecord::RecordInvalid
           render json: @output, status: 400
         rescue ActiveRecord::StaleObjectError, ActiveRecord::RecordNotUnique => e
           @output = format_error_message(e)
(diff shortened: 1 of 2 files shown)


app/controllers/api/v1/graphql_controller.rb
```diff
@@ -53,7 +53,7 @@ def parse_graphql_result
         # Mutations are not batched, so we can return errors in the root
         rescue ActiveRecord::RecordInvalid, RuntimeError, NameError, GraphQL::Batch::NestedError => e
           @output = parse_json_exception(e)
-          CheckSentry.notify(e)
+          CheckSentry.notify(e) unless ActiveRecord::RecordInvalid
```
caiosba, line 56: @vasconsaurus you need to check that `e` is an exception of that type. When you just check for the constant itself, it will always return `true` since it's defined:

```
> !!ActiveRecord::RecordInvalid
=> true
```

There are different ways to check for the class of an object, for example using `is_a?` or `kind_of?`:

```
> e = StandardError.new
=> #<StandardError: StandardError>

> e.is_a?(ActiveRecord::RecordInvalid)
=> false

> e.kind_of?(ActiveRecord::RecordInvalid)
=> false

> e = ActiveRecord::RecordInvalid.new
=> #<ActiveRecord::RecordInvalid: Record invalid>

> e.is_a?(ActiveRecord::RecordInvalid)
=> true

> e.kind_of?(ActiveRecord::RecordInvalid)
=> true
```

caiosba (changes requested):

app/controllers/api/v1/graphql_controller.rb
vasconsaurus: derp! of course! will update it

app/controllers/api/v1/graphql_controller.rb
```diff
@@ -53,7 +53,7 @@ def parse_graphql_result
         # Mutations are not batched, so we can return errors in the root
         rescue ActiveRecord::RecordInvalid, RuntimeError, NameError, GraphQL::Batch::NestedError => e
           @output = parse_json_exception(e)
-          CheckSentry.notify(e)
+          CheckSentry.notify(e) unless e.is_a?(ActiveRecord::RecordInvalid)
```
hartsick, line 56: I think rather than doing a clause here, we should use the Sentry ruby gem's configuration for ignoring certain exceptions. Here are the docs on how to use: https://docs.sentry.io/platforms/ruby/configuration/options/, cmd+f `excluded_exceptions`.

This would go in the Sentry config (check-api/config/initializers/02_sentry.rb)

hartsick (changes requested):
What you have works well, but I have a different approach I'd suggest that is a little more conventional for Sentry's ruby library and would let us have one place to reference which exceptions we're ignoring in the application!

caiosba (approved):

PR #1561, opened 2023-07-10
CV2-2872 - count Chinese and Japanese words better

computermacgyver:
## Description

Per CV2-2872 we have an issue where Chinese and Japanese sentences are counted as only one word simply because they are not space-separated. 

This ticket counts two Chinese characters as one word and 4-Japanese kana as one word. This is simple and let's us continue using regular expressions in Ruby without having to do anything more in-depth.

References: CV2-2872

## How has this been tested?

Additional tests added

N/A

## Checklist

- [x] I have made needed changes to the README
- [x] If I added a third party module, I included a rationale for doing so and followed our current [guidelines](https://meedan.atlassian.net/wiki/spaces/ENG/overview#Choose-the-%E2%80%9Cright%E2%80%9D-3rd-party-module)

[CV2-2872]: https://meedan.atlassian.net/browse/CV2-2872?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/bot/alegre.rb
+++ b/app/models/bot/alegre.rb
@@ -196,9 +196,10 @@ class Bot::Alegre < BotUser
     # Get the number of space-separated words (Does not work with Chinese/Japanese)
     space_separted_words = text.gsub(/[^\p{L}\s]/u, '').strip.chomp.split(/\s+/).size
 
-    # This splits the text on any non unicode word boundary (works with Chinese, Japanese)
+    # This removes URLs
+    # Then it splits the text on any non unicode word boundary (works with Chinese, Japanese)
     # We then clean each word and remove any empty ones
-    unicode_words = text.scan(/(?u)\w+/).map{|w| w.gsub(/[^\p{L}\s]/u, '').strip.chomp}.reject{|w| w.length==0}
+    unicode_words = text.gsub(/https?:\/\/\S+/u, '').scan(/(?u)\w+/).map{|w| w.gsub(/[^\p{L}\s]/u, '').strip.chomp}.reject{|w| w.length==0}
     # For each word, we:
     # Get the number of Chinese characters. We'll assume one character is like one word
     # Get the number of Japanese hiragana/katakana (kana) characters.
--- a/app/models/bot/alegre.rb
+++ b/app/models/bot/alegre.rb
@@ -214,7 +214,10 @@ class Bot::Alegre < BotUser
 
   def self.get_items_from_similar_text(team_id, text, fields = nil, threshold = nil, models = nil, fuzzy = false)
     team_ids = [team_id].flatten
-    return {} if text.blank? || self.get_number_of_words(text) < 3
+    if text.blank? || self.get_number_of_words(text) < 3
+      Rails.logger.info("[Alegre Bot] get_items_from_similar_text returning early due to blank/short text #{text}")
+      return {}
+    end
     fields ||= ALL_TEXT_SIMILARITY_FIELDS
     threshold ||= self.get_threshold_for_query('text', nil, true)
     models ||= [self.matching_model_to_use(team_ids)].flatten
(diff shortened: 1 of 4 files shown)


skyemeedan (approved):
Seems like a good first approximation! This rabbit-holed me for a few minutes of quick googling that suggested ~1.6 Han characters per word, depending on dialect of Chinese ... but that is probably beside the point for this classification since the purpose of this function is really just "is this text long enough be vectorized or just fall back to ES".

computermacgyver:
> Seems like a good first approximation! This rabbit-holed me for a few minutes of quick googling that suggested ~1.6 Han characters per word, depending on dialect of Chinese ... but that is probably beside the point for this classification since the purpose of this function is really just "is this text long enough be vectorized or just fall back to ES".

Yeah, I think anything in the range [1,2] would work for our purposes. I've updated the code to go with 2 Chinese characters equating roughly to one word in a space-separated language. Since we apply the ceiling function, this means 3 Chinese characters count as "2 words", which seems fine. At the end of the day, we're only really needing to know if it's  one/two "words" or more than that.

Here's a [not so random paper on the subject](https://arxiv.org/abs/1506.00572) :joy: :mirror: , but that paper only looked at character equivalence, not words.

caiosba (approved):
Great work Scott, thanks for the comments in the code and for the tests!

app/models/concerns/smooch_search.rb
```diff
@@ -146,7 +146,7 @@ def search_for_similar_published_fact_checks_no_cache(type, query, team_ids, aft
           text = [link.pender_data['description'].to_s, text.to_s.gsub(/https?:\/\/[^\s]+/, '').strip].max_by(&:length)
         end
         return [] if text.blank?
-        words = text.split(/\s+/)
+        words = Bot::Alegre.get_number_of_words(text)
```
computermacgyver, line 149: FYI @caiosba I think we can and should use the new function here in Smooch Bot as well. Let me know your thoughts, please.

app/models/concerns/team_rules.rb
```diff
@@ -21,6 +21,7 @@ def has_less_than_x_words(pm, value, _rule_id)
       return false unless pm.media&.type == 'Claim'
       smooch_message = get_smooch_message(pm)
       return false if smooch_message['text'].blank?
+      #TODO: Consider using Bot::Alegre.number_of_words
       smooch_message['text'].to_s.gsub(Bot::Smooch::MESSAGE_BOUNDARY, '').split(/\s+/).select{ |w| (w =~ /^[0-9]+$/).nil? }.size <= value.to_i
```
computermacgyver, line 25: I'm not familiar with this code and unsure whether we're better leaving it as is (i.e., white space separated tokens) or using Bot::Alegre.number_of_words. Any preference @caiosba ? If you're not sure, I'll leave as is and file a ticket in the backlog for further investigation in the future.

app/models/concerns/smooch_search.rb
caiosba: I agree @computermacgyver

app/models/concerns/team_rules.rb
caiosba: Good catch. I think we can leave that code as is. It gets all messages from a conversation and ignores menu options first. But it used in rules which is not actively used (if used at all).

app/models/bot/alegre.rb
```diff
@@ -711,7 +730,7 @@ def self.is_text_too_short?(pm, length_threshold)
     unless pm.alegre_matched_fields.blank?
       fields_size = []
       pm.alegre_matched_fields.uniq.each do |field|
-        fields_size << pm.send(field).to_s.split(/\s/).length if pm.respond_to?(field)
+        fields_size << self.get_number_of_words(pm.send(field)) if pm.respond_to?(field)
```
computermacgyver, line 733: This is the root of CV2-3476 . I started using `get_number_words` here, but dropped the `.to_s` unintentionally without knowing that this handled the case of `nil`. (i.e., `nil.to_s` generates a blank string). I'll update get_number_of_words to include `.to_s` and return 0 for `nil` input.

PR #1562, opened 2023-07-10
Remove ticket link from pull_request_template.md

vasconsaurus:
PRs should only include the ticket number and not include the links to our jira board, since it's not publicly accessible.


caiosba (approved):
Good catch, thanks Manu!

PR #1577, opened 2023-07-17
Cv2 3449 stop running known failed transcriptions

DGaffney:
## Description

In some cases, audio transcription fails for known, unsolvable reasons. Namely, we know these to currently be (a) when an audio file is too short to contain audio, and (b) when a video file has no audio information attached. Currently, in check-api we keep trying again for any non-COMPLETED value response from the transcription controller. Instead, we should mark these cases as DONE to denote that they were not successfully COMPLETED, but also should not be re-enqueued.

References: CV2-3449, CV2-3448 (alegre work was completed in this ticket)

## How has this been tested?

Not directly tested yet - added automated test for what i expect to occur given this change

The change in `alegre.rb` - I believe this is all that's needed to bypass triggering the job - @amoedoamorim would love to know what you think

## Checklist

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -23,7 +23,7 @@ checks:
       threshold: 4
   return-statements:
     config:
-      threshold: 4
+      threshold: 5
 plugins:
   fixme:
     enabled: false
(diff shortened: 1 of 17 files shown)


caiosba (changes requested):
The change itself looks straightforward, but we need to fix the test :)

PR #1578, opened 2023-07-18
Ignore some events from WhatsApp Cloud API which are not relevant.

caiosba:
## Description

First, allow core bots to ignore requests. By default, it's false, so no requests are ignored.

For WhatsApp Cloud API, ignore events of type "read" and "sent", since we just react to "delivered".

Reference: CV2-3453.

## How has this been tested?

I added a functional test for this case.

## Checklist

### DIFF
--- a/app/models/concerns/smooch_capi.rb
+++ b/app/models/concerns/smooch_capi.rb
@@ -4,6 +4,11 @@ module SmoochCapi
   extend ActiveSupport::Concern
 
   module ClassMethods
+    def should_ignore_capi_request?(request)
+      event = request.params.dig('entry', 0, 'changes', 0, 'value', 'statuses', 0, 'status').to_s
+      ['read', 'sent'].include?(event)
+    end
+
     def valid_capi_request?(request)
       valid = false
       if request.params['hub.mode'] == 'subscribe'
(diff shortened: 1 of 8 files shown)


app/models/concerns/smooch_capi.rb
```diff
@@ -4,6 +4,11 @@ module SmoochCapi
   extend ActiveSupport::Concern
 
   module ClassMethods
+    def should_ignore_capi_request?(request)
+      event = request.params.dig('entry', 0, 'changes', 0, 'value', 'statuses', 0, 'status').to_s
```
hartsick, line 8: just curious - will we ever hit this for a non-CAPI request? if so, would it make sense to check for the request origin specifically (like, if we have something that indicates origin in the params)?

I'm assuming the params will just be different based on the request origin and that's how we're protecting against non-CAPI request types, but wonder if there's a way to make clearer to us.

if not, I think the concern and method name indicates its use. I just have always found this pattern (indicating request origin by the method name we're calling, without trying to understand the request origin before calling the method) a little confusing

hartsick (approved):
daaaaaamn look at that test! looks great, I had a design question/thought that isn't necessary to address for this PR

app/models/concerns/smooch_capi.rb
caiosba: Good question! Right, we will never hit that for non-CAPI requests because that structure is very specific. In other methods (for cases we handle), we also check for headers that indicate that the request is coming from CAPI.

PR #1586, opened 2023-07-19
When creating a `TiplineMessage`, use the `language` saved in the Smooch payload if available

caiosba:
## Description

When creating a `TiplineMessage`, use the `language` saved in the Smooch payload if available. Otherwise, just fallback to the current tipline user language, which is the behavior today.

This is particularly important for newsletters, where what matters is the newsletter language.

Related to CV2-3298.

## How has this been tested?

Existing tests should still cover this.

## Checklist

### DIFF
--- a/app/models/bot/smooch.rb
+++ b/app/models/bot/smooch.rb
@@ -928,7 +928,7 @@ class Bot::Smooch < BotUser
         last_smooch_response = self.send_message_to_user(uid, '', { 'type' => 'image', 'mediaUrl' => report.report_design_image_url })
         Rails.logger.info "[Smooch Bot] Sent report visual card to user #{uid} for item with ID #{pm.id}, response was: #{last_smooch_response.to_json}"
       end
-      self.save_smooch_response(last_smooch_response, parent, data['received'], fallback_template, lang)
+      self.save_smooch_response(last_smooch_response, parent, data['received'], fallback_template, lang || 'en')
     end
   end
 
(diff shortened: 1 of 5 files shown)


app/models/bot/smooch.rb
```diff
@@ -941,7 +941,7 @@ def self.get_id_from_send_response(response)
     (RequestStore.store[:smooch_bot_provider] == 'TURN' || RequestStore.store[:smooch_bot_provider] == 'CAPI') ? response_body&.dig('messages', 0, 'id') : response&.body&.message&.id
   end
 
-  def self.save_smooch_response(response, pm, query_date, fallback_template = nil, lang = 'en', custom = {}, expire = nil)
+  def self.save_smooch_response(response, pm, query_date, fallback_template = nil, lang = nil, custom = {}, expire = nil)
```
hartsick, line 944: question: could you help me understand why you moved the "en" default out of this method signature? it seems like now we have more places to track down where it's being set, and I'm not seeing the benefit

hartsick (commented):
the change to TiplineMessage look good to me, but I have a question about the changes to where we set the default language. if I'm reading the change correctly (I may be missing something!) then I prefer how we had it before because it seems more understandable and maintainable

app/models/bot/smooch.rb
caiosba: We just have two types of cases today:

(1) Newsletters delivered: For that, we want to pass in the languages of the newsletters.

(2) Messages that were sent but failed because of the 24-hours window problem and for which we want to send a template message... WhatsApp requires a language to be present, so for those cases it's better to be sure it's `en` otherwise a template message is not sent.

But since we're relying on that Smooch payload to also save `TiplineMessage`'s, I prefer to fallback to the user language than to English, which should be a default only for the template messages.  Currently, it won't make a difference, but I'm thinking about future cases.

Since you worked on this more than me recently, I am happy to revert this last commit if you think it's more clear.

caiosba: I revert this last commit for now.

hartsick: gotcha! yeah, I think it's good to think about future cases, but would be clearer to make that change when/if we need it. thanks for taking this comment into consideration!

PR #1592, opened 2023-07-25
CV2-3338 initial drafting on multiple vector models

DGaffney:
## Description

This work adds support for multiple vector models being used at the same time in the backend. Specifically, this is a feature for allowing extra parameters to be sent to the POST and GET /text/similarity endpoints on Alegre from Check-API. To do this, we do several things:

1. Extend `get_alegre_model_in_use` with `get_alegre_models_in_use` which returns a distinct-keyed list of models for POSTs,
2. Extend `get_text_similarity_model ` with `get_text_similarity_model s` which returns a distinct-keyed list of models for GETs,
3. Adds a `self.settings.merge(settings)` to the end of `TeamBotInstallation#alegre_settings` so that we can support and retrieve dash-cased model names from the settings (because we do `tbi.send(setting_name)` to retrieve them - we can't do that with vector names),
4. Removes erroneous model overwrite in `ReindexAlegreWorkspace#get_request_doc` since we now add `models` key earlier,
5. Adds a ton of tests for making sure this works in retrieving and setting the right values out of `TeamBotInstallation`s in many different configurations.

References: CV2-3338

## How has this been tested?

Extensive unit tests were added, but this hasn't been run against integration tests yet. The unit tests cover, I believe, any possible state for a team. I believe this should work because we already support all the fields we're sending - we're just sending several of them in the list of values instead of one value that we send today.

`alegre.rb` changes are obviously the most important as those are the substantive ones!

## Checklist


DGaffney:
Ok I don't know why this is failing on this one test, but it's failing. It seems completely unrelated to this code but I could be wrong. Either way, I think this is ready for feedback!

DGaffney:
@caiosba is there anything else we want to do here before we merge?

PR #1610, opened 2023-08-08
Prototype: NLU feature for tipline bot

caiosba:
## Description

This pull request adds an NLU feature to the tipline bot. It's meant to be a prototype version, a first approach just to evaluate the integration with Alegre's text similarity feature, user experience, accuracy and performance.

Natural-language understanding (NLU) or natural-language interpretation is a subtopic of natural-language processing in AI that deals with machine reading comprehension.

For the tiplines, the idea is to try to map user messages to menu options. Summary of changes:

- All the code related to this feature is in a new module (concern) `SmoochNlu`
- Existing tiplines should still behave the same
- New step added when evaluating a message sent by a user: if it doesn't match a menu option right away (so, it's not an exact match to a keyword or an option that was clicked), then try to match this message to a menu option using NLU
- If NLU is enabled (it's a new root-level setting for the tipline bot installation), then send that message to Alegre text similarity endpoint, with the proper context, and check if any result is returned
- Evaluate that message as a menu option if the result with highest score returned by Alegre matches a menu option for that workspace and language
- Added helper methods to enable/disable NLU and add/remove matching keywords for the menu options using Rails Console, since we won't have an interface for that right now

References: CV2-3261.

## How has this been tested?

Automated tests were added for all cases and for all lines of code.

- The current tiplines should still work the same
- There are a bunch of `FIXME` for future improvements, please let me know other improvements so I can add more `FIXME`'s for them or implement them right away if they are important even for this first prototype version.

## Checklist


app/models/concerns/smooch_nlu.rb
```diff
@@ -0,0 +1,108 @@
...
+
+    def remove_nlu_keyword(team_slug, language, menu, menu_option_index, keyword)
+      self.update_nlu_keywords(team_slug, language, menu, menu_option_index, keyword, 'remove')
+    end
+
+    def nlu_menu_option_from_message(message, options)
+      option = nil
+      if self.nlu_enabled?
+        # FIXME: No need to call Alegre if it's an exact match to one of the keywords
+        # FIXME: No need to call Alegre if message is too short
```
computermacgyver, line 97: Generally agree with all the FIXME's here. In this case, however, I don't think there is anything that is "too short" to call Alegre. Even for the one word input "newsletter" we'd call Alegre if that was not a direct keyword and look for matches.

I guess the only time we'd not call is if the message has no word characters at all (i.e., if the message is only whitespace and punctuation we could skip the call to Alegre)

app/models/concerns/smooch_nlu.rb
```diff
@@ -0,0 +1,108 @@
...
+        # FIXME: Handle error responses from Alegre
+        response = Bot::Alegre.request_api('get', '/text/similarity/', { text: message, models: self.nlu_models_to_use, context: { context: 'smooch-nlu-menu', team: Team.find(self.config['team_id']).slug } })
```
computermacgyver, line 99: We should add `threshold: 0.6` to this call otherwise Alegre will default to 0.9, which is much higher than we'll want. Obviously that value should be configurable in the future just like `nlu_models_to_use`, but we can start with 0.6 :pray:

app/models/concerns/smooch_nlu.rb
```diff
@@ -0,0 +1,108 @@
...
+          team: team_slug,
+          menu: menu,
+          menu_option_id: menu_option_id
+        }
+      }
+    end
+
+    # FIXME: Make it more flexible
+    def nlu_models_to_use
+      [Bot::Alegre::MEAN_TOKENS_MODEL]
```
computermacgyver, line 53: Very soon I'd like to use paraphrase-ml locally, but until we have that available I wonder if 
`[Bot::Alegre::OPENAI_ADA_MODEL]` is a better choice. Another option would be using both: `[Bot::Alegre::MEAN_TOKENS_MODEL, Bot::Alegre::OPENAI_ADA_MODEL]`. I'd be curious for your subjective experience trying out these options, @caiosba . I can also try locally myself.

app/models/concerns/smooch_nlu.rb
```diff
@@ -0,0 +1,108 @@
...
+    def nlu_menu_option_from_message(message, options)
+      option = nil
+      if self.nlu_enabled?
+        # FIXME: No need to call Alegre if it's an exact match to one of the keywords
+        # FIXME: No need to call Alegre if message is too short
+        # FIXME: Handle error responses from Alegre
+        response = Bot::Alegre.request_api('get', '/text/similarity/', { text: message, models: self.nlu_models_to_use, context: { context: 'smooch-nlu-menu', team: Team.find(self.config['team_id']).slug } })
+        best_result = response['result'].to_a.sort_by{ |result| result['_score'] }.last
+        unless best_result.nil?
+          option = options.find{ |option| !option['smooch_menu_option_id'].blank? && option['smooch_menu_option_id'] == best_result.dig('_source', 'context', 'menu_option_id') }
```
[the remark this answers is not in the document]
caiosba: Fixed.

app/models/concerns/smooch_nlu.rb
caiosba: Ah, good to know. Thanks, I deleted it.

app/models/concerns/smooch_nlu.rb
caiosba: Good catch! Done, now this is abstracted as a method... for now, hard-coded as 0.6.

app/models/concerns/smooch_nlu.rb
hartsick, line 1: **question** Would this make sense to be a standalone class rather than a concern? I think that it's a good candidate since it has its own internal state, and then could be more easily unit tested and the interface more clearly defined. I know we access some methods through the Smooch bot, which we could delegate if it makes sense.

test/models/bot/smooch_6_test.rb
```diff
@@ -745,4 +750,49 @@ def send_message_outside_24_hours_window(template, pm = nil)
     assert !Bot::Smooch.run(payload.to_json)
     assert_equal '000000:222222', ts.reload.uid
   end
+
+  test 'should not enable NLU for workspace without tipline' do
+    team = create_team
+    assert !Bot::Smooch.enable_nlu(team.slug)
+  end
+
+  test 'should process menu option using NLU' do
```
hartsick, line 759: I think this is a good through test! Do we have anything that tests the method that the bot class uses (`nlu_menu_option_from_message`)? If not, could be good to.

hartsick (commented):
I want to spend some time reviewing the implementation of the couple of more complex methods, but wanted to add these comments first!

app/models/concerns/smooch_nlu.rb
caiosba: Yes, good idea!

test/models/bot/smooch_6_test.rb
caiosba: Yes, under the hood that's what's happening.

hartsick: I think it would to test that method since it's the interface that we use in the implementation - even if it's just a smoke test (though great to have tests around the others if we intend to use them in rails console)

caiosba: Yes, I agree! I just re-implemented `SmoochNlu` as a standalone class so I'm going to add a unit test for that method.

caiosba:
@hartsick , thanks for your review! I just re-implemented `SmoochNlu` as a standalone class (https://github.com/meedan/check-api/blob/feature/CV2-3261-tipline-bot-nlu/app/lib/smooch_nlu.rb), and also with more granular unit tests for it (https://github.com/meedan/check-api/blob/feature/CV2-3261-tipline-bot-nlu/test/lib/smooch_nlu_test.rb). Please kindly re-review the PR :)

app/lib/smooch_nlu.rb
```diff
@@ -0,0 +1,115 @@
...
+      # FIXME: No need to call Alegre if it's an exact match to one of the keywords
+      # FIXME: No need to call Alegre if message has no word characters
+      # FIXME: Handle error responses from Alegre
+      params = {
+        text: message,
+        models: ALEGRE_MODELS_AND_THRESHOLDS.keys,
+        per_model_threshold: ALEGRE_MODELS_AND_THRESHOLDS,
+        context: {
+          context: ALEGRE_CONTEXT_KEY,
+          team: Team.find(Bot::Smooch.config['team_id']).slug
```
computermacgyver, line 53: Can we just use `@team_slug` here since it is a class variable?
(Reading further I think `menu_option_from_message` is treated as a static / class function; so, it won't have this class variable)

computermacgyver (approved):

app/lib/smooch_nlu.rb
caiosba: Yeah, not available at that scope... it's an instance variable while this method is a class method.

hartsick (approved):
caio, this is great!!

PR #1626, opened 2023-08-21
CV2-3439: fix Sentry error

melsawy:
## Description

Fix Sentry error:: NoMethodError: undefined method `[]' for nil:NilClass

References: CV2-3439

## Checklist

### DIFF
--- a/app/models/bot_user.rb
+++ b/app/models/bot_user.rb
@@ -154,7 +154,7 @@ class BotUser < User
       result = RelayOnRailsSchema.execute(query, variables: {}, context: {})
       User.current = current_user
       Team.current = current_team
-      JSON.parse(result.to_json)['data']['node']
+      JSON.parse(result.to_json)&.dig('data', 'node')
     rescue StandardError => e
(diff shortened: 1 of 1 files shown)


app/models/bot_user.rb
```diff
@@ -154,7 +154,7 @@ def graphql_result(fragment, object, team)
       result = RelayOnRailsSchema.execute(query, variables: {}, context: {})
       User.current = current_user
       Team.current = current_team
-      JSON.parse(result.to_json)['data']['node']
+      JSON.parse(result.to_json)&.dig('data', 'node')
```
caiosba, line 157: @melsawy , did you confirm this is happening for a specific case and that for other cases it still works? I just want to be sure we're not swallowing some important error related, for example, to the GraphQL upgrade. I see we have a `rescue` and a `Rails.logger` call... maybe for now I suggest we also log `result` (add it as a parameter we send to `CheckSentry.notify`) and then based on the cases where this fails we can see what to fix.

caiosba (changes requested):

PR #1630, opened 2023-08-22
CV2-3650: Improve NLU option selection

computermacgyver:
## Description

This branch makes a few improvements / tweaks to the NLU prototype.

1. It does not do an Alegre look up if options is nil. This occurs on the first message of a conversation, for example. In the future, we might want to consider all menu options across all workflows/languages, but the current fix simply doesn't return any option
2. It looks at all the Alegre results and chooses the menu option that has the highest scoring match and exists
This is a bug fix (otherwise we had issues where the top result was a menu option that didn't exist)  In the future we should consider a disambiguation menu for close options but for now I think this approach could work ok.
3. Log user intents in all cases, even when NLU is not enabled or options is nil
4. Add a method list_keywords to be used on the terminal to list out current keywords for a team/language/menu.
5. Add language to the context hash sent to Alegre  - **NOTE** this is a breaking change and requires existing keywords to be reindexed in Alegre
6. Disable NLU while in the query state (so, already waiting for more user search input)

References: CV2-3650

## How has this been tested?

Local testing and end-to-end with a local tipline installed on Telegram

We now log all unmatched intents even when NLU is disabled (the default). This will increase our logs, but I've checked with @caiosba and confirmed we don't expect an undue performance hit. The advantage of logging all intents is to gain better data into what intents we should support going forward.
 
## Checklist

### DIFF
--- a/app/lib/smooch_nlu.rb
+++ b/app/lib/smooch_nlu.rb
@@ -40,7 +40,8 @@ class SmoochNlu
   def self.menu_option_from_message(message, options)
     # FIXME: Raise exception if not in a tipline context (so, if Bot::Smooch.config is nil)
     option = nil
-    if Bot::Smooch.config.to_h['nlu_menus_enabled']
+    if Bot::Smooch.config.to_h['nlu_menus_enabled'] && !options.nil?
+      # FIXME: In the future we could consider menus across all languages when options is nil
       # FIXME: No need to call Alegre if it's an exact match to one of the keywords
       # FIXME: No need to call Alegre if message has no word characters
       # FIXME: Handle error responses from Alegre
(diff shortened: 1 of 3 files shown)


app/lib/smooch_nlu.rb
```diff
@@ -55,21 +58,33 @@ def self.menu_option_from_message(message, options)
...
-        user_query: message,
-        alegre_query: params,
-        alegre_response: response,
-        selected_option: option
-      }
-      Rails.logger.info("[Smooch NLU] [Menu Option From Message] #{log.to_json}")
+
+      # FIXME: Deal with ties (i.e., where two options have an equal count)
     end
+    # In all cases log for analysis
```
computermacgyver, line 77: I'd like @caiosba input here. Ideally, I'd like to do this logging for any message where we expected a menu option and didn't get one. That will allow us to analyze the intents we currently fail to respond to. I don't know, however, how 'heavy' an operation this is or if the `menu_option_from_message(message, options)` is a heavy traffic function such that this would have an impact on performance?

caiosba: It's OK to log for all iterations here.

computermacgyver: Great. Thanks!

caiosba (approved):
Thanks @computermacgyver , it looks good! It's missing a test for the `list_keywords` method, but since this method is not called by the application, I think we can skip it for now. I took note of that and I'll add a test to it in another PR where I'm going to add a couple of other missing tests.

test/lib/smooch_nlu_test.rb
```diff
@@ -59,6 +60,30 @@ def create_team_with_smooch_bot_installed
...
+    }
+    # Since the demo team has only one language and menu all of the following are nearly the same
+    assert nlu.list_keywords('en', 'main') == expected_output
+    assert nlu.list_keywords('en', ['main']) == expected_output
+
+    # These calls should include an empty secondary menu
+    expected_output['en']['secondary'] = []
+    assert nlu.list_keywords() == expected_output
+    assert nlu.list_keywords('en') == expected_output
+    assert nlu.list_keywords(['en']) == expected_output
```
caiosba, line 84: Suggestion: Replace `assert <actual> == <expected>` by `assert_equal <expected>, <actual>`.

caiosba (approved):
Looks good! I just added one suggestion, but not a blocker.

PR #1640, opened 2023-08-27
CV2-2256 refactor re-index nested field rake task

melsawy:
## Description

 - Refactor re-index nested field rake task to collect data using one query instead of iterate using single object.
so we collect either `comments`, `tags` or `requests` using one query to reduce the execution time
- Send object id to background job to update cached field instead of sending the whole object

References: CV2-2256

## How has this been tested?

re-run `bundle exec rails check:project_media:sync_es_nested_field['field:requests']` locally against live db and done in minutes instead of days 

- Etc.

## Checklist

### DIFF
--- a/app/lib/check_cached_fields.rb
+++ b/app/lib/check_cached_fields.rb
@@ -132,14 +132,11 @@ module CheckCachedFields
           pg_field_name: options[:pg_field_name],
           recalculate: options[:recalculate],
         }
-        obj_info = {type: obj.class.name, id: obj.id}.to_json
-        self.delay_for(1.second).update_cached_field_bg(name, obj_info, ids, callback, index_options)
+        self.delay_for(1.second).update_cached_field_bg(name, obj, ids, callback, index_options)
       end
     end
 
-    def update_cached_field_bg(name, obj_info_json, ids, callback, options)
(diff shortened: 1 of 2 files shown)


app/lib/check_cached_fields.rb
```diff
@@ -132,11 +132,14 @@ def update_cached_field(name, obj, condition, ids, callback, options)
           pg_field_name: options[:pg_field_name],
           recalculate: options[:recalculate],
         }
-        self.delay_for(1.second).update_cached_field_bg(name, obj, ids, callback, index_options)
+        obj_info = {type: obj.class.name, id: obj.id}.to_json
```
caiosba, line 135: Question: Why send `obj_info` with `type` and `id`? Are there cases where `obj_info` will contain other keys? If not, I suggest to pass instead two separate parameters `obj_type` (or `obj_class`) and `obj_id`.

caiosba (approved):
Left a question, but not a blocker.

app/lib/check_cached_fields.rb
melsawy: I reverted this commit as there is a case related to `after_destroy` which means the object not exists 
I'll file a ticket related to this change request to handle all cases

PR #1647, opened 2023-09-06
Cv2 3435 alegre with presto for audio

DGaffney:
## Description

This ticket integrates alegre's presto usage into Check-API for dealing with callbacks from storage requests on alegre. As stated in the ticket, I think we can actually get away with a pretty narrow implementation: 

At the point that the webhook for storing data is executed, we don’t actually need to make a callback-based search query - we can use the search functionality nearly exactly as written today. When a new item is added in Check-API, we send a request to Alegre to store it, which then executes a callback against presto. Alegre receives the response from Presto, and at that point, we send a callback in turn from Alegre back to Check-API. At that point, when we receive the callback on Check-API from Alegre, we know that the data is now properly stored within Alegre - so we can just synchronously send a request to Alegre to do the lookup as it exists today. In fact, we should probably…. entirely rollback the portion of changes on Alegre for the GET requests (e.g. search-side) for now. The overwhelmingly common flow is for Alegre GETs to be used when we already know that the item is now safely stored, so in those rare cases where we aren’t sure, we should route through the new webhook setup instead of making everything be doubly-called-back. Make sense?

References: 3435

## How has this been tested?

I've updated and fixed the tests on alegre to assert an expected message packet to be sent to alegre, and I'm mapping that directly to the added webhook, and then patching directly into existing functionality. Should be pretty minor if any difference downstream of that.

We will need to extensively test full service-to-service integrations just because this is involving so many different repos....

## Checklist


computermacgyver:
There are functions such as `Bot::Alegre.get_items_with_similar_media` that expect Alegre to return similar results based on the a media file specified as a URL. My reading of https://github.com/meedan/alegre/pull/340 is that this no longer works unless the audio has been stored already in Alegre with that URL. This assumption does not hold when this function is called from Smooch Search (https://github.com/meedan/check-api/blob/c983b834a859c4052553b8cf102e13f79ba78577/app/models/concerns/smooch_search.rb#L167C5-L167C5) and likely also when it is called from Check Web search. 

Currently Alegre will temporarily hash and store an Audio file to perform the search, but this will no longer be the case after https://github.com/meedan/alegre/pull/340 is merged. We therefore need to identify such instances and either (a) create exceptions for audio to fail gracefully or (b) solve the more general case of how we'll use callbacks for tipline/check-web search

caiosba (commented):
I'm going to review this once I'm able to run it locally, in order to make some tests.

app/models/bot/alegre.rb
```diff
@@ -490,7 +496,12 @@ def self.request_api(method, path, params = {}, query_or_body = 'body', retries
       response_body = response.body
       Rails.logger.info("[Alegre Bot] Alegre response: #{response_body.inspect}")
       ActiveRecord::Base.connection.reconnect! if RequestStore.store[:pause_database_connection]
-      JSON.parse(response_body)
+      parsed_response = JSON.parse(response_body)
+      if parsed_response.dig("queue") == 'audio__Model' && parsed_response.dig("body", "callback_url") != nil
+        redis = Redis.new(REDIS_CONFIG)
+        redis_response = redis.blpop(parsed_response.dig("body", "id"), timeout=120)
+        return JSON.parse(redis_response[1])
+      end
```
computermacgyver, line 504: Do we need a line `parsed_response` after `end` to ensure `parsed_response` is the return value if non-callback functions? (I always get confused with this part of Ruby)

computermacgyver (commented):
Looking great

app/models/bot/alegre.rb
```diff
@@ -126,6 +127,11 @@ def get_language_from_transcription
     end
   end
 
+  def self.valid_request?(request)
+    request.query_parameters['token'] == CheckConfig.get('alegre_token')
```
caiosba, line 131: Let's also be sure that `!request.query_parameters['token'].blank?`

app/models/concerns/alegre_webhooks.rb
```diff
@@ -0,0 +1,23 @@
...
+      #at this point, we know that the data is stored and ready for querying, so we can safely query based on package of:
+      # return requests.post(f"{check_api_host}/api/webhooks/alegre", json={
+      #     "action": action,
+      #     "model_type": model_type,
+      #     "data": data,
+      # })
+      begin
+        doc_id = request.params["data"]["requested"]["body"]["id"]
+        redis = Redis.new(REDIS_CONFIG)
+        redis.lpush(doc_id, request.params.to_json)
```
caiosba, line 16: Let's add a prefix to the Redis keys, just for better scoping/debugging if needed, like a namespace... so, instead of `doc_id`, something like `"alegre:webhook:#{doc_id}"`, for example.

app/models/bot/alegre.rb
caiosba: Let's please add a prefix to the Redis keys, just for better scoping/debugging if needed, like a namespace... so, instead of just `parsed_response.dig("body", "id")`, something like "alegre:webhook:#{parsed_response.dig("body", "id")}", for example.

caiosba: We do, @computermacgyver !

DGaffney: Duh..... ok I will address in morning!

app/models/concerns/alegre_webhooks.rb
DGaffney: works for me!

app/models/bot/alegre.rb
```diff
@@ -490,7 +496,13 @@ def self.request_api(method, path, params = {}, query_or_body = 'body', retries
       response_body = response.body
       Rails.logger.info("[Alegre Bot] Alegre response: #{response_body.inspect}")
       ActiveRecord::Base.connection.reconnect! if RequestStore.store[:pause_database_connection]
-      JSON.parse(response_body)
+      parsed_response = JSON.parse(response_body)
+      if parsed_response.dig("queue") == 'audio__Model' && parsed_response.dig("body", "callback_url") != nil
+        redis = Redis.new(REDIS_CONFIG)
+        redis_response = redis.blpop("alegre:webhook:#{parsed_response.dig("body", "id")}", 120)
+        return JSON.parse(redis_response[1])
+      end
```
caiosba, line 504: Request: Please add a test for this case.

app/models/concerns/alegre_webhooks.rb
```diff
@@ -0,0 +1,24 @@
...
+      #     "action": action,
+      #     "model_type": model_type,
+      #     "data": data,
+      # })
+      begin
+        doc_id = request.params["data"]["requested"]["body"]["id"]
+        redis = Redis.new(REDIS_CONFIG)
+        key = "alegre:webhook:#{doc_id}"
+        redis.lpush(key, request.params.to_json)
+        redis.expire(key, 1.day.to_i)
```
caiosba, line 18: This block is not tested, but I'll explain why in the next comment.

test/models/bot/alegre_3_test.rb
```diff
@@ -580,4 +580,40 @@ def self.extract_project_medias_from_context(search_result)
   test "should not resort matches if format is unknown" do
     assert_equal 'Foo', Bot::Alegre.return_prioritized_matches('Foo')
   end
+
+  test "should process webhook" do
```
caiosba, line 584: If I'm reading it correctly, looks like this test is not actually doing anything. I suggest you move this test to https://github.com/meedan/check-api/blob/develop/test/controllers/webhooks_controller_test.rb and change it to do the following:

1. Create testing data and stub Alegre response, like you're already doing
2. Be sure that the Redis key is nil (you can use `assert_nil`)
3. Call the webhook endpoint by executing `post :index, params: { name: :alegre }, body: payload` in the test (you'll need to define `payload` similarly to how you are defining `request` as an `OpenStruct` currently) - there are examples in https://github.com/meedan/check-api/blob/develop/test/controllers/webhooks_controller_test.rb
4. Be sure that the key in Redis is not null and contains the expected value (you can use `assert_no_nil` or even `assert_equal`)
5. Use `travel_to` and be sure that the key was removed from Redis

caiosba (changes requested):
Thanks Devin! We're almost there! I added a couple of requests about the tests, but other than that, please also do:

- Rebase this branch against latest `develop`
- Add `alegre_token: dev` to `config/config.yml.example`

test/test_helper.rb
```diff
@@ -151,7 +151,7 @@ def before_all
     @start = Time.now
 
     create_metadata_stuff
-    @exporter = Check::OpenTelemetryTestConfig.current_exporter
+    # @exporter = Check::OpenTelemetryTestConfig.current_exporter
```
caiosba, line 154: Why you commented this line?

test/controllers/webhooks_controller_test.rb
```diff
@@ -232,4 +232,44 @@ def setup
     assert_equal '200', response.code
     assert_match /ignored/, response.body
   end
+
+  test "should process alegre webhook zzz" do
```
caiosba, line 236: Please delete the `zzz` :zzz: from the test name lol

app/models/bot/alegre.rb
```diff
@@ -126,6 +127,12 @@ def get_language_from_transcription
     end
   end
 
+  def self.valid_request?(request)
+    token = request.params["token"] || request.query_parameters["token"]
+    !token.blank? && token == CheckConfig.get('alegre_token')
+  end
```
caiosba, line 133: Please move this to the `AlegreWebhooks` concern.

computermacgyver (approved):
Great work.

caiosba (approved):
I pushed a few cosmetic changes, simplified one test and added one more test. Good to go after Alegre is merged!

PR #1653, opened 2023-09-14
Do not crash when calculating number of items in a list that has an invalid team task filter

caiosba:
## Description

I noticed it happening on live only for duplicated workspaces. When a list has a team task filter that doesn't have a valid ID, an exception would happen. I was able to reproduce that exception in a unit test.

The fix was to discard the team task filter if the ID is blank.

Fixes CV2-3742.

## How has this been tested?

TDD. I was able to implement a unit test that reproduced the issue.

## Checklist

### DIFF
--- a/lib/check_search.rb
+++ b/lib/check_search.rb
@@ -488,8 +488,8 @@ class CheckSearch
 
   def team_tasks_conditions
     conditions = []
-    return conditions unless @options.has_key?('team_tasks') && @options['team_tasks'].class.name == 'Array'
-    @options['team_tasks'].delete_if{ |tt| tt['response'].blank? }
+    return conditions unless @options['team_tasks'].class.name == 'Array'
+    @options['team_tasks'].delete_if{ |tt| tt['response'].blank? || tt['id'].blank? }
     @options['team_tasks'].each do |tt|
       must_c = []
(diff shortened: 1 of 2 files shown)


lib/check_search.rb
```diff
@@ -488,8 +488,8 @@ def file_filter
 
   def team_tasks_conditions
     conditions = []
-    return conditions unless @options.has_key?('team_tasks') && @options['team_tasks'].class.name == 'Array'
-    @options['team_tasks'].delete_if{ |tt| tt['response'].blank? }
+    return conditions unless @options['team_tasks'].class.name == 'Array'
```
melsawy, line 491: why did you remove `@options.has_key?('team_tasks')` from unless condition?
I think we should keep it and only update `delete_if` to include `tt['id'].blank?`

melsawy (changes requested):

lib/check_search.rb
caiosba: @melsawy because I think it's redundant with `@options['team_tasks'].class.name` ... is there is no `team_tasks` key, `@options['team_tasks'].class.name` is going to return `NilClass`

melsawy (approved):

PR #1664, opened 2023-09-24
CV2-3680: Store tipline messages we send

melsawy:
## Description

Store Tipline messages sent to user in background and set a `state = sent`

References: CV2-3680

## How has this been tested?

Added unit tests also enabled tipline locally for testing.

- Etc.

## Checklist


app/models/bot/smooch.rb
```diff
@@ -682,12 +682,37 @@ def self.send_error_message(message, is_supported)
...
+    response
+  end
+
+  def self.store_sent_tipline_message(uid, external_id, sent_at, payload_json, team_id, platform, language)
+    payload = JSON.parse(payload_json)
+    tm = TiplineMessage.new
+    tm.uid = uid
+    tm.state = 'sent'
+    tm.direction = :outgoing
+    tm.language =  language
```
caiosba, line 708: Lint: There are two spaces after `=`

app/models/tipline_message.rb
```diff
@@ -3,7 +3,6 @@ class TiplineMessage < ApplicationRecord
 
   belongs_to :team
```
caiosba, line 5: Please add a validation to be sure that state is one of `sent`, `delivered` or `received`.

db/migrate/20230920055719_add_state_column_to_tipline_message.rb
```diff
@@ -0,0 +1,10 @@
+class AddStateColumnToTiplineMessage < ActiveRecord::Migration[6.1]
+  def change
+    add_column :tipline_messages, :state, :string
+    remove_index :tipline_messages, name: "index_tipline_messages_on_external_id"
+    add_index :tipline_messages, :external_id
```
caiosba, line 5: Why you are removing and re-adding the same index? Is it because the previous one was a unique index and this one is not? Well, if so, yes, we may have the same `external_id` for two `TiplineMessage`'s, but one with `sent` state and another with `delivered` state, for example... so I suggest we replace this `add_index` by a unique index for `external_id, state`. This may need to happen after the `update_all` below.

db/migrate/20230920055719_add_state_column_to_tipline_message.rb
```diff
@@ -0,0 +1,10 @@
...
+    # updated state for existing records
+    TiplineMessage.where(direction: "incoming").update_all(state: 'received')
+    TiplineMessage.where(direction: "outgoing").update_all(state: 'delivered')
```
caiosba, line 8: Just to be safe, please add `state: nil` to these `where`'s.

caiosba (changes requested):

app/models/tipline_message.rb
melsawy: Done

db/migrate/20230920055719_add_state_column_to_tipline_message.rb
melsawy: For now I keep the regular index for `external_id` column only as I found many tests need to fix due to unique index for `external_id, state` but locally I see the `external_id, state` are unique

caiosba: Hopefully there is something in a helper that can reduce the number of changes needed in the tests (probably many are using a hard-coded ID `123456` and we should replace by `random_string`), but it's important to add the unique index because I've seen WhatsApp sending the same notification for the same event so we need to guarantee that these are unique since they are used for statistics.

caiosba (changes requested):
Getting there, Sawy! I just added one more comment about the unique index for `external_id`, `state`.

app/models/bot/smooch.rb
```diff
@@ -696,12 +696,37 @@ def self.send_error_message(message, is_supported)
...
     else
-      self.zendesk_send_message_to_user(uid, text, extra, force, preview_url)
+      response = self.zendesk_send_message_to_user(uid, text, extra, force, preview_url)
     end
+    # store a TiplineMessage
+    external_id = self.get_id_from_send_response(response)
+    sent_at = DateTime.now
+    payload_json = { text: text }.merge(extra).to_json
+    team_id = self.config['team_id'].to_i
+    platform = RequestStore.store[:smooch_bot_platform] || 'unknown'
```
caiosba, line 710: Sawy, I think we are using "Unknown" (with capital "U") as the fallback... please confirm and change.

caiosba (approved):
Sawy, I just left one comment, but you can merge after addressing it.

app/models/bot/smooch.rb
melsawy: Changed to `Unknown`

PR #1673, opened 2023-09-28
3718 – Update how we retry tests

vasconsaurus:
## Description

When running locally we might not want the tests to retry, to do this
we would need to comment out the retry line. We want to make this easier
on the developer

Converting the env variable to an integer because ENV[] on ruby returns a string
https://ruby-doc.org/core-3.0.0/ENV.html

Set TEST_RETRY_COUNT to 3 on Travis because we were using the default
before (which is 3)
https://www.rubydoc.info/gems/minitest-retry/

References: 3718

## How has this been tested?

Changed a test to make it fail.
Ran tests locally without setting TEST_RETRY_COUNT, test failed and didn't retry.
Set the env var inside the container (export TEST_RETRY_COUNT=3) and ran the tests again, it retried 3 times.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ before_script:
 - docker-compose -f docker-compose.yml -f docker-test.yml up -d
 - tail -f log/test.log &
 - until curl --silent -I -f --fail http://localhost:3000 ; do printf .; sleep 1; done
-- docker-compose exec -e TEST_RETRY_COUNT=3 api test/setup-parallel-env.sh
+- docker-compose exec api test/setup-parallel-env.sh
 - docker-compose exec api bundle exec rake assets:precompile
 - sleep 10
 - touch tmp/parallel_runtime_test.log
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,13 +44,13 @@ jobs:
   include:
     - stage: tests
       name: functional-tests
-      script: docker-compose exec -e TEST_RETRY_COUNT=3 PATTERN='models mailers integration workers lib contract' api test/run-tests.sh
+      script: docker-compose exec -e TEST_RETRY_COUNT=3 -e PATTERN='models mailers integration workers lib contract' api test/run-tests.sh
     - stage: tests
       name: unit-tests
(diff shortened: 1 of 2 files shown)


vasconsaurus:
Got two errors here, but they seem unrelated to this

```
# functional-tests

Failure:
GraphqlController3Test#test_should_set_smooch_user_slack_channel_url_in_background 
rails test test/controllers/graphql_controller_3_test.rb:344
```

```
# units-test

Error:
SecurityMailerTest#test_should_send_security_notification:
rails test test/mailers/security_mailer_test.rb:4
```

.travis.yml
```diff
@@ -21,7 +21,7 @@ before_script:
 - docker-compose -f docker-compose.yml -f docker-test.yml up -d
 - tail -f log/test.log &
 - until curl --silent -I -f --fail http://localhost:3000 ; do printf .; sleep 1; done
-- docker-compose exec api test/setup-parallel-env.sh
+- docker-compose exec -e TEST_RETRY_COUNT=3 api test/setup-parallel-env.sh
```
caiosba, line 24: @vasconsaurus , we should pass that environment variable in these `script` lines here, no?

https://github.com/meedan/check-api/blob/develop/.travis.yml#L47-L53

vasconsaurus: yes! missed those, will update

caiosba (approved):
Good to go once tests pass.

vasconsaurus:
@caiosba we still have one failing test (in functional-tests)
```
Failure:
GraphqlController3Test#test_should_set_smooch_user_slack_channel_url_in_background 
Expected: "http://SLKIZZDJXE.com"
  Actual: nil
rails test test/controllers/graphql_controller_3_test.rb:344
```

WDYT?

caiosba:
> @caiosba we still have one failing test (in functional-tests)
> 
> ```
> Failure:
> GraphqlController3Test#test_should_set_smooch_user_slack_channel_url_in_background 
> Expected: "http://SLKIZZDJXE.com"
>   Actual: nil
> rails test test/controllers/graphql_controller_3_test.rb:344
> ```
> 
> WDYT?

It's a flaky one. You can merge if it's the only failure.

PR #1698, opened 2023-10-17
Block tipline users when they send more than X messages in 24 hours.

caiosba:
## Description

"X" is defined by a configuration key but has a default value.

Reference: CV2-3860.

## How has this been tested?

I added a unit test for different cases.

## Checklist

### DIFF
--- a/app/models/tipline_message.rb
+++ b/app/models/tipline_message.rb
@@ -36,7 +36,7 @@ class TiplineMessage < ApplicationRecord
   def verify_user_rate_limit
     rate_limit = CheckConfig.get('tipline_user_max_messages_per_day', 1500, :integer)
     # Block tipline user when they have sent more than X messages in 24 hours
-    if self.state == 'received' && TiplineMessage.where(uid: self.uid, created_at: Time.now.ago(1.day)..Time.now, state: 'received').count > rate_limit 
+    if self.state == 'received' && TiplineMessage.where(uid: self.uid, created_at: Time.now.ago(1.day)..Time.now, state: 'received').count > rate_limit
       Bot::Smooch.block_user(self.uid)
     end
   end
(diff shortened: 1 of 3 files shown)


app/models/tipline_message.rb
```diff
@@ -6,6 +6,8 @@ class TiplineMessage < ApplicationRecord
   validates_presence_of :team, :uid, :platform, :language, :direction, :sent_at, :payload, :state
   validates_inclusion_of :state, in: ['sent', 'received', 'delivered']
 
+  after_commit :verify_user_rate_limit, on: :create
```
skyemeedan, line 9: I'm not familiar with ruby, would we want to verify the rate limit before the commit?

app/models/tipline_message.rb
```diff
@@ -29,6 +31,16 @@ def media_url
     media_url || payload['mediaUrl']
   end
 
+  private
+
+  def verify_user_rate_limit
+    rate_limit = CheckConfig.get('tipline_user_max_messages_per_day', 1500, :integer)
+    # Block tipline user when they have sent more than X messages in 24 hours
+    if self.state == 'received' && TiplineMessage.where(uid: self.uid, created_at: Time.now.ago(1.day)..Time.now, state: 'received').count > rate_limit 
+      Bot::Smooch.block_user(self.uid)
```
skyemeedan, line 40: Will block user generate a log message somewhere so we can find what happened?

skyemeedan (approved):
Looks good to me,  what will the experience be for the user sending the message? will they get an error message?

app/models/tipline_message.rb
caiosba: Ah, it's because once a user is blocked we don't even process their incoming request, so we won't even attempt to save their message.

caiosba:
> Looks good to me, what will the experience be for the user sending the message? will they get an error message?

No, they just won't get any response at all (same behavior as WhatsApp, even when you manually block a contact).

app/models/tipline_message.rb
caiosba: Yes! Here: https://github.com/meedan/check-api/blob/develop/app/models/concerns/smooch_blocking.rb#L23

PR #1725, opened 2023-11-10
CV2-3828 First pass on check api integration updates for audio on alegre

DGaffney:
## Description

This work introduces the beginning of a rewrite of the check-api/alegre integration. We do this by using the sync endpoint on alegre, which does an entire store/search loop using presto within a single query. We will continue that pattern through all other modalities, so this ticket is important in conceptually considering the structure of the new code, which, once all modalities are covered, will replace the existing monolith of code that is the check-api/alegre integration.

References: 3828

## How has this been tested?

I've added what I believe are complete, well documented, and representative unit tests for the entire new set of code. I think that its passing 100% right now, but will make sure if its not!

This code will eventually replace the entire check-api/alegre integration. This should be scrutinized and we should have conversations about whether or not it will be extensible, whether or not it will be easier to work with, and what sorts of business logic it may disrupt. please pick it apart!

## Checklist

### DIFF
--- a/app/lib/smooch_nlu.rb
+++ b/app/lib/smooch_nlu.rb
@@ -79,7 +79,7 @@ class SmoochNlu
           language: language,
         }.merge(context)
       }
-      response = Bot::Alegre.request('get', '/text/similarity/', params)
+      response = Bot::Alegre.request('post', '/text/similarity/search/', params)
 
       # One approach would be to take the option that has the most matches
       # Unfortunately this approach is influenced by the number of keywords per option
--- a/app/lib/smooch_nlu.rb
+++ b/app/lib/smooch_nlu.rb
@@ -50,7 +50,7 @@ class SmoochNlu
       alegre_params = common_alegre_params.merge({ quiet: true })
     end
     # FIXME: Add error handling and better logging
-    Bot::Alegre.request_api(alegre_operation, '/text/similarity/', alegre_params) if alegre_operation && alegre_params
+    Bot::Alegre.request(alegre_operation, '/text/similarity/', alegre_params) if alegre_operation && alegre_params
     keywords
   end
 
@@ -79,7 +79,7 @@ class SmoochNlu
           language: language,
         }.merge(context)
       }
-      response = Bot::Alegre.request_api('get', '/text/similarity/', params)
+      response = Bot::Alegre.request('get', '/text/similarity/', params)
 
       # One approach would be to take the option that has the most matches
       # Unfortunately this approach is influenced by the number of keywords per option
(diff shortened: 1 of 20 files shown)


melsawy:
@DGaffney please fix Code Climate(CC) issues and Travis tests so I can start the review

melsawy (changes requested):
@DGaffney please fix Code Climate(CC) issues and Travis tests so I can start the review

caiosba:
@DGaffney let us know please when integration tests are passing for this branch.

caiosba (changes requested):
Code looks good overall! I tested the branch locally with the following features:

- [x] NLU
- [x] Text similarity matching
- [x] Tipline text similarity search

All of that worked well. @DGaffney , I just have two questions:

(1) [These lines](https://github.com/meedan/check-api/blob/cv2-3828-audio-update/app/models/concerns/alegre_v2.rb#L45-L49) are not covered by the tests, which I think is actually a good thing and a confirmation that all requests from Check API are now using POST instead of GET? If you confirm, can we delete those lines of code?

(2) The method `get_context` is defined in two different places, [here](https://github.com/meedan/check-api/blob/cv2-3828-audio-update/app/models/concerns/alegre_v2.rb#L149) and [here](https://github.com/meedan/check-api/blob/cv2-3828-audio-update/app/models/concerns/alegre_similarity.rb#L120), but both modules are included by the same class. Can we remove one of the two then? According to the coverage report, the one being called is the one in `alegre_v2.rb`, and the one in `alegre_similarity.rb` is being ignored - please review which one should stay since they are slightly different.

DGaffney:
> Code looks good overall! I tested the branch locally with the following features:
> 
> * [x]  NLU
> * [x]  Text similarity matching
> * [x]  Tipline text similarity search
> 
> All of that worked well. @DGaffney , I just have two questions:
> 
> (1) [These lines](https://github.com/meedan/check-api/blob/cv2-3828-audio-update/app/models/concerns/alegre_v2.rb#L45-L49) are not covered by the tests, which I think is actually a good thing and a confirmation that all requests from Check API are now using POST instead of GET? If you confirm, can we delete those lines of code?
> 
> (2) The method `get_context` is defined in two different places, [here](https://github.com/meedan/check-api/blob/cv2-3828-audio-update/app/models/concerns/alegre_v2.rb#L149) and [here](https://github.com/meedan/check-api/blob/cv2-3828-audio-update/app/models/concerns/alegre_similarity.rb#L120), but both modules are included by the same class. Can we remove one of the two then? According to the coverage report, the one being called is the one in `alegre_v2.rb`, and the one in `alegre_similarity.rb` is being ignored - please review which one should stay since they are slightly different.

Good points on both cases! I've modified to accommodate. Thank you!

caiosba (approved):
Thanks Devin for addressing the review comments!

PR #1722, opened 2023-11-07
3884 – Improve seeds script: create more Tipline requests

vasconsaurus:
## Description

This is the first part of ticket 3884.
We were creating claim items with 1 request and other types of items with 0. We want to change that, we want to:
- create enough requests so that we trigger pagination (creating 15 does that)
- keep generating some items with 1 request (we want some variation)
- add requests to different types of items (claims, links, audio, image and video)

References: 3884

## How has this been tested?

By running the seeds script inside the container either adding to a user or creating a new one.

PR #1732, opened 2023-11-20
3993 – Improvements for Check API seed script: Add an item with multiple medias

vasconsaurus:
## Description

We now add an item of each media type with multiple medias – which means it has many confirmed relationships. We do this so the center column on the item page has a good size list to check functionality against. 

As per Brian’s [request](https://github.com/meedan/check-api/pull/1722#issuecomment-1798729043)

References: 3993

## How has this been tested?

Ran the seeds script to create and add data.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -252,8 +252,8 @@ ActiveRecord::Base.transaction do
 
   # 2. Creating Items in different states
   # 2.1 Create medias: claims, audios, images, videos and links
-  # media_data_collection = [ data['Claim'], data['UploadedAudio'], data['UploadedImage'], data['UploadedVideo'], data['Link']]
-  media_data_collection = [ data['Claim'] ]
+  media_data_collection = [ data['Claim'], data['UploadedAudio'], data['UploadedImage'], data['UploadedVideo'], data['Link']]
+  # media_data_collection = [ data['Claim'] ]
   media_data_collection.each do |media_data|
     begin
       media_type = data.key(media_data)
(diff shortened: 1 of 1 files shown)


vasconsaurus:
@brianfleming can you take a look at this one and see if there is anything else you need?

db/seeds.rb
```diff
@@ -112,8 +112,13 @@ def create_claim_description_for_imported_fact_check(user, project, team)
...
 
-def create_confirmed_relationship(parent, child)
-  Relationship.create!(source_id: parent.id, target_id: child.id, relationship_type: Relationship.confirmed_type)
+def create_confirmed_relationship(parent, children)
+  if children.class == Array
+    children.each { |child| Relationship.create!(source_id: parent.id, target_id: child.id, relationship_type: Relationship.confirmed_type)}
+  else
+    child = children
+    Relationship.create!(source_id: parent.id, target_id: child.id, relationship_type: Relationship.confirmed_type)
+  end
```
caiosba, line 121: If I'm reading this correctly, this method is basically handling the case when there can be one or more children?
If so, I suggestion the following rewrite (just a suggestion, no need to change):

```
[children].flatten.each { |child| Relationship.create!(source_id: parent.id, target_id: child.id, relationship_type: Relationship.confirmed_type) }
```

caiosba (approved):
I added a suggestion but not a blocker :)

melsawy (approved):
I left one comment  for simplicity and it's up to you to keep current one or use my suggestion

db/seeds.rb
vasconsaurus: I wondered about this one, will update it ☺️

brianfleming:
so...much...media!!
Thanks @vasconsaurus , love it

highfive

PR #1723, opened 2023-11-07
3884 – Improve seeds script: create more suggestions

vasconsaurus:
## Description

This is the second part of ticket 3884: have at least one item that has enough suggestions so that pagination is triggered. We have now:
- One item of each media type with a confirmed relationship
- One item of each media type with a confirmed relationship and 5 suggested relationships

References: 3884

## How has this been tested?

By running the seeds script inside the container either adding to a user or creating a new one.

## Images
Screenshot 2023-11-07 at 17 47 58
Screenshot 2023-11-07 at 17 48 51

PR #1735, opened 2023-11-21
3994 – Improve seeds file: add variation to requests

vasconsaurus:
## Description

Add some variation to tipline requests as per Brian's (request)[https://github.com/meedan/check-api/pull/1723#issuecomment-1802465942]

References: 3994

## How has this been tested?

Ran the seeds script to create an user, and to add data to the same user.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -168,12 +168,11 @@ def create_tipline_user_and_data(project_media, team)
   fields = {
     smooch_user_id: uid,
     smooch_user_app_id: random_string,
-    smooch_user_data: smooch_user_data.to_json,
-    smooch_report_received: Time.now.to_i
+    smooch_user_data: smooch_user_data.to_json
   }
-
+  
   Dynamic.create!(annotation_type: 'smooch_user', annotated: team, annotator: BotUser.smooch_user, set_fields: fields.to_json)
-
+  
   # Tipline request
   smooch_data = {
     'role': 'appUser',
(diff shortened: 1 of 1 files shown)


vasconsaurus:
@brianfleming another one for you to take a look. I think it's the last one (for now) 🎉

brianfleming:
So good @vasconsaurus I ran it, and now I see so much great stuff!

power

db/seeds.rb
```diff
@@ -125,9 +111,22 @@ def create_project_medias_with_channel(user, project, team, data)
 end
 
 def create_tipline_user_and_data(project_media, team)
+  tipline_message_data = {
+    link: 'https://www.nytimes.com/interactive/2023/09/28/world/europe/russia-ukraine-war-map-front-line.html',
+    audio: 'https://media.smooch.io/apps/5c193e2380a9c90022e40b86/conversations/d445678ac87d2e9f44485e40/j5x5g1L28Y7fDrzU9DSwuwUR/wnHkwjykxOqU3SMWpEpuVzSa.oga',
+    video: 'https://media.smooch.io/apps/5c193e2380a9c90022e40b86/conversations/d445678ac87d2e9f44485e40/HEFL3cuzvNnCDXZd0ElYziN2/AOVFpYOfMm_ssRUizUQhJHDD.mp4',
+    image: 'https://media.smooch.io/apps/5c193e2380a9c90022e40b86/conversations/d445678ac87d2e9f44485e40/sUn5Sq9xRiacbV0Af2qmjl5G/bOoeoeV9zNA51ecial0eWDG6.jpeg',
```
caiosba, line 118: Please replace these by fake URLs (we're moving out of Smooch anyway). The important bit here is just the file extensions, but the URLs don't need to be real.

db/seeds.rb
```diff
@@ -170,14 +169,14 @@ def create_tipline_user_and_data(project_media, team)
     smooch_user_app_id: random_string,
     smooch_user_data: smooch_user_data.to_json
   }
-
+  
   Dynamic.create!(annotation_type: 'smooch_user', annotated: team, annotator: BotUser.smooch_user, set_fields: fields.to_json)
-
+
```
caiosba, line 174: Trailing white space :)

caiosba (changes requested):
Great stuff Manu, I just added two comments. One more note: we're using `sample` in a few places, which means that we won't necessarily get at least of example of each type (status, request links, request channel, etc.). I think it's fine for now, just noting that this is something we may change in the future if we really need to have at least one example of each type.

caiosba (approved):

PR #1745, opened 2023-11-29
Tipline history pagination

dariusk:
I had to add an optional `external_id` filter to the `tipline_messages` function. This is required because I need to be able to associate `TiplineRequest` with `TiplineMessages` and I can't derive a cursor value for `TiplineMessages` from `TiplineRequest` because it's a totally different object so graphql wouldn't track that. The `value_json` in `TiplineRequest` contains a reference to `_id` which matches `external_id` on `TiplineMessages`. This lets me do a filter like the following, where `externalId` is derived from `TiplineRequest` is derived from `ProjectMedia`:

```graphql
tipline_messages(uid:"abc123",externalId:"def456") {
    edges {
      node {
        cursor
        dbid,
      }
    }
}
```

Things to consider:

 - not sure I did this correctly, please let me know
 - I don't understand why rails removed the index on `external_id` -- seems like we would want it since we are searching by it?

### DIFF
--- a/app/graph/types/team_type.rb
+++ b/app/graph/types/team_type.rb
@@ -299,9 +299,14 @@ class TeamType < DefaultObject
 
   field :tipline_messages, TiplineMessageType.connection_type, null: true do
     argument :uid, GraphQL::Types::String, required: true
+    argument :external_id, GraphQL::Types::String, required: false
   end
 
-  def tipline_messages(uid:)
-    TiplineMessagesPagination.new(object.tipline_messages.where(uid: uid).order('id ASC'))
+  def tipline_messages(uid:, external_id: nil)
+    if external_id
+      TiplineMessagesPagination.new(object.tipline_messages.where(uid: uid, external_id: external_id).order('id ASC'))
+    else
+      TiplineMessagesPagination.new(object.tipline_messages.where(uid: uid).order('id ASC'))
+    end
   end
 end
--- a/app/graph/types/team_type.rb
+++ b/app/graph/types/team_type.rb
@@ -302,6 +302,6 @@ class TeamType < DefaultObject
   end
 
   def tipline_messages(uid:)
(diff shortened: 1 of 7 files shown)


caiosba (changes requested):
Looks good to me overall! Yeah, I don't understand why Rails updated your `schema.rb`... this is generated based on your current local database, so maybe you have an old database locally. Please revert that, this is only needed when a database migration is added.

Also, please run `docker-compose run api bundle exec rake lapis:graphql:schema` and push the updated `public/relay.json` and `lib/realy.idl` (actually, Check Web won't be able to use the new argument until `relay.json` is updated).

caiosba (approved):
Looks good! I just added a test for this new case: https://github.com/meedan/check-api/pull/1745/commits/28f70bca1f53ad33068901dd18a37ba6ef4a676a.

This is good to go once the frontend is implemented.

@melsawy please review too.

app/graph/types/team_type.rb
```diff
@@ -299,9 +299,14 @@ def shared_teams
...
   end
 
-  def tipline_messages(uid:)
-    TiplineMessagesPagination.new(object.tipline_messages.where(uid: uid).order('sent_at DESC'))
+  def tipline_messages(uid:, external_id: nil)
+    if external_id
+      TiplineMessagesPagination.new(object.tipline_messages.where(uid: uid, external_id: external_id).order('id ASC'))
+    else
+      TiplineMessagesPagination.new(object.tipline_messages.where(uid: uid).order('id ASC'))
+    end
```
melsawy, line 310: we can simplify this one to be 
```
conditions = { uid:  uid }
conditions[:external_id] = external_id unless external_id.blank?
TiplineMessagesPagination.new(object.tipline_messages.where(conditions).order('id ASC'))
```

melsawy (approved):
Looks good!  just leave one comment to simplify `tipline_messages` method

PR #1751, opened 2023-12-04
CV2-4069 fix a few broken paths that are now mistakenly trying to add items rather than search for them

DGaffney:
## Description

I missed a few paths when changing everything to POST-only - this is a small tweak to accommodate.

References: 4069

## How has this been tested?

I believe this shouldn't change any tests, since all previous tests worked fine and we deployed off that, but will fix if it comes up!

Triple check I got the rest of these?

## Checklist

### DIFF
--- a/app/models/concerns/smooch_search.rb
+++ b/app/models/concerns/smooch_search.rb
@@ -164,7 +164,7 @@ module SmoochSearch
         return [] if media_url.blank?
         media_url = self.save_locally_and_return_url(media_url, type, feed_id)
         threshold = Bot::Alegre.get_threshold_for_query(type, pm)[0][:value]
-        alegre_results = Bot::Alegre.get_items_with_similar_media(media_url, [{ value: threshold }], team_ids, "/#{type}/similarity/")
+        alegre_results = Bot::Alegre.get_items_with_similar_media(media_url, [{ value: threshold }], team_ids, "/#{type}/similarity/search/")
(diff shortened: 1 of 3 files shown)


caiosba (approved):
Looks good to me. I reviewed all cases from here:

```
check-api$ vim -p $(grep -rl '/similarity' app lib config)
```

Let's just wait for CI before merging since maybe some mocks need to be updated.

skyemeedan (approved):
Is there a test that covers this behavior?

DGaffney:
> Is there a test that covers this behavior?

Presumably, the functionality that tests these exact functions should be stubbed at the request level rather than at the `Bot::Alegre` level which is where I think they're stubbed currently. I'd rather get this out and follow up with a better test post deploy - @caiosba how would you like to proceed?

caiosba:
I think we're good to go. The tests here are mocked, the real tests are on the integration tests side.

caiosba:
> > the real tests are on the integration tests side.
> > but since we deployed the bug without test failure, it seems like they don't cover this problem?

Yes, we don't. This is a problem we currently have... we don't really test (integration test, end-to-end) the tipline interactions (mainly because it would be hard to automate remote-control of WhatsApp Web, for example). This is why the bug was just identified from real tipline requests and not captured by the current integration test (which covers items created using the web client, in which the bug doesn't happen) or by the current unit tests (which mock the Alegre responses).

PR #1755, opened 2023-12-08
Remove the hardcoded minimum number of words to run a full similarity search

computermacgyver:
## Description

We have a setting in Check Web that specifies the number of words below which we should run a keyword search rather than a full similarity search with Alegre. Unfortunately, we also hardcode a limit in Check API as three words. This leads to the bad experience where when the Check Web setting is lower than three no results are returned. 

E.g., if the Check Web setting is set at 1, then two word queries return no results. 

This ticket removes the hardcoded minimum.

References: CV2-4085

## How has this been tested?

Updated automated tests. Will test end-to-end after merge

## Checklist


computermacgyver:
Merged per discussion with @caiosba . Will verify on QA once this deploys

PR #1763, opened 2023-12-18
[Refactor] 4067 – remove unused mutations

vasconsaurus:
## Description

While listing mutations for the Pentest, Caio realized that some mutations are not in use and could be deleted. 
Steps taken:
- we listed the unused mutations (on the jira ticket)
- deleted from `app/graph/types/mutation_type.rb`
- deleted respective mutation files and definitions
- deleted tests that used the unused mutations
- after each mutation was deleted we:
    - updated relay
    - made sure check-web compiled correctly 
- confirmed that the mutations are not being used by external API clients by checking Cloudwatch logs
    - found two mutations that WERE being used: `smoochBotAddSlackChannelUrl` and `destroyFactCheck`
    - so I rebased and dropped the commits specific to those two and added back a test related to the smooch mutation

After this PR is merged, I'll merge updates to the [check wiki](https://github.com/meedan/check/wiki/GraphQL-query-examples).
- _Delete item_ was removed
- _Create a task in a specific item_ was removed

References: 4067

## How has this been tested?

Ran the unit and integration tests.


caiosba (approved):
Code deletion FTW

PR #1767, opened 2023-12-22
4151 – Remove is_admin from user creation in seeds script

vasconsaurus:
## Description

`is_admin` means the user is a super admin. Which we don’t want.
We removed `is_admin` attributes from the seed script.

References: 4151, 4122 ([PR #1781](https://github.com/meedan/check-web/pull/1781#issuecomment-1867805893))

## How has this been tested?

Ran the seeds.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -272,7 +272,7 @@ ActiveRecord::Base.transaction do
   # 1. Creating what we need for the workspace
   # We create a user, team and project OR we fetch one
   if answer == "1"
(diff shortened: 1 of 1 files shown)


vasconsaurus:
@caiosba is this all that is needed for this one?

caiosba (approved):
Great. I think it makes more sense. This way we're able to properly test different roles and permissions.

caiosba:
> @caiosba is this all that is needed for this one?

Yes, because AFAIU these are the only places in the seed script where users are created, right?

vasconsaurus:
> > @caiosba is this all that is needed for this one?
> 
> Yes, because AFAIU these are the only places in the seed script where users are created, right?

yes 🙂

PR #1777, opened 2024-01-11
Add rate limiting to login requests

jayjay-w:
## Description

Add rate limiting to login requests using rack-attack. As a starting point, we are setting a limit of 5 request per minute to /api/users/sign_in.

References: CV2-4164

## How has this been tested?

Tested by confirming that:

 - Repeated calls to /api/users/sign_in will result in a `429: Too many requests` error. This can be validated using both the check web application or an api client. . The default limit is 10 requests per minute.
- Repeated calls to /api/graphql will result in a `429` error as well. The default limit is 100 requests per minute.

## Checklist

### DIFF
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -78,6 +78,4 @@ Rails.application.configure do
   config.after_initialize do
     PaperTrail.enabled = ENV['PAPERTRAIL_ENABLED'] || false
   end
-
-  config.cache_store = :memory_store
 end
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -78,4 +78,6 @@ Rails.application.configure do
   config.after_initialize do
     PaperTrail.enabled = ENV['PAPERTRAIL_ENABLED'] || false
   end
+
+  config.cache_store = :memory_store
 end
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -23,7 +23,8 @@ Rails.application.configure do
 
   # Show full error reports and disable caching.
   config.consider_all_requests_local       = true
-  config.action_controller.perform_caching = false
+  config.action_controller.perform_caching = true
+  config.cache_store = :memory_store
 
   # Raise exceptions instead of rendering exception templates.
   config.action_dispatch.show_exceptions = false
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -1,13 +1,13 @@
 class Rack::Attack
   # Throttle login attempts by IP address
-  throttle('logins/ip', limit: ChheckConfig.get('login_rate_limit', 10).to_i, period: 60.seconds) do |req|
+  throttle('logins/ip', limit: CheckConfig.get('login_rate_limit', 10, :integer), period: 60.seconds) do |req|
     if req.path == '/api/users/sign_in' && req.post?
       req.ip
     end
   end
 
   # Throttle login attempts by email address
-  throttle('logins/email', limit: ChheckConfig.get('login_rate_limit', 10).to_i, period: 60.seconds) do |req|
+  throttle('logins/email', limit: CheckConfig.get('login_rate_limit', 10, :integer), period: 60.seconds) do |req|
     if req.path == '/api/users/sign_in' && req.post?
       # Return the email if present, nil otherwise
       req.params['user']['email'].presence if req.params['user']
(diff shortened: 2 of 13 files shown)


config/environments/test.rb
```diff
@@ -78,4 +78,6 @@
   config.after_initialize do
     PaperTrail.enabled = ENV['PAPERTRAIL_ENABLED'] || false
   end
+
+  config.cache_store = :memory_store
```
computermacgyver, line 82: I know very little about Ruby/Rails and even less about the rack-attack gem, but I'm guessing this line means it is storing attempts in local RAM. As we run multiple instances of `check-api` in production, each will have their own RAM. This means that the limits set in `Rack::Attack` will not be strictly accurate. E.g., we want no more than 5 login attemps from a single ip in 60 seconds, but if we are running 3 instances of `check-api` an attacker might get up to 3 x 5 = 15 attempts in 60 seconds before being throttled. Is there an option to store in Redis? I'm not sure what impact that could have on performance, however.

computermacgyver: I guess this line is actually about the test environment, where RAM is fine. I'm not sure how it's configured for QA/Live.

jayjay-w: Yes this is only on the test environment, but it will probably go away. Currently the rate limiting is not working on test, and I am trying to figure out why. Something seems to be disabling Rack::Attack. This line here was just one of the attempts to get it to work.

config/initializers/rack_attack.rb
```diff
@@ -0,0 +1,21 @@
+class Rack::Attack
+  # Throttle login attempts by IP address
+  throttle('logins/ip', limit: ChheckConfig.get('login_rate_limit', 10).to_i, period: 60.seconds) do |req|
```
caiosba, line 3: Typo: `Chheck`

caiosba, line 3: Typo: `Chheck`

config/initializers/rack_attack.rb
```diff
@@ -0,0 +1,21 @@
...
+    if req.path == '/api/users/sign_in' && req.post?
+      req.ip
+    end
+  end
+
+  # Throttle login attempts by email address
+  throttle('logins/email', limit: ChheckConfig.get('login_rate_limit', 10).to_i, period: 60.seconds) do |req|
```
caiosba, line 10: Typo: `Chheck`

config/initializers/rack_attack.rb
```diff
@@ -0,0 +1,21 @@
...
+    if req.path == '/api/users/sign_in' && req.post?
+      # Return the email if present, nil otherwise
+      req.params['user']['email'].presence if req.params['user']
+    end
+  end
+
+   # Throttle all graphql requests by IP address
+   throttle('api/graphql', limit: ChheckConfig.get('api_rate_limit', 100).to_i, period: 60.seconds) do |req|
```
caiosba, line 18: Typo: `Chheck`

caiosba (changes requested):
There is a typo, but after I fixed it locally, I was able to verify both features:

- User account locked after failed login attempts
- Rate limit error when many calls are made to the GraphQL API endpoint

FYI @jayjay-w: You don't need to explicitly call `.to_i` when retrieving a configuration value... you can pass a `:integer` parameter to `CheckConfig.get`: https://github.com/meedan/check-api/blob/develop/app/lib/check_config.rb#L10

config/initializers/rack_attack.rb
jayjay-w: Thank you, I missed that

test/lib/check_rack_attack_test.rb
```diff
@@ -0,0 +1,31 @@
...
+    limit = 100
+
+    limit.times do
+      post api_graphql_path
+      assert_response :unauthorized
+    end
+
+    get api_graphql_path
+    assert_response :too_many_requests
+  end
```
caiosba, line 18: @jayjay-w , I also suggest that you use `stub_config` here (there are examples in code) so that you don't need to test with 100 requests (you can test with 3, for example)... I think this is impacting how long the tests take.

jayjay-w: Thank you, this is fixed as well!

caiosba (approved):

caiosba:
hey @jayjay-w , there was a failed test :)

```
Error:
SessionsControllerTest#test_should_unlock_locked_user_accounts_after_specified_time:
TypeError: ActiveSupport::TimeWithZone can't be coerced into Integer
    test/controllers/sessions_controller_test.rb:97:in `+'
    test/controllers/sessions_controller_test.rb:97:in `block in <class:SessionsControllerTest>'
```

jayjay-w:
Interesting...@caiosba I will fix it in a new PR.

PR #1798, opened 2024-02-09
4273 – Refactor API Seeds

vasconsaurus:
## First

Apologies as this got way too big, I probably should have broken it up by action (creating relationships, tipline requests etc)

## Description

Our old seeds script was built to create items for one user, once the need to have more users appeared there wasn't an easy way to update it. 

This file separates everything into two objects: **setup** and **populated_workspaces**.

**[1] Setup** 
- creates the users, teams and team_users
    - 3 users
    - 4 teams 
- t takes the approach of looking into creating those instances as configuration, so it's more verbose, but I think it's easier to understand what is being created.

**[2] Populated Workspaces**
- we initialize a new populated workspace with the value from Setup
- then with populate_projects we use nested_attributes to create projects, project medias, claim descriptions and fact checks
- after that we have other methods we can call on our instance to run actions to change the state in our workspace:
    - populated_workspaces.saved_searches
    - populated_workspaces.main_user_feed
    - populated_workspaces.share_feed
    - populated_workspaces.confirm_relationships
    - populated_workspaces.suggest_relationships
    - populated_workspaces.tipline_requests
    - populated_workspaces.publish_fact_checks
    - populated_workspaces.clusters 

References: 4273, 4174

## How has this been tested?

By running the seeds to create a new user and add a new user.

## Notes
- This script is actually faster than the old one, but it creates more users and teams than before, so it can be quite slow.
    - I think with more people testing this we can better understand a balance between creating all the things vs time.
    - Right now it creates items from all media types for the main user, but skips creating links for the invited users (this drops the time it takes to run from 10 minutes to 6)
    - For now there are ways to edit it to create less:
        -  Users: To create only one user, comment lines to skip adding invited users
           - 77 (`Setup.users.all_users.merge!(invited_users)`) 
           - 87 (`Setup.teams.all_teams.merge!(invited_teams)`)
       - Items: Play around with how many items get created, there are different ways to do this:
          - you can edit `MEDIAS_PARAMS`, comment out type of medias you do not want to create
          - you can edit what gets passed to each project in populate_projects, look for project_media_attributes inside project_params (there is one for `projects_params_main_user_a` and `project_params_invited_users`)
      - Hopefully we can make all this more dynamic for the user, so they can just input what they want to be created
    - Be mindful of sidekiq queues, a lot get created
        - keep track here: http://localhost:3000/sidekiq
        - if for whatever reason you want to run it again right away, you should probably clear the queues (they add up). In the rails console run `Sidekiq::Queue.all.map(&:clear)`
- It does not add imported fact checks yet

I tried to summarize what I thought was important, if there are any questions please let me know.
I know it's a lot, so I'm sure there is a bunch of confusing stuff in there.


brianfleming:
FYI @vasconsaurus I pushed up a merge with develop to try to clear up those bundle message we chatted about

brianfleming:
love it @vasconsaurus :

```
—————
Stretch your legs, this might take a while.
On a mac took about 10 minutes to create all populated workspaces.
...
```

brianfleming (commented):
This ran without errors for me, poking around the generated workspaces. Will let the experts comment on the code

brianfleming:
Just so I understand the expectations of the data @vasconsaurus
are the counts media, requests, fact-checks just random numbers? Meaning they are not back by actual data correct? I'm guessing thats the case since the number of fact-checks matches the other number, which isn't realistic.

Screenshot 2024-02-29 at 4 07 57 PM

caiosba:
That's correct, @brianfleming - for now those are just random numbers because we currently only list the cards. In the next sprint, when we have the item page, the numbers will reflect real data.

caiosba (approved):
Thanks @vasconsaurus , the code is way more clear, great refactoring!

We don't need to keep this file, right? -> `db/seeds_backup.rb`? Can we just delete it?

vasconsaurus:
> Thanks @vasconsaurus , the code is way more clear, great refactoring!
> 
> We don't need to keep this file, right? -> `db/seeds_backup.rb`? Can we just delete it?

yes! will delete it and merge this

PR #1823, opened 2024-03-08
4380 – Remove create_report method from the FetchBot

vasconsaurus:
## Description
This method was causing an issue and is not needed anymore.
We can simply substitute it by `fc.publish_report = true`

## Issue:
When trying to import some Fact Checks, using the `FetchBot`,
with language set to `pt_BR`, the language was still being set as `en`.

That was happening because inside `import_claim_review` after it runs
`set_claim_and_fact_check`, it was running `create_report`.
https://github.com/meedan/check-api/blob/4a44ab093fbae68e25114c4cef0af537daec9f2b/app/models/bot/fetch.rb#L187-L199

Inside `create_report` it was setting the language to the workspace’s
default language. So, we might want to add a language as `pt_BR`,
but if the default language of the workspace is `en`, that gets overwritten.
https://github.com/meedan/check-api/blob/4a44ab093fbae68e25114c4cef0af537daec9f2b/app/models/bot/fetch.rb#L338

References: 4380, 4347

## How has this been tested?

By creating imported fact checks using the Fetch Bot.

## Things to pay attention to

We had one failing test: `"should not import duplicate claim reviews"`. We got an error `#<RuntimeError: No permission to create Dynamic>` when trying to save a `FactCheck`, it was rolled back and didn't create the `FactCheck` and `ProjectMedia`, causing the test to fail. 

This was fixed by adding `reports.skip_check_ability = true` to `update_report`. 
https://github.com/meedan/check-api/blob/e681cf2388da208e7286e7135c82bdb14f7d2c07/app/models/fact_check.rb#L76

Basically, we don’t really need to check for permissions in the `FactCheck` `update_report` method, since it’s a private method that is just called from a `FactCheck` callback.

## Checklist

### DIFF
--- a/app/models/bot/fetch.rb
+++ b/app/models/bot/fetch.rb
@@ -196,7 +196,6 @@ class Bot::Fetch < BotUser
             self.set_status(claim_review, pm, status_fallback, status_mapping)
             self.set_analysis(claim_review, pm)
             self.set_claim_and_fact_check(claim_review, pm, user, team)
-            self.create_report(claim_review, pm, team, user, auto_publish_reports)
             self.create_tags(claim_review, pm, user)
           end
         end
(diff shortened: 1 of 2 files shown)


caiosba (approved):
Thanks, Manu. And very good documentation in the PR :star: 

Good to go if tests pass!

PR #1834, opened 2024-03-22
4346 – Seeds script: create cluster data with real values

vasconsaurus:
## Description

We updated how we create the clusters in the seeds script, so it will:
- add all the cluster values (some were missing from the prior version)
- add the real values (some were random values, now only channel is random)

References: 4346

## How has this been tested?

By running the script to create all the data, and then running it to add data to an user.

- Currently all the cluster's items are from the same team. We decide to finish this first version first, and update it to add the other team's items to the cluster on a next ticket/PR.

### User A / Team A*
Screenshot 2024-03-22 at 16 25 57
Screenshot 2024-03-22 at 16 26 10
Screenshot 2024-03-22 at 16 34 51
Screenshot 2024-03-22 at 16 35 06

_* I ran the script twice, so that is why User A has 4 feeds instead of 2._

### User C / Team C
Screenshot 2024-03-22 at 16 26 49
Screenshot 2024-03-22 at 16 27 09

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -31,9 +31,9 @@ end
 
 MEDIAS_PARAMS = [
   *CLAIMS_PARAMS,
-  *UPLOADED_AUDIO_PARAMS,
-  *UPLOADED_IMAGE_PARAMS,
-  *UPLOADED_VIDEO_PARAMS,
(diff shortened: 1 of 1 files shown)


caiosba (approved):
Very clean! Great work, Manu!

PR #1836, opened 2024-03-26
CV2-4350: improve load claims and bulk-update clusters

melsawy:
## Description

Improve clusterization rake task

- Do not load all unique claim media items in memory 
- Update clusters in batches instead of one by one

References: CV2-4350

## How has this been tested?

Verified rake task against live DB before/after my changes.

- Etc.

## Checklist


lib/tasks/check_khousheh.rake
```diff
@@ -285,8 +294,16 @@ namespace :check do
...
-                    # FIXME: Update clusters in batches
-                    cluster.update_columns(updated_cluster_attributes)
+                    cluster_items[cluster.id] = updated_cluster_attributes
+                  end
+                end
+                # Bulk-update Cluster
+                unless cluster_items.blank?
+                  begin
+                    cluster_items_values = cluster_items.values.to_a
+                    Cluster.upsert_all(cluster_items_values, unique_by: :id)
```
caiosba, line 304: @melsawy , this won't work unless you change also how `updated_cluster_attributes` is calculated... between lines 266 - 277 (on `develop`) you can see that `updated_cluster_attributes` are calculated based on the current `cluster` instance, which used to be updated at every iteration... now that you're planning to bulk-update clusters, you need to change these lines to read from the updated cluster instance in memory, not from the database. Makes sense? Let's talk if it's not clear.

caiosba (changes requested):

caiosba (commented):
@melsawy , I noticed two differences:

- For the input file, the edges were the same, but not the nodes... this version generated 3 nodes less than `develop`
- For the clusters, this version created 3 clusters less than the version on `develop` (using the same output file as `develop`)

Do you have any idea why?

caiosba (approved):
Thanks Sawy, I tested it locally now and it works well. Good to merge after fixing the last CC issue.

PR #1838, opened 2024-03-27
Cv2 4269 check api stop indexing titles like platform-team-012345678-90

ahmednasserswe:
## Description

Stop indexing titles like platform-team-012345678-90 and add related tests.

References: CV2-4269

## How has this been tested?

Locally and unit tests added

## Checklist

- [✔ ] I have performed a self-review of my own code
- [ ✔] I have added unit and feature tests, if the PR implements a new feature or otherwise would benefit from additional testing
- [ ✔] I have added logging, exception reporting, and custom tracing with any additional information required for debugging
- [✔ ] I considered secure coding practices when writing this code. Any security concerns are noted above.
- [✔ ] I have commented my code in hard-to-understand areas, if any
- [non needed ] I have made needed changes to the README
- [✔ ] My changes generate no new warnings

[CV2-4269]: https://meedan.atlassian.net/browse/CV2-4269?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/bot/alegre.rb
+++ b/app/models/bot/alegre.rb
@@ -20,7 +20,7 @@ class Bot::Alegre < BotUser
 
   REPORT_TEXT_SIMILARITY_FIELDS = ['report_text_title', 'report_text_content', 'report_visual_card_title', 'report_visual_card_content']
   ALL_TEXT_SIMILARITY_FIELDS = REPORT_TEXT_SIMILARITY_FIELDS + ['original_title', 'original_description', 'extracted_text', 'transcription', 'claim_description_content', 'fact_check_title', 'fact_check_summary']
-
+  BAD_TITLE_REGEX = /^[a-z\-]+-[0-9\-]+$/
   ::ProjectMedia.class_eval do
     attr_accessor :alegre_similarity_thresholds, :alegre_matched_fields
 
@@ -197,7 +197,7 @@ class Bot::Alegre < BotUser
 
   def self.get_items_from_similar_text(team_id, text, fields = nil, threshold = nil, models = nil, fuzzy = false)
     team_ids = [team_id].flatten
-    if text.blank? || /[a-z\-]+-[0-9\-]+/ =~ text
+    if text.blank? || BAD_TITLE_REGEX =~ text
       Rails.logger.info("[Alegre Bot] get_items_from_similar_text returning early due to blank/bad text #{text}")
       return {}
     end
@@ -296,8 +296,8 @@ class Bot::Alegre < BotUser
 
   def self.get_language_from_text(pm, text)
     lang = 'und'
-    if !text.blank? && /[a-z\-]+-[0-9\-]+/ !~ text
-        lang = self.get_language_from_alegre(text)
+    if !text.blank? && BAD_TITLE_REGEX !~ text
+      lang = self.get_language_from_alegre(text)
     end
     self.save_annotation(pm, 'language', { language: lang })
     lang
--- a/app/models/bot/alegre.rb
+++ b/app/models/bot/alegre.rb
@@ -196,21 +196,19 @@ class Bot::Alegre < BotUser
   end
 
   def self.get_items_from_similar_text(team_id, text, fields = nil, threshold = nil, models = nil, fuzzy = false)
-    if /[a-z\-]+-[0-9\-]+/ !~ text
-        team_ids = [team_id].flatten
-        if text.blank?
-          Rails.logger.info("[Alegre Bot] get_items_from_similar_text returning early due to blank text #{text}")
-          return {}
-        end
-        fields ||= ALL_TEXT_SIMILARITY_FIELDS
-        threshold ||= self.get_threshold_for_query('text', nil, true)
-        models ||= [self.matching_model_to_use(team_ids)].flatten
-        Hash[self.get_similar_items_from_api(
-          '/text/similarity/search/',
-          self.similar_texts_from_api_conditions(text, models, fuzzy, team_ids, fields, threshold),
-          threshold
-        ).collect{|k,v| [k, v.merge(model: v[:model]||Bot::Alegre.default_matching_model)]}]
+    team_ids = [team_id].flatten
+    if text.blank? || /[a-z\-]+-[0-9\-]+/ =~ text
+      Rails.logger.info("[Alegre Bot] get_items_from_similar_text returning early due to blank/bad text #{text}")
+      return {}
     end
+    fields ||= ALL_TEXT_SIMILARITY_FIELDS
+    threshold ||= self.get_threshold_for_query('text', nil, true)
+    models ||= [self.matching_model_to_use(team_ids)].flatten
+    Hash[self.get_similar_items_from_api(
+      '/text/similarity/search/',
+      self.similar_texts_from_api_conditions(text, models, fuzzy, team_ids, fields, threshold),
+      threshold
+    ).collect{|k,v| [k, v.merge(model: v[:model]||Bot::Alegre.default_matching_model)]}]
   end
 
   def self.unarchive_if_archived(pm)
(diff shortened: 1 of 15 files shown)


test/models/bot/alegre_4_test.rb
```diff
@@ -0,0 +1,86 @@
...
+  end
+
+  def teardown
+    super
+  end
+
+  test "should not send bad text for langid" do
+    text = "platform~1234"
+    pm1 = create_project_media team: @team, quote: text
+    Bot::Alegre.stubs(:get_language_from_alegre).returns("und")
```
computermacgyver, line 25: This line should stub to be "en" NOT "und" . The goal is to test that `get_language_from_alegre` is **not** called. I  understand the test fails as it was originally written, and that is because the current `text = "platform~1234"` does not match the correct pattern. Updating the `text` value is the correct way to solve this not changing the subbed returned value

test/models/bot/alegre_4_test.rb
```diff
@@ -0,0 +1,86 @@
...
+  test "/text/similarity/search/ request should not be called for blank title" do
+    text = ""
+    pm1 = create_project_media team: @team, quote: text
+    Bot::Alegre.stubs(:request).raises("Request method called when it should not be")
+    assert_nothing_raised do
+        Bot::Alegre.get_items_from_similar_text(@team, text)
+    end
+    Bot::Alegre.unstub.stubs(:request)
+  end
+
```
computermacgyver, line 65: Please add a test with a valid text value that should make it through the if statement and call  `request`. We can use `assert_raises` (https://www.rubydoc.info/gems/minitest/Minitest%2FAssertions:assert_raises) and the same pattern as the other tests. This test should go with the other tests for `get_items_from_similar_text` at https://github.com/meedan/check-api/blob/801d44b98a33d60e61e29f3da91512ccf11efaff/test/models/bot/alegre_2_test.rb#L525-L527

test/models/bot/alegre_4_test.rb
```diff
@@ -0,0 +1,86 @@
...
+  test "/text/similarity/ request should not be called for blank title" do
+    text = ""
+    pm1 = create_project_media team: @team, quote: text
+    Bot::Alegre.stubs(:request).raises("Request method called when it should not be")
+    assert_nothing_raised do
+        Bot::Alegre.send_to_text_similarity_index(pm1, create_field_instance, text, Bot::Alegre.item_doc_id(pm1, create_field_instance))
+    end
+    Bot::Alegre.unstub.stubs(:request)
+  end
+
```
computermacgyver, line 85: As above, please add a test for a valid text value that will call `request`

test/models/bot/alegre_4_test.rb
```diff
@@ -0,0 +1,86 @@
...
+  test "should send text for langid" do
+    text = "anything else"
+    pm1 = create_project_media team: @team, quote: text
+    Bot::Alegre.stubs(:get_language_from_alegre).returns("en")
+    assert_equal Bot::Alegre.get_language_from_text(pm1, text), "en"
+    Bot::Alegre.unstub(:get_language_from_alegre)
+  end
+
+  test "/text/similarity/search/ request should not be called for bad title" do
```
computermacgyver, line 46: rename test to "get_items_from_similar_texts should not search bad titles" Rename other tests similarly, please.

computermacgyver: Also please move this test to below https://github.com/meedan/check-api/blob/801d44b98a33d60e61e29f3da91512ccf11efaff/test/models/bot/alegre_2_test.rb#L525-L527 per the JIRA ticket instructions given there is already a test for blank strings there.

test/models/bot/alegre_4_test.rb
```diff
@@ -0,0 +1,86 @@
...
+    text = "instagram-cekfakta-2023-25562004"
+    pm1 = create_project_media team: @team, quote: text
+    Bot::Alegre.stubs(:request).raises("Request method called when it should not be")
+    assert_nothing_raised do
+        Bot::Alegre.get_items_from_similar_text(@team, text)
+    end
+    Bot::Alegre.unstub.stubs(:request)
+  end
+
+  test "/text/similarity/search/ request should not be called for blank title" do
```
computermacgyver, line 56: Can remove this test as it is the same test as https://github.com/meedan/check-api/blob/801d44b98a33d60e61e29f3da91512ccf11efaff/test/models/bot/alegre_2_test.rb#L525-L527

app/models/bot/alegre.rb
```diff
@@ -295,7 +295,10 @@ def self.get_language(pm)
   end
 
   def self.get_language_from_text(pm, text)
-    lang = text.blank? ? 'und' : self.get_language_from_alegre(text)
+    lang = 'und'
+    if !text.blank? && /[a-z\-]+-[0-9\-]+/ !~ text
+        lang = self.get_language_from_alegre(text)
```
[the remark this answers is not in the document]
computermacgyver: Please adjust accordingly here and the other location the same code climate comment was made.

computermacgyver (changes requested):

app/models/bot/alegre.rb
```diff
@@ -197,8 +197,8 @@ def self.get_number_of_words(text)
 
   def self.get_items_from_similar_text(team_id, text, fields = nil, threshold = nil, models = nil, fuzzy = false)
     team_ids = [team_id].flatten
-    if text.blank?
-      Rails.logger.info("[Alegre Bot] get_items_from_similar_text returning early due to blank text #{text}")
+    if text.blank? || /[a-z\-]+-[0-9\-]+/ =~ text
```
caiosba, line 200: I saw this regular expression being used in different places - so, if we're using it more than once, I suggest we save it to a constant, like we do with these constants: https://github.com/meedan/check-api/blob/develop/app/models/bot/alegre.rb#L12-L22. This way, if we need to change, there is no risk that we change in one place and not in another.

Also, I think we want exact matches, right? Like, "A very long title that contains this-123" would be valid according to this regular expression, but really we just wanted to match if it was "this-123", correct? Then, the regular expression should be `/^[a-z\-]+-[0-9\-]+$/`.

caiosba (changes requested):

app/models/bot/alegre.rb
ahmednasserswe: thats true, thanks for the fix.

test/models/bot/alegre_2_test.rb
```diff
@@ -526,6 +526,36 @@ def teardown
...
+    text = "This should pass to Alegre"
+    pm1 = create_project_media team: @team, quote: text
+    Bot::Alegre.stubs(:request).raises("Request method called when it should not be")
+    assert_raises do
+        Bot::Alegre.get_items_from_similar_text(@team, text)
+    end
+    Bot::Alegre.unstub.stubs(:request)
+  end
+
+  test "/text/similarity/ request should be called for good title" do
```
computermacgyver, line 549: This method should be in `alegre_4_test.rb` with the other tests that use `send_to_text_similarity_index`

computermacgyver (changes requested):

test/models/bot/alegre_4_test.rb
```diff
@@ -0,0 +1,65 @@
...
+  test "should not send blank text for langid" do
+    text = ""
+    pm1 = create_project_media team: @team, quote: text
+    Bot::Alegre.stubs(:get_language_from_alegre).returns("en")
+    assert_equal Bot::Alegre.get_language_from_text(pm1, text), "und"
+    Bot::Alegre.unstub(:get_language_from_alegre)
+  end
+
+  test "should send text for langid" do
+    text = "anything else"
```
computermacgyver, line 39: Let's change this to `text = "instagram-cekfakta-2023-25562004 with more should pass"

ahmednasserswe: changed the test to "instagram-cekfakta-2023-25562004 with more should pass" in the other good title tests too.

caiosba:
@melsawy please review this.

melsawy (approved):

computermacgyver (approved):

caiosba (approved):

PR #1845, opened 2024-04-02
CV2-4432: Fix sentry issue

melsawy:
## Description

Fix sentry issue and make code more readable.

References: CV2-4432

## How has this been tested?

Depend on current tests.

- Etc.

## Checklist

### DIFF
--- a/app/controllers/api/v1/admin_controller.rb
+++ b/app/controllers/api/v1/admin_controller.rb
@@ -67,7 +67,11 @@ class Api::V1::AdminController < Api::V1::BaseApiController
     tbi = TeamBotInstallation.find(params[:id])
     auth = session['check.facebook.authdata']
     status = nil
-    if params[:token].to_s.gsub('#_=_', '') == tbi.get_smooch_authorization_token
+    if auth.blank?
+      status = 400
+      @message = I18n.t(:invalid_facebook_authdata)
+      CheckSentry.notify('Could not authenticate Facebook account for tipline Messenger integration.')
+    elsif params[:token].to_s.gsub('#_=_', '') == tbi.get_smooch_authorization_token
       q_params = {
         client_id: CheckConfig.get('smooch_facebook_app_id'),
         client_secret: CheckConfig.get('smooch_facebook_app_secret'),
--- a/app/controllers/api/v1/admin_controller.rb
+++ b/app/controllers/api/v1/admin_controller.rb
@@ -70,7 +70,8 @@ class Api::V1::AdminController < Api::V1::BaseApiController
     if auth.blank?
       status = 400
       @message = I18n.t(:invalid_facebook_authdata)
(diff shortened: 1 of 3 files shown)


caiosba (changes requested):
Sawy, I suggest a different approach here: when `auth` is `nil`, is because something went wrong with the Facebook authentication. In this case:

- We want to be sure that the user _won't_ see a success page
- We want to be sure that we're still notified by Sentry

So my suggestion is:

- If `auth` is `nil`, set a message and the status as 400, like we do here: https://github.com/meedan/check-api/blob/5218a065feda97420abe3b2663e04c715c63dab8/app/controllers/api/v1/admin_controller.rb#L79-L81
- The `@message` can be something like "It was not possible to authenticate your Facebook account, please reach out to our support team." (must be a localizable string).
- Notify Sentry with a message like "Could not authenticate Facebook account for tipline Messenger integration."

But I agree with the rewrite you did for the query parameters, it's more readable.

app/controllers/api/v1/admin_controller.rb
```diff
@@ -67,8 +67,18 @@ def save_facebook_credentials_for_smooch_bot(platform) # "platform" is either "i
     tbi = TeamBotInstallation.find(params[:id])
     auth = session['check.facebook.authdata']
     status = nil
-    if params[:token].to_s.gsub('#_=_', '') == tbi.get_smooch_authorization_token
-      response = Net::HTTP.get_response(URI("https://graph.facebook.com/me/accounts?client_id=#{CheckConfig.get('smooch_facebook_app_id')}&client_secret=#{CheckConfig.get('smooch_facebook_app_secret')}&access_token=#{auth['token']}&limit=100"))
+    if auth.blank?
+      status = 400
+      @message = I18n.t(:invalid_facebook_authdata)
+      CheckSentry.notify('Could not authenticate Facebook account for tipline Messenger integration.')
```
caiosba, line 73: Sawy, please add more information to the Sentry notification (not to the message, but as a second hash parameter) so we know to which workspace it refers to (maybe at least the `TeamBotInstallation` ID).

app/controllers/api/v1/admin_controller.rb
```diff
@@ -67,8 +67,18 @@ def save_facebook_credentials_for_smooch_bot(platform) # "platform" is either "i
...
+    elsif params[:token].to_s.gsub('#_=_', '') == tbi.get_smooch_authorization_token
+      q_params = {
+        client_id: CheckConfig.get('smooch_facebook_app_id'),
+        client_secret: CheckConfig.get('smooch_facebook_app_secret'),
+        access_token: auth&.dig('token'),
```
caiosba, line 78: Since in this case `auth` is not blank, please remove the `&`.

caiosba (changes requested):

caiosba (approved):
Thanks Sawy for addressing comments from the previous PR review!

PR #1847, opened 2024-04-03
Shared feed import

caiosba:
## Description

Mutation to import medias from shared feeds.

Reference: CV2-4109.

## How has this been tested?

Automated tests implemented.

## Checklist


app/models/cluster.rb
```diff
@@ -16,4 +16,41 @@ def items
...
+      parent = self.import_media_to_team(team, from_project_media, claim_title, claim_context)
+    else
+      parent = ProjectMedia.where(id: parent_id, team_id: team.id).last
+    end
+    self.class.delay_for(1.second).import_other_medias_to_team(self.id, parent.id)
+    parent
+  end
+
+  def self.import_other_medias_to_team(cluster_id, parent_id)
+    cluster = Cluster.find(cluster_id)
```
melsawy, line 47: I suggest using `find_by_id` to avoid raising errors in case the item not exists for any reason.

app/models/cluster.rb
```diff
@@ -16,4 +16,41 @@ def items
...
+    parent = ProjectMedia.find(parent_id)
```
melsawy, line 48: Same as above comment suggest using `find_by_id`

app/models/cluster.rb
```diff
@@ -16,4 +16,41 @@ def items
...
+    team = parent.team
+    cluster.project_medias.where.not(team_id: team.id).limit(50).find_each do |pm|
```
melsawy, line 50: Why `50` hard coded is this a pre-defined value? if yes I prefer to read a value for config.

app/models/cluster.rb
```diff
@@ -16,4 +16,41 @@ def items
...
+      next if ProjectMedia.where(team_id: team.id, media_id: pm.media_id).exists?
+      target = cluster.import_media_to_team(team, pm)
+      Relationship.create!(source: parent, target: target, relationship_type: Relationship.confirmed_type)
+    end
```
melsawy, line 54: I suggest using `find_in_batches` to minimize number of queries
i.e .
```
    cluster.project_medias.where.not(team_id: team.id).limit(50).find_in_batches(:batch_size => 50) do |pms|
      ProjectMedia.where(media_id: pms.map(&:media_id)).where.not(team_id: team.id).find_each do |pm|
        target = cluster.import_media_to_team(team, pm)
        Relationship.create!(source: parent, target: target, relationship_type: Relationship.confirmed_type)
      end
    end
```

melsawy (changes requested):
@caiosba I left few comments most of them to minimize number of queries.

app/models/cluster.rb
caiosba: Done!

app/models/cluster.rb
caiosba: Done!

app/models/cluster.rb
caiosba: But I actually changed to 100.

app/models/cluster.rb
caiosba: Good idea! Thank you.

app/models/cluster.rb
```diff
@@ -16,4 +16,41 @@ def items
...
+      set_claim_description: claim_title,
+      set_claim_context: claim_context
+    })
+  end
+
+  def import_medias_to_team(team, claim_title, claim_context, parent_id = nil)
+    parent = nil
+    if parent_id.nil?
+      # Find the first item in this cluster for which the media_id doesn't exist in the target team yet
+      from_project_media = self.project_medias.where.not(team_id: team.id).find { |item| !ProjectMedia.where(team_id: team.id, media_id: item.media_id).exists? }
```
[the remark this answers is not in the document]
caiosba: Thanks, good idea, done. I also added `.select(:media_id)` for performance reasons, since this is the only field I need.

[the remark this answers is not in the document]
caiosba: Hum, actually, we need to be sure that the `ProjectMedia` is part of the cluster. I pushed another way of doing it, please review.

app/models/cluster.rb
caiosba: Hum, actually, we need to be sure that the `ProjectMedia` is part of the cluster. I pushed another way of doing it, please review.

caiosba:
@melsawy , please re-review. If you have a better idea for the queries, let me know - but we can't iterate from `ProjectMedia`, we need to be sure they are part of the cluster - We could join with `cluster_project_medias` but then I think would actually be more complex than a few straight `SELECT`'s, since we have a maximum number for importing anyway.

melsawy (approved):

PR #1862, opened 2024-04-23
API Key Management Backend

jayjay-w:
## Description

Update the ApiKey model by adding new fields:
 - title
 - description
 - user_id
 - team_id

This change also adds GraphQL mutations to create and delete api keys in a given team.

References: CV2-4448

## How has this been tested?

1. Authenticate as a user who is an admin in a team
2. Run this mutation to test the creation of a new API Key:

```graphql
mutation create {
  createApiKey(input: {title: "test-api-key", description: "This is a test api key"}) {
    api_key {
      id
      title
      description
    }
  }
}
```
3. Repeat this severally to create several api keys in the team
4. Run the query below to list all the api keys in the team:

```graphql
query read {
  team {
    api_keys {
      edges {
        node {
          dbid
          title
          description
        }
      }
    }
  }
}
```

Confirm that the permissions are set as expected. Only team admins can manage api keys.

## Checklist


app/graph/mutations/api_key_mutations.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    def resolve(**inputs)
+      bot_user = BotUser.new
+      bot_user.name = inputs[:title]
+      bot_user.login = inputs[:title]
+      bot_user.save!
+
+      bot_user.api_key.title = inputs[:title]
+      bot_user.api_key.description = inputs[:description]
+      bot_user.api_key.save!
```
melsawy, line 25: Will be more clear to get `api_key` then do a save for api_key object.
i.e.
api_key = bot_user.api_key
api_key.title = inputs[:title]
api_key.description = inputs[:description]
 api_key.save!

app/graph/mutations/api_key_mutations.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  class Create < Mutations::CreateMutation
+    include CreateFields
+
+    def resolve(**inputs)
+      bot_user = BotUser.new
+      bot_user.name = inputs[:title]
+      bot_user.login = inputs[:title]
+      bot_user.save!
```
melsawy, line 21: you can use a `begin ... rescue ` to save object

app/models/api_key.rb
```diff
@@ -34,6 +38,11 @@ def generate_access_token
     end
   end
 
+  def set_user_and_team
+    self.user = User.current unless User.current.nil? || !self.class.column_names.include?('user_id')
+    self.team = Team.current unless Team.current.nil?|| !self.class.column_names.include?('team_id')
```
melsawy, line 43: Why do you need to add columns check?

db/migrate/20240424175612_add_team_id_to_api_keys.rb
```diff
@@ -0,0 +1,5 @@
+class AddTeamIdToApiKeys < ActiveRecord::Migration[6.1]
+  def change
+    add_column :api_keys, :team_id, :integer
```
melsawy, line 3: You can add this migration to the above one `AddTitleDescUserIdToApiKeys` as we still in same PR

melsawy (changes requested):

amoedoamorim:
@caiosba @jayjay-w I confirm the backend fulfills all current needs: Create, Read and Delete ApiKeys! :+1:

amoedoamorim:
Screenshot from 2024-05-02 11-35-42
I'm noticing API keys created locally are valid for 1 month only

app/graph/mutations/api_key_mutations.rb
```diff
@@ -0,0 +1,36 @@
...
+    extend ActiveSupport::Concern
+
+    included do
+      argument :title, GraphQL::Types::String, required: false
+      argument :description, GraphQL::Types::String, required: false
+    end
+  end
+
+  class Create < Mutations::CreateMutation
+    include CreateFields
```
caiosba, line 15: If this is the only place we're using `CreateFields`, it doesn't need to a module, right? We could just add the `argument` calls straight in this class body.

app/graph/mutations/api_key_mutations.rb
```diff
@@ -0,0 +1,36 @@
...
+
+        bot_user.api_key.title = inputs[:title]
+        bot_user.api_key.description = inputs[:description]
+        bot_user.api_key.save!
+
+        { success: true, api_key: bot_user.api_key }
+      rescue
+        { success: false, error: 'Could not create API key' }
+      end
+    end
```
caiosba, line 32: I'm curious, Jay: Why you had to use a custom `resolve` method? The `Mutations::CreateMutation` base class you're using should take care of everything. If this is because you need the `BotUser` instance to be created, I suggest you handle it at the model level as a `before_create` callback.

app/models/ability.rb
```diff
@@ -66,6 +66,7 @@ def admin_perms
     can :destroy, FeedTeam do |obj|
       obj.team_id == @context_team.id || obj.feed.team_id == @context_team.id
     end
+    can :cud, ApiKey, :team_id => @context_team.id
```
caiosba, line 69: Please add the `read` permission as well.

app/graph/types/team_type.rb
```diff
@@ -295,4 +295,14 @@ def tipline_messages(uid:)
...
+
+  field :api_key, ApiKeyType, null: true do
+    argument :dbid, GraphQL::Types::Int, required: true
+  end
+
+  def api_key(dbid:)
+    object.get_api_key(dbid)
+  end
+
+  field :api_keys, ApiKeyType.connection_type, null: true
```
caiosba, line 307: Since this is sensitive information, please implement a custom `resolve` for this connection, which checks if `ability.can?(:read, api_key)` for each API key returned (it will break pagination but it's OK - there is a limit on the number of API keys a workspace can create).

app/models/api_key.rb
caiosba: +1

app/graph/types/team_type.rb
```diff
@@ -295,4 +295,14 @@ def tipline_messages(uid:)
   def articles(article_type:)
     object.explainers if article_type == 'explainer'
   end
+
+  field :api_key, ApiKeyType, null: true do
+    argument :dbid, GraphQL::Types::Int, required: true
+  end
+
+  def api_key(dbid:)
+    object.get_api_key(dbid)
```
caiosba, line 304: Please check if the user has permission to read the API key before returning it... something like:
```
api_key = object.get_api_key(dbid)
ability.can?(:read, api_key) ? api_key : nil
```

db/migrate/20240420104318_add_title_desc_user_id_to_api_keys.rb
```diff
@@ -0,0 +1,8 @@
+class AddTitleDescUserIdToApiKeys < ActiveRecord::Migration[6.1]
+  def change
+    add_column :api_keys, :title, :string
+    add_column :api_keys, :description, :string
+    add_column :api_keys, :user_id, :integer
+    add_column :api_keys, :team_id, :integer
```
caiosba, line 6: I think it's better to add the new columns as `reference` other than `integer`... any reason for not doing so?

caiosba (changes requested):
Thanks Jay! I added a few comments related to permissions and existing code.

caiosba:
@jayjay-w , some other questions based on things defined in the spec and that I didn't see implemented here:

1. I think there is an additional step missing which is to add the `BotUser` to the team (by creating a `TeamUser`)... is it happening automatically? I didn't expect to. I think we need to explicitly create a team user between the `BotUser` and the `Team`, with role `editor`, when the API key is created.
2. It's still missing the validation on how many API keys a workspace can create, right? You can assume that the maximum can be defined in a configuration key set in SSM and retrieved by `CheckConfig`, and use something like 20 as the default.
3. We need to log requests made by an API key... I think it can be happen here: https://github.com/meedan/check-api/blob/develop/app/controllers/api/v1/graphql_controller.rb#L45

amoedoamorim:
@jayjay-w I generated the relay.json and pushed to your branch to allow testing the frontend with your branch

amoedoamorim (changes requested):
Hey @jayjay-w a few observations here. First one I shared with you on Slack already, but reposting here for a complete PR review:

1. The `team` returned in the `CreateApiKeyPayload` is always null. It should return the team in order to update the api_keys connection after creating. :white_check_mark: 
Screenshot from 2024-05-03 16-35-48

2. Creating a key with an existing name returns a null `api_key` in the payload, but no errors, as if the mutation was successful.
This should raise an error. :white_check_mark: 
Screenshot from 2024-05-03 16-36-08

3. Issue above happens even between different workspaces. API key names should be unique per workspace scope, there should be no problem having an API key with the same name in different workspaces.

amoedoamorim:
> 3. Issue above happens even between different workspaces. API key names should be unique per workspace scope, there should be no problem having an API key with the same name in different workspaces.

This is still happening. I'm not able to name keys the same in different workspaces

If I delete a key and attempt to reuse the name an error is thrown, is this intentional?

Screenshot from 2024-05-06 22-41-17

amoedoamorim:
Hey @jayjay-w the fixes look good but now the sorting seems a bit weird. 

image

Can you sort `team.api_keys` by created most recent first, please?

app/models/api_key.rb
```diff
@@ -34,7 +42,30 @@ def generate_access_token
     end
   end
 
+  def create_bot_user
+    if self.bot_user.blank? && self.team.present?
+      bot_name = "#{self.team.slug}-bot-#{self.title}"
+      new_bot_user = BotUser.new(api_key: self, name: bot_name, login: bot_name)
+      new_bot_user.skip_check_ability = true
+      new_bot_user.set_role 'editor'
```
caiosba, line 50: I was not sure if this would indeed create a `TeamUser` instance but I confirmed it does.

caiosba (approved):
Thanks @jayjay-w for addressing the review comments!

app/controllers/api/v1/graphql_controller.rb
```diff
@@ -64,6 +65,21 @@ def parse_graphql_result
...
+        user_name = ''
+        uid = 0
+        unless User.current.nil?
+          uid = User.current.id
+          user_name = User.current.name
+          team = Team.current || User.current.current_team
+          team = team.nil? ? '' : team.name
+          role = User.current.role
+        end
+        Rails.logger.info("[Graphql] Logging activity: uid: #{uid} user_name: #{user_name} team: #{team} role: #{role}")
```
melsawy, line 80: I think we should log activity if User.current is set yes?
if yes  it's better to move the logger inside unless block and remove the default values for team, role, user_name and uuid

melsawy (approved):
@jayjay-w I left a minor comment but code LGTM

amoedoamorim (approved):

PR #1906, opened 2024-05-30
CV2-4637 comment out functionality that would call soon-to-be deprecated endpoints

DGaffney:
## Description
We need to sunset a few pieces of functionality that rely on soon-to-be deprecated alegre endpoints. Specifically, on alegre, we're trying to remove dependencies on direct fingerprinting within alegre, or relying on the old alegre SharedModel workers - and as a result, the following commented-out functionalities will no longer work. We discussed and I believe it is acceptable to simply remove this functionality on the next deploy, but lets please make sure we're all on board / have had that discussion. Also, would it be better to simply remove the code rather than comment it out? I leave that decision to @caiosba 

References: CV2-4637

## How has this been tested?

Not yet tested - removing this is likely to cause some tests to error out so we will shakedown the changes during that process.

None

## Checklist

### DIFF
--- a/app/models/concerns/article.rb
+++ b/app/models/concerns/article.rb
@@ -13,7 +13,7 @@ module Article
     before_validation :set_user
     validates_presence_of :user
 
-    after_commit :update_elasticsearch_data, :send_to_alegre, :notify_bots, on: [:create, :update]
+    after_commit :update_elasticsearch_data, :notify_bots, on: [:create, :update] #:send_to_alegre, 
     after_commit :destroy_elasticsearch_data, on: :destroy
   end
 
@@ -38,7 +38,7 @@ module Article
   end
 
   def send_to_alegre
-    self.class.delay_for(1.second).send_to_alegre(self.id)
+    # self.class.delay_for(1.second).send_to_alegre(self.id)
   end
 
   def notify_bots
--- a/app/models/concerns/article.rb
+++ b/app/models/concerns/article.rb
@@ -13,7 +13,7 @@ module Article
     before_validation :set_user
     validates_presence_of :user
 
-    after_commit :update_elasticsearch_data, :notify_bots, on: [:create, :update] #:send_to_alegre, 
+    after_commit :update_elasticsearch_data, :send_to_alegre, :notify_bots, on: [:create, :update]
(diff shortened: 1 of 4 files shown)


app/models/concerns/article.rb
```diff
@@ -13,7 +13,7 @@ module Article
     before_validation :set_user
     validates_presence_of :user
 
-    after_commit :update_elasticsearch_data, :send_to_alegre, :notify_bots, on: [:create, :update]
+    after_commit :update_elasticsearch_data, :notify_bots, on: [:create, :update] #:send_to_alegre,
```
caiosba, line 16: I thought the issue was only with feed/request?

DGaffney: It is apparently not - how would you like to proceed now knowing this?

computermacgyver: This is a different `send_to_alegre`. It is simply a stub in `explainer.rb` at present. We do not need to comment and remove here.

app/models/feed.rb
```diff
@@ -181,7 +181,7 @@ def saved_search_was
   def self.save_request(feed_id, type, query, webhook_url, result_ids)
     media = Request.get_media_from_query(type, query, feed_id)
     request = Request.create!(feed_id: feed_id, request_type: type, content: query, webhook_url: webhook_url, media: media, skip_check_ability: true)
-    request.attach_to_similar_request!
+    # request.attach_to_similar_request!
```
computermacgyver, line 184: I don't want to remove `send_to_alegre` in its entirety. It still works fine for text. I feel we should only comment the one part of the function that tries to send media, but leave everything else as is.

computermacgyver (changes requested):

caiosba (commented):
Looks good to me, but, @computermacgyver can you double check as well please?

caiosba (commented):
Looks good to me, but, @computermacgyver can you double check as well please?

computermacgyver (approved):
Looks great. I'm agreed to removing this functionality for now. Per the call we've had today, we'll want to bring it back in asynchronous form at some point. Given that, I think we are best leaving the code commented and not removing it entirely. We could also comment the test rather than removing it for the same reason.

PR #1925, opened 2024-06-12
CV2-4741 increase timeout from one to two minutes

DGaffney:
## Description

Simple bump from one minute to two minutes for timeout errors in waiting on alegre responses.

References: CV2-4741

## How has this been tested?

Not tested yet - should be straightforward

NA

## Checklist

### DIFF
--- a/app/models/concerns/alegre_v2.rb
+++ b/app/models/concerns/alegre_v2.rb
@@ -432,7 +432,7 @@ module AlegreV2
     def wait_for_results(project_media, args)
       return {} if similarity_disabled_for_project_media?(project_media)
       cached_data = get_cached_data(get_required_keys(project_media, nil))
-      timeout = args[:timeout] || 120
+      timeout = args[:timeout] || CheckConfig.get('alegre_timeout', 120, :integer).to_f
       start_time = Time.now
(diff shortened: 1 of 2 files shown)


app/models/concerns/alegre_v2.rb
```diff
@@ -432,7 +432,7 @@ def is_cached_data_not_good(cached_data)
     def wait_for_results(project_media, args)
       return {} if similarity_disabled_for_project_media?(project_media)
       cached_data = get_cached_data(get_required_keys(project_media, nil))
-      timeout = args[:timeout] || 60
+      timeout = args[:timeout] || 120
```
caiosba, line 435: @DGaffney , since you're here, I suggest we turn it into a configuration, so we have more freedom to play with it without requiring code changes. But we can use 120 as the default value, so we don't need to set anything in SSM right now. My suggestion is to replace this line by:

```ruby
timeout = args[:timeout] || CheckConfig.get('alegre_timeout', 120, :integer)
```

And then add `alegre_timeout` to `config/config.yml.example` as well.

What do you think?

DGaffney: @caiosba totally reasonable! Will do that.

caiosba (approved):
:ship:

PR #1942, opened 2024-07-04
Try to make sure that clusters center is never null.

caiosba:
## Description

And/or that at least clusters without a center are never returned.

We don't know exactly in which cases this crash happens. So this commit approaches it in two different ways:

* For a given feed, only return clusters that have a center, by adding a `INNER JOIN project_medias` clause to the feed clusters query.
* For the `ClusterType` GraphQL endpoint, fallback to the cluster model `center` method, which already falls back to the first item of the cluster if the center is not present anymore.

Fixes CV2-4850.

## How has this been tested?

Existing tests should cover this.

I'm wondering if for some reason empty clusters are being created. If this is the case, we should fix the rake task as well.

## Checklist

### DIFF
--- a/app/graph/types/cluster_type.rb
+++ b/app/graph/types/cluster_type.rb
@@ -18,12 +18,6 @@ class ClusterType < DefaultObject
 
   field :center, ProjectMediaType, null: true
 
-  def center
-    RecordLoader
-      .for(ProjectMedia)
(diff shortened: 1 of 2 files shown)


melsawy (approved):
This is a good approach Dr Caio to call a center method that fallback to first item

PR #1976, opened 2024-07-31
Some fixes and improvements for the articles feature

caiosba:
## Description

- [ ] Don't include blank media in search queries (lists and counts)
- [ ] Update item title when fact-check is updated
- [ ] Filter out explainers already applied to the item

Reference: CV2-5000.

## How has this been tested?

Automated tests implemented.

## Checklist

### DIFF
--- a/app/models/concerns/team_associations.rb
+++ b/app/models/concerns/team_associations.rb
@@ -93,6 +93,7 @@ module TeamAssociations
     conditions[key] = obj.id
     relationship_type = Team.sanitize_sql(Relationship.confirmed_type.to_yaml)
     ProjectMedia.where(conditions)
+    .joins(:media).where('medias.type != ?', 'Blank')
     .joins("LEFT JOIN relationships r ON r.target_id = project_medias.id AND r.relationship_type = '#{relationship_type}'")
     .where('r.id IS NULL').count
   end
--- a/lib/check_search.rb
+++ b/lib/check_search.rb
@@ -46,7 +46,7 @@ class CheckSearch
     @file = file
   end
 
-  MEDIA_TYPES = %w[claims links twitter youtube tiktok instagram facebook telegram weblink images videos audios blank]
+  MEDIA_TYPES = %w[claims links twitter youtube tiktok instagram facebook telegram weblink images videos audios]
   SORT_MAPPING = {
     'recent_activity' => 'updated_at', 'recent_added' => 'created_at', 'demand' => 'demand',
     'related' => 'linked_items_count', 'last_seen' => 'last_seen', 'share_count' => 'share_count',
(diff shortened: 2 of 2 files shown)


app/models/concerns/project_media_getters.rb
```diff
@@ -153,7 +153,7 @@ def get_title
       return self.send(title_mapping[title_field]).to_s
     end
     title = self.original_title
-    [self.analysis['file_title'], self.analysis['title'], self.fact_check_title, self.claim_description_content].each do |value|
+    [self.analysis['file_title'], self.analysis['title'], self.fact_check_title(true), self.claim_description_content].each do |value|
```
melsawy, line 156: I think for `self.fact_check_title(true)`  we should not call cached field with true to enforce regenerate cached field and make sure that cached field updated in relevant actions

app/models/concerns/project_media_getters.rb
```diff
@@ -174,7 +174,7 @@ def media_slug
   end
 
   def get_description
-    return self.fact_check_summary if self.get_main_channel == CheckChannels::ChannelCodes::FETCH
+    return self.fact_check_summary(true) if self.get_main_channel == CheckChannels::ChannelCodes::FETCH
```
melsawy, line 177: I think for `self.fact_check_summary(true)` we should not call cached field with true to enforce regenerate cached field and make sure that cached field updated in relevant actions

app/models/team.rb
```diff
@@ -497,6 +497,9 @@ def filtered_explainers(filters = {})
     # Filter by text
     query = query.where('(title ILIKE ? OR url ILIKE ? OR description ILIKE ?)', *["%#{filters[:text]}%"]*3) if filters[:text].to_s.size > 2
 
+    # Exclude the ones already applied to a target item
+    query = query.where.not(id: ProjectMedia.find(filters[:target_id].to_i).explainer_ids) unless ProjectMedia.find_by_id(filters[:target_id].to_i).nil?
```
melsawy, line 501: I do not know if the ProjectMedia with `filters[:target_id]` exist this means that the query to find the ProjectMedia item will run twice one to check that item exists and the second one for the original query
if yes I think it's better to get the object then do our check
i.e 
pm = ProjectMedia.find_by_id(filters[:target_id].to_i)
query = query.where.not(id: pm.explainer_ids) unless pm.nil?

app/models/team.rb
```diff
@@ -533,6 +536,9 @@ def filtered_fact_checks(filters = {})
     # Filter by text
     query = query.where('(fact_checks.title ILIKE ? OR fact_checks.url ILIKE ? OR fact_checks.summary ILIKE ?)', *["%#{filters[:text]}%"]*3) if filters[:text].to_s.size > 2
 
+    # Exclude the ones already applied to a target item
+    query = query.where.not('fact_checks.id' => ProjectMedia.find(filters[:target_id].to_i).fact_check_id) unless ProjectMedia.find_by_id(filters[:target_id].to_i).nil?
```
melsawy, line 540: if the query run twice in case the item exists then I suggest to do the following:
pm = ProjectMedia.find_by_id(filters[:target_id].to_i)
query = query.where.not('fact_checks.id' => pm.fact_check_id) unless pm.nil?

app/models/concerns/project_media_getters.rb
caiosba: There was a race condition so I had to force that refresh... the event was triggering a call to this method, so it would fail.

app/models/concerns/project_media_getters.rb
caiosba: There was a race condition so I had to force that refresh... the event was triggering a call to this method, so it would fail.

app/models/team.rb
caiosba: Good idea, done!

app/models/team.rb
caiosba: Good idea, done!

melsawy (approved):

PR #1979, opened 2024-08-01
CV2-5006: fix Sentry issue

melsawy:
## Description

Fix sentry issue.

References: CV2-5006

## How has this been tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can verify the changes. Please describe whether or not you implemented automated tests.

- Etc.

## Checklist

### DIFF
--- a/lib/check_statistics.rb
+++ b/lib/check_statistics.rb
@@ -95,7 +95,7 @@ module CheckStatistics
             all = 0
             user = 0
             business = 0
-            data.dig('conversation_analytics', 'data', 0, 'data_points')&.each do |data_point|
+            data.dig('conversation_analytics', 'data', 0, 'data_points').each do |data_point|
               count = data_point['conversation']
               all += count
               user += count if data_point['conversation_direction'] == 'USER_INITIATED' || data_point['conversation_category'] == 'SERVICE'
--- a/lib/check_statistics.rb
+++ b/lib/check_statistics.rb
@@ -95,7 +95,7 @@ module CheckStatistics
             all = 0
             user = 0
             business = 0
-            data['conversation_analytics']['data'][0]['data_points'].each do |data_point|
+            data.dig('conversation_analytics', 'data', 0, 'data_points')&.each do |data_point|
               count = data_point['conversation']
               all += count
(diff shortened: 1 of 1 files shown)


caiosba (changes requested):
@melsawy, I'll follow up in the ticket on why we can't just ignore the cases where the data is not present.

lib/check_statistics.rb
```diff
@@ -95,7 +95,7 @@ def number_of_whatsapp_conversations(team_id, start_date, end_date, type = 'all'
             all = 0
             user = 0
             business = 0
-            data['conversation_analytics']['data'][0]['data_points'].each do |data_point|
+            data.dig('conversation_analytics', 'data', 0, 'data_points')&.each do |data_point|
```
melsawy, line 98: @caiosba I suggest to sent a notification in case data_points is nil
i.e 
data_points = data.dig('conversation_analytics', 'data', 0, 'data_points')
send_notification if data_points.nil?

caiosba: OK! Let's add details to the ticket.

melsawy: @caiosba I see that the message that I should add will be close to the Sentry error so I keep raising error for nil data be removing `&.each do |data_point|` and iterate data without `&.`

lib/check_statistics.rb
```diff
@@ -95,7 +95,7 @@ def number_of_whatsapp_conversations(team_id, start_date, end_date, type = 'all'
             all = 0
             user = 0
             business = 0
-            data['conversation_analytics']['data'][0]['data_points'].each do |data_point|
+            data.dig('conversation_analytics', 'data', 0, 'data_points').each do |data_point|
```
caiosba, line 98: @melsawy, if we make this change, what is going to trigger this line? -> https://github.com/meedan/check-api/blob/develop/lib/check_statistics.rb#L115. My suggestion would be to raise an exception if that array is empty but the workspace has `TiplineMessage`'s for the month.

caiosba: Yes, better to keep it as is, then.

PR #1982, opened 2024-08-02
5010 – Fix tag with trailing space error

vasconsaurus:
## Description

When trying to create an imported fact-check that had a tag with a trailing space it would fail (`text has already been taken`).

We have a normalization step but that runs `before_validation` on `TagText`.
So when we are checking if the tag exists in line 92 from `tag.rb`, the tag has not been normalized. This means:

- We look for 'tag', it is not there. We create it.
- We look for ' tag', it is not there. We try to create it, it gets normalized, so it becomes 'tag'. Then we fail to create it,
because it already exists.

Adding the strip when comparing tags fixes this, but I'm not sure if it is the best solution.

References: 5010

## How has this been tested?

- I added one big test that replicated the issue we saw: `rails test test/controllers/graphql_controller_12_test.rb:613`
- And added one model test: `rails test test/models/tag_test.rb:275`

- Etc.

## Checklist

### DIFF
--- a/app/models/annotations/tag.rb
+++ b/app/models/annotations/tag.rb
@@ -89,7 +89,7 @@ class Tag < ApplicationRecord
   def get_tag_text_reference
     if self.tag.is_a?(String)
       team_id = self.team&.id
-      tag_text = TagText.where(text: (self.tag).strip, team_id: team_id).last
+      tag_text = TagText.where(text: self.tag.strip, team_id: team_id).last
       if tag_text.nil? && team_id.present?
         tag_text = TagText.new
         tag_text.text = self.tag
--- a/app/models/annotations/tag.rb
+++ b/app/models/annotations/tag.rb
@@ -89,7 +89,7 @@ class Tag < ApplicationRecord
   def get_tag_text_reference
     if self.tag.is_a?(String)
       team_id = self.team&.id
-      tag_text = TagText.where(text: self.tag, team_id: team_id).last
+      tag_text = TagText.where(text: (self.tag).strip, team_id: team_id).last
       if tag_text.nil? && team_id.present?
         tag_text = TagText.new
         tag_text.text = self.tag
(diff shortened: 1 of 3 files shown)


vasconsaurus:
@caiosba Could you please take a look?
- This fixes it, but I'm not sure if it's a good solution.
- And although the first test helped me get this to work, I'm not sure we should keep it. WDYT?

app/models/annotations/tag.rb
```diff
@@ -89,7 +89,7 @@ def self.run_bulk_create_callbacks(ids_json, pmids_json)
   def get_tag_text_reference
     if self.tag.is_a?(String)
       team_id = self.team&.id
-      tag_text = TagText.where(text: self.tag, team_id: team_id).last
+      tag_text = TagText.where(text: (self.tag).strip, team_id: team_id).last
```
caiosba, line 92: Minor, only FYI: The parenthesis are not needed here.

caiosba:
That's a good solution, Manu. And thanks for adding the tests. Were you able to reproduce the same exception in the test, before the fix?

app/models/annotations/tag.rb
```diff
@@ -89,7 +89,7 @@ def self.run_bulk_create_callbacks(ids_json, pmids_json)
   def get_tag_text_reference
     if self.tag.is_a?(String)
       team_id = self.team&.id
-      tag_text = TagText.where(text: self.tag, team_id: team_id).last
+      tag_text = TagText.where(text: self.tag.strip, team_id: team_id).last
```
melsawy, line 92: Also I suggest to trim database value so we can catch existing one with trailing space
i.e
tag_text = TagText.where(team_id: team_id).where('trim(text) = ?', self.tag.strip).last

melsawy: Oh I see that we normalize tag before saving in DB and did `self.text.strip.gsub(/^#/, '')` so we can skip this comment

melsawy (approved):

caiosba (approved):

PR #1994, opened 2024-08-17
Cv2 5082 article indexing to presto

DGaffney:
## Description

Moving other side-effecting article storage to vectorize in presto

References: CV2-5082

## How has this been tested?

Not yet tested - changing code first to determine effect of changes then work through necessary changes from there - testing will need to just ensure that articles do in fact get indexed if you send these commands to alegre from check-api

None in particular!

## Checklist

### DIFF
--- a/app/models/concerns/alegre_v2.rb
+++ b/app/models/concerns/alegre_v2.rb
@@ -36,6 +36,10 @@ class TemporaryProjectMedia
   def is_audio?
     self.type == "audio"
   end
+
+  def is_uploaded_media?
+    self.is_image? || self.is_audio? || self.is_video?
+  end
 end
 
 module AlegreV2
@@ -170,7 +174,7 @@ module AlegreV2
         return content_hash_for_value(project_media.media.url)
       elsif project_media.is_a?(TemporaryProjectMedia)
         return Rails.cache.read("url_sha:#{project_media.url}")
-      elsif !project_media.is_text?
+      elsif project_media.is_uploaded_media?
         return project_media.media.file.filename.split(".").first
       else
         return content_hash_for_value(project_media.send(field).to_s)
@@ -313,6 +317,10 @@ module AlegreV2
         delete_package(project_media, field, params),
         project_media
       )
+    rescue StandardError => e
+      error = Error.new(e)
+      Rails.logger.error("[AutoTagger Bot] Exception for event `#{body['event']}`: #{error.class} - #{error.message}")
+      CheckSentry.notify(error, bot: "alegre", project_media: project_media, params: params, field: field)
     end
 
     def get_per_model_threshold(project_media, threshold)
--- a/app/models/concerns/alegre_v2.rb
+++ b/app/models/concerns/alegre_v2.rb
@@ -1,7 +1,7 @@
 require 'active_support/concern'
 class AlegreTimeoutError < StandardError; end
 class TemporaryProjectMedia
-  attr_accessor :team_id, :id, :url, :text, :type
+  attr_accessor :team_id, :id, :url, :text, :type, :field
   def media
     media_type_map = {
       "claim" => "Claim",
@@ -284,8 +284,8 @@ module AlegreV2
       generic_package_text(project_media, field, params)
     end
 
-    def index_async_with_params(params, type, suppress_response=true)
-      request("post", async_path_for_type(type), params.merge(suppress_response: suppress_response))
+    def index_async_with_params(params, type, suppress_search_response=true)
+      request("post", async_path_for_type(type), params.merge(suppress_search_response: suppress_search_response))
     end
 
     def get_sync_with_params(params, type)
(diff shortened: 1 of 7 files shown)


app/models/concerns/article.rb
```diff
@@ -84,7 +84,7 @@ def send_to_alegre(id)
       obj = self.find_by_id(id)
       return if obj.project_media.nil?
       obj.text_fields.each do |field|
-        ::Bot::Alegre.send_field_to_similarity_index(obj.project_media, field)
+        ::Bot::Alegre.relate_project_media_async(obj.project_media, field)
```
caiosba, line 87: This affects fact-checks actually, so we don't need to change it to the scope of this ticket. Let's focus only on Explainers first.

app/models/explainer.rb
```diff
@@ -106,7 +108,7 @@ def self.search_by_similarity(text, language, team_id)
         language: language
       }
     }
-    response = Bot::Alegre.request('post', '/text/similarity/search/', params)
+    response = Bot::Alegre.get_sync_raw_params(params, "text")
```
caiosba, line 111: I see this line is the same as line 85 above... but line 85 is sending content to be indexed in Alegre, while this line is searching for similar content from Alegre... so why they are the same?

app/models/explainer.rb
```diff
@@ -63,24 +63,26 @@ def self.update_paragraphs_in_alegre(id, previous_paragraphs_count, timestamp)
...
       params = {
+        content_hash: Bot::Alegre.content_hash_for_value(paragraph.strip),
         doc_id: Digest::MD5.hexdigest(['explainer', explainer.id, 'paragraph', count].join(':')),
+        context: base_context.merge({ paragraph: count }),
         text: paragraph.strip,
         models: ALEGRE_MODELS_AND_THRESHOLDS.keys,
-        context: base_context.merge({ paragraph: count })
       }
-      Bot::Alegre.request('post', '/text/similarity/', params)
+      Bot::Alegre.get_sync_raw_params(params, "text")
```
caiosba, line 85: Suggestion: This method actually asks Alegre to index information, right? I suggest a more clear name for it (even if it's just an abstract method in `Alegre v2` that calls this method).

app/models/explainer.rb
DGaffney: We don't have separate functionality for sending items to index vs getting search results - we do it all in one step either with async or sync. We can.... create new functionality to break that up?

app/models/explainer.rb
DGaffney: Will do

app/models/concerns/alegre_v2.rb
```diff
@@ -148,28 +159,34 @@ def get_type(project_media)
...
 
+    def content_hash_for_value(value)
+      value.nil? ? nil : Digest::MD5.hexdigest(value)
+    end
+
     def content_hash(project_media, field)
       if Bot::Alegre::ALL_TEXT_SIMILARITY_FIELDS.include?(field)
-        Digest::MD5.hexdigest(project_media.send(field))
+        content_hash_for_value(project_media.send(field))
       else
```
melsawy, line 169: I think you can continue with `else if`  to handle all cases

app/models/concerns/alegre_v2.rb
```diff
@@ -148,28 +159,34 @@ def get_type(project_media)
       type
     end
 
+    def content_hash_for_value(value)
+      value.nil? ? nil : Digest::MD5.hexdigest(value)
+    end
+
     def content_hash(project_media, field)
```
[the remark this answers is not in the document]
melsawy: I see there is no arg for `case` so I think `if .... else if` will be more clear

melsawy (approved):
LGTM just left a single comment related to `else if`

app/models/explainer.rb
```diff
@@ -106,7 +108,7 @@ def self.search_by_similarity(text, language, team_id)
         language: language
       }
     }
-    response = Bot::Alegre.request('post', '/text/similarity/search/', params)
+    response = Bot::Alegre.index_async_with_params(params, "text")
```
caiosba, line 111: This should be `get_async_with_params`, no?

DGaffney: thats right - will fix

app/models/concerns/alegre_v2.rb
```diff
@@ -148,28 +159,32 @@ def get_type(project_media)
...
       if Bot::Alegre::ALL_TEXT_SIMILARITY_FIELDS.include?(field)
-        Digest::MD5.hexdigest(project_media.send(field))
+        content_hash_for_value(project_media.send(field))
+      elsif project_media.is_link?
+        return content_hash_for_value(project_media.media.url)
+      elsif project_media.is_a?(TemporaryProjectMedia)
+        return Rails.cache.read("url_sha:#{project_media.url}")
+      elsif !project_media.is_text?
+        return project_media.media.file.filename.split(".").first
```
caiosba, line 174: Suggestion: It's stronger to change this validation to something like `is_uploaded_media?`, which basically checks if it's audio, image or video. I'm saying this because we also have `blank` and may have other types.

app/models/concerns/alegre_v2.rb
```diff
@@ -485,25 +512,27 @@ def wait_for_results(project_media, args)
...
+        project_media.field = field
+        project_media.url = media_url
+        project_media.id = Digest::MD5.hexdigest(project_media.url).to_i(16)
+        project_media.team_id = team_ids
+        project_media.type = type
       end
+      get_similar_items_v2_async(project_media, nil, threshold)
+      wait_for_results(project_media, args)
+      response = get_similar_items_v2_callback(project_media, nil)
+      delete(project_media, nil) if project_media.is_a?(TemporaryProjectMedia)
```
caiosba, line 534: As we were talking last week, let's be sure that if `delete` fails, we either retry or at least notify Sentry.

app/models/explainer.rb
caiosba, line 1: Thanks, I think the method names are more clear now!

caiosba (approved):
Thanks, Devin. I added a couple of suggestions.

jayjay-w (approved):

PR #1996, opened 2024-08-18
CV2-5086 Move Smooch NLU to presto

DGaffney:
## Description

More work to move some of the lower level Alegre uses over to presto.

References: CV2-5086

## How has this been tested?

Not yet tested - changing code first to determine effect of changes then work through necessary changes from there - testing will need to just ensure that articles do in fact get indexed if you send these commands to alegre from check-api

Nothing in particular 

## Checklist


DGaffney:
@caiosba not sure why this last error is happening so may want to discuss but otherwise ready for review! I'm going to hold in draft until later though so we can not have a flood of these at the same time.

app/lib/smooch_nlu.rb
```diff
@@ -46,13 +46,13 @@ def update_keywords(language, keywords, keyword, operation, doc_id, context)
       keywords << keyword
       alegre_operation = 'post'
       alegre_params = common_alegre_params.merge({ text: keyword, models: ALEGRE_MODELS_AND_THRESHOLDS.keys })
+      Bot::Alegre.get_sync_raw_params(alegre_params, "text") if alegre_operation && alegre_params
     elsif operation == 'remove'
       keywords -= [keyword]
       alegre_operation = 'delete'
       alegre_params = common_alegre_params.merge({ quiet: true })
+      Bot::Alegre.request_delete_from_raw(alegre_params, "text") if alegre_operation && alegre_params
```
caiosba, line 54: This can be simplified further, it seems... you don't need `alegre_operation` anymore, since the calls were moved to the condition blocks... right?

caiosba (changes requested):

app/lib/smooch_nlu.rb
DGaffney: you're right! Updated.

PR #2005, opened 2024-08-27
5120 – Create Tags in the background

vasconsaurus:
### Context
While working on  `API/Zapier Error: The app did not respond in-time. It may or may not have completed successfully` we thought tags creation could be causing the slow response.

We couldn’t pinpoint it as the cause, but it’s still a good performance improvement and should help improve the response time. So we are moving forward with it.

### Implementation Notes

Sidekiq’s `delayed extensions` were deprecated in version 5 and removed in 7. So we are taking this opportunity to create a `GenericJob`, and eventually substitute the instances of `.delay`/`.delay_for` we have.

References: 5120, 4959

## How has this been tested?

`rails test test/models/tag_test.rb:288`
`rails test test/workers/generic_worker_test.rb`

- Etc.

## Checklist

### DIFF
--- a/app/models/concerns/project_media_creators.rb
+++ b/app/models/concerns/project_media_creators.rb
@@ -264,7 +264,9 @@ module ProjectMediaCreators
 
   def create_tags_in_background
     if self.set_tags.is_a?(Array)
-      ProjectMedia.run_later_in(1.second, 'create_tags', project_media_id: self.id, tags_json: self.set_tags.to_json, user_id: self.user_id)
+      project_media_id = self.id
+      tags_json = self.set_tags.to_json
+      ProjectMedia.run_later_in(1.second, 'create_tags', project_media_id, tags_json, user_id: self.user_id)
     end
   end
 end
--- a/app/models/concerns/project_media_creators.rb
+++ b/app/models/concerns/project_media_creators.rb
@@ -264,7 +264,7 @@ module ProjectMediaCreators
 
   def create_tags_in_background
     if self.set_tags.is_a?(Array)
-        GenericWorker.perform_in(1.second, 'ProjectMedia', 'create_tags', project_media_id: self.id, tags_json: self.set_tags.to_json, user_id: self.user_id)
+      ProjectMedia.run_later_in(1.second, 'create_tags', project_media_id: self.id, tags_json: self.set_tags.to_json, user_id: self.user_id)
     end
   end
 end
--- a/app/models/concerns/project_media_creators.rb
+++ b/app/models/concerns/project_media_creators.rb
@@ -263,6 +263,7 @@ module ProjectMediaCreators
   end
 
   def create_tags
-    self.set_tags.each { |tag| Tag.create!(annotated: self, tag: tag.strip, skip_check_ability: true) } if self.set_tags.is_a?(Array)
+    # self.set_tags.each { |tag| Tag.create!(annotated: self, tag: tag.strip, skip_check_ability: true) } if self.set_tags.is_a?(Array)
+    self.set_tags.each { |tag| GenericWorker.perform_in(1.second, 'tag', 'create!', annotated: self, tag: tag.strip, skip_check_ability: true) } if self.set_tags.is_a?(Array)
   end
 end
--- a/app/models/concerns/project_media_creators.rb
+++ b/app/models/concerns/project_media_creators.rb
@@ -263,9 +263,6 @@ module ProjectMediaCreators
   end
 
   def create_tags
-    # self.set_tags.each { |tag| Tag.create!(annotated: self, tag: tag.strip, skip_check_ability: true) } if self.set_tags.is_a?(Array)
-    # require 'byebug'
-    # byebug
-    self.set_tags.each { |tag| GenericWorker.perform_in(1.second, 'Tag', 'create!', annotated: self, tag: tag.strip, skip_check_ability: true) } if self.set_tags.is_a?(Array)
+    self.set_tags.each { |tag| GenericWorker.perform_in(1.second, 'Tag', 'create!', annotated_type: 'ProjectMedia' , annotated_id: self.id, tag: tag.strip, skip_check_ability: true) } if self.set_tags.is_a?(Array)
   end
 end
--- a/app/models/concerns/project_media_creators.rb
+++ b/app/models/concerns/project_media_creators.rb
@@ -262,9 +262,9 @@ module ProjectMediaCreators
     fc
   end
 
-  def create_tags
+  def create_tags_in_background
     if self.set_tags.is_a?(Array)
-        GenericWorker.perform_in(1.second, 'ProjectMedia', 'create_tags_in_background', project_media_id: self.id, tags_json: self.set_tags.to_json, user_id: self.user_id)
+        GenericWorker.perform_in(1.second, 'ProjectMedia', 'create_tags', project_media_id: self.id, tags_json: self.set_tags.to_json, user_id: self.user_id)
     end
   end
 end
--- a/app/models/concerns/project_media_creators.rb
+++ b/app/models/concerns/project_media_creators.rb
@@ -263,7 +263,9 @@ module ProjectMediaCreators
   end
 
   def create_tags
-    # self.set_tags.each { |tag| Tag.create!(annotated: self, tag: tag.strip, skip_check_ability: true) } if self.set_tags.is_a?(Array)
-    self.set_tags.each { |tag| GenericWorker.perform_in(1.second, 'tag', 'create!', annotated: self, tag: tag.strip, skip_check_ability: true) } if self.set_tags.is_a?(Array)
+    self.set_tags.each { |tag| Tag.create!(annotated: self, tag: tag.strip, skip_check_ability: true) } if self.set_tags.is_a?(Array)
+    # require 'byebug'
+    # byebug
+    # self.set_tags.each { |tag| GenericWorker.perform_in(1.second, 'Tag', 'create!', annotated: self, tag: tag.strip, skip_check_ability: true) } if self.set_tags.is_a?(Array)
   end
 end
--- a/app/models/concerns/project_media_creators.rb
+++ b/app/models/concerns/project_media_creators.rb
@@ -264,7 +264,7 @@ module ProjectMediaCreators
 
   def create_tags
     if self.set_tags.is_a?(Array)
-        GenericWorker.perform_in(1.second, 'ProjectMedia', 'create_tags_in_background', project_media_id: self.id, tags_json: self.set_tags.to_json)
+        GenericWorker.perform_in(1.second, 'ProjectMedia', 'create_tags_in_background', project_media_id: self.id, tags_json: self.set_tags.to_json, user_id: self.user_id)
     end
   end
 end
--- a/app/models/concerns/project_media_creators.rb
+++ b/app/models/concerns/project_media_creators.rb
@@ -263,9 +263,9 @@ module ProjectMediaCreators
   end
 
   def create_tags
-    self.set_tags.each { |tag| Tag.create!(annotated: self, tag: tag.strip, skip_check_ability: true) } if self.set_tags.is_a?(Array)
+    # self.set_tags.each { |tag| Tag.create!(annotated: self, tag: tag.strip, skip_check_ability: true) } if self.set_tags.is_a?(Array)
     # require 'byebug'
     # byebug
-    # self.set_tags.each { |tag| GenericWorker.perform_in(1.second, 'Tag', 'create!', annotated: self, tag: tag.strip, skip_check_ability: true) } if self.set_tags.is_a?(Array)
+    self.set_tags.each { |tag| GenericWorker.perform_in(1.second, 'Tag', 'create!', annotated: self, tag: tag.strip, skip_check_ability: true) } if self.set_tags.is_a?(Array)
   end
 end
--- a/app/models/concerns/project_media_creators.rb
+++ b/app/models/concerns/project_media_creators.rb
@@ -263,6 +263,8 @@ module ProjectMediaCreators
   end
 
   def create_tags
-    self.set_tags.each { |tag| GenericWorker.perform_in(1.second, 'Tag', 'create!', annotated_type: 'ProjectMedia' , annotated_id: self.id, tag: tag.strip, skip_check_ability: true) } if self.set_tags.is_a?(Array)
+    if self.set_tags.is_a?(Array)
+        GenericWorker.perform_in(1.second, 'ProjectMedia', 'create_tags_in_background', project_media_id: self.id, tags_json: self.set_tags.to_json)
+    end
   end
 end
--- a/app/models/project_media.rb
+++ b/app/models/project_media.rb
@@ -531,6 +531,7 @@ class ProjectMedia < ApplicationRecord
   def self.create_tags_in_background(**params)
     params = params.with_indifferent_access
     project_media = ProjectMedia.find_by_id(params['project_media_id'])
+
     unless project_media.nil?
       tags = JSON.parse(params['tags_json'])
       tags.each { |tag| Tag.create! annotated: project_media, tag: tag.strip, skip_check_ability: true }
--- a/app/models/project_media.rb
+++ b/app/models/project_media.rb
@@ -530,9 +530,11 @@ class ProjectMedia < ApplicationRecord
 
 
   def self.create_tags_in_background(**params)
-    project_media = ProjectMedia.find(params['project_media_id'])
-    tags = JSON.parse(params['tags_json'])
-    tags.each { |tag| Tag.create! annotated: project_media, tag: tag.strip, skip_check_ability: true }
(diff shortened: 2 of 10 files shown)


vasconsaurus:
@caiosba I tried to summarize what I've been thinking and some of the issues I went through.
Could you take a look and see if this makes any sense at all?

### On where I think this work could go
This is a work in progress. I thought it would be nice if we could maybe have a wrapper or something, so the usage of this worker would be similar to `delay`. So instead of `GenericWorker.perform_in...` we could call something like `Tag.run_in..`.

### Tests
I wrote two tests to try to guide this work, I think they make sense?
```
rails test test/models/tag_test.rb:288
rails test test/workers/generic_worker_test.rb
```

### On Tag creation:
#### Contexts
- We have different contexts: Creating tags when creating a project media, creating tags by themselves.
- Since we are talking about this from the Zapier/Check Integration context, that means creating a tag when we create a project media.
#### How it works
- If a project media has a `set_tags` array, it runs an `after_create` callback `create_tags`: I thought it made sense to have that callback happen in the background.
#### Annotations
- This part of the codebase touches annotations.
- I had an issue because originally `create_tags` sent the entire project_media object as annotator. Here: https://github.com/meedan/check-api/blob/f1f3b6cb7acee60daf7d2ef0b1fd357253a6a0cf/app/models/concerns/project_media_creators.rb#L266
- But I looked through our graphql queries wiki and saw that just sending the `annotator_type` and `annotator_id` should be enough to create a Tag. Here: https://github.com/meedan/check/wiki/GraphQL-query-examples#add-a-tag-to-an-item
- I updated `create_tags` to send only those two values, and it seems to work.
- So I didn't have to touch annotations, but I do wonder if we are getting rid of them, if this change here makes sense. Or if we want to go another way about this.

app/models/concerns/project_media_creators.rb
```diff
@@ -263,6 +263,6 @@ def create_claim_description_and_fact_check
   end
 
   def create_tags
-    self.set_tags.each { |tag| Tag.create!(annotated: self, tag: tag.strip, skip_check_ability: true) } if self.set_tags.is_a?(Array)
+    self.set_tags.each { |tag| GenericWorker.perform_in(1.second, 'Tag', 'create!', annotated_type: 'ProjectMedia' , annotated_id: self.id, tag: tag.strip, skip_check_ability: true) } if self.set_tags.is_a?(Array)
```
caiosba, line 266: Suggestion: I think it's going to be easier to encapsulate and to unit-test if we wrap the whole execution in a single worker run. Something like:

```ruby
def create_tags
  if self.set_tags.is_a?(Array)
    GenericWorker.perform_in(1.second, 'ProjectMedia', 'create_tags_in_background', self.id, self.set_tags.to_json)
  end
end

def self.create_tags_in_background(project_media_id, tags_array_as_json)
  project_media = ProjectMedia.find(project_media_id)
  tags = JSON.parse(tags_array_as_json)
  tags.each { |tag| Tag.create! annotated: project_media, tag: tag.strip, skip_check_ability: true) }
end
```

Rationale:

- For clarity, I think it's important for the signature of `GenericWorker` to be explicitly class/method/arguments (I see it's implicit now) - more on that in the next comment
- Each tag in a separate background job run means background job overhead for a small task, and it can also lead to inconsistencies

app/workers/generic_worker.rb
```diff
@@ -0,0 +1,14 @@
+class GenericWorker
+
+  include Sidekiq::Worker
+
+  def perform(klass_name, *args)
+    klass = klass_name.constantize
+    options = args.extract_options!
```
caiosba, line 7: Suggestion for clarity: The signature could be explicitly `klass_name, klass_method, method_args`.

caiosba (changes requested):
Thanks @vasconsaurus , the direction is correct, I left a suggestion for better encapsulation and more clear code. Please let me know if it makes sense to you.

caiosba:
@jayjay-w @melsawy , as we talked, please support/follow @vasconsaurus here too.

melsawy:
> @jayjay-w @melsawy , as we talked, please support/follow @vasconsaurus here too.

Sure Caio

vasconsaurus:
@melsawy @jayjay-w could you please review? (it is still a work in progress)
Please take into consideration @caiosba requests, and I added some more context in this [comment](https://github.com/meedan/check-api/pull/2005#issuecomment-2312620318)

app/models/project_media.rb
```diff
@@ -485,6 +485,13 @@ def add_nested_objects(ms)
     ms.attributes[:requests] = requests
   end
 
+
+  def self.create_tags_in_background(**params)
+    project_media = ProjectMedia.find(params['project_media_id'])
```
melsawy, line 490: You should use `find_by_id` to avoid raising error in case item not found
i.e 
project_media = ProjectMedia.find_by_id params['project_media_id']
unless project_media.nil?
 .....

app/workers/generic_worker.rb
```diff
@@ -0,0 +1,14 @@
+class GenericWorker
+
+  include Sidekiq::Worker
+
+  def perform(klass_name, klass_method, *method_args)
+    klass = klass_name.constantize
```
melsawy, line 6: We should also pass User.current to the worker so we can set User.current  before calling the background method
User.current  = passed_variable
......set the logic...
User.current  = nil

test/models/project_media_8_test.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  test "when creating an item with tag/tags, tags should be created in the background" do
+    Sidekiq::Testing.inline!
+
+    t = create_team
+    p = create_project team: t
+    assert_nothing_raised do
+      create_project_media project: p, tags: ['one']
+    end
```
melsawy, line 16: This test should be something like this to test background jobs
t = create_team
Sidekiq::Testing.fake! do
 pm = create_project_media team: t, tags: ['one']
assert_difference 'Tag.count' do
 Sidekiq::Worker.drain_all
end
 assert_equal 1, pm.annoations('tag').count
end

melsawy (commented):
Left comments related to User.current and tests

app/workers/generic_worker.rb
vasconsaurus: @melsawy, a question: Why do we need `User.current` here?

app/models/project_media.rb
vasconsaurus: @melsawy do we want to send a log or error to Sentry (if the tags are not created because the `project_media` is `nil`)?

melsawy: It's a good suggestion, agreed

app/workers/generic_worker.rb
melsawy: We log the revisions for tag creation in case User.current exists otherwise no logs for this action

test/models/project_media_8_test.rb
vasconsaurus: @melsawy a question for when you are back:
Is there a difference between using `Sidekiq::Testing.inline!` vs `Sidekiq::Testing.fake!` \  `Sidekiq::Worker.drain_all`?

I updated it like this:
https://github.com/meedan/check-api/blob/353612139eca53830ed062ac8612cc03b36a9a11/test/models/project_media_8_test.rb#L27-35

app/workers/generic_worker.rb
vasconsaurus: @melsawy another question for when you are back:
The `GenericWorker` is as the name says, so, do we always want to set `User.current`? Does it make sense to always set `User.current`?

I wasn't sure about that, so I wrote it like this for now, but I'm not too sure if it makes sense
https://github.com/meedan/check-api/blob/353612139eca53830ed062ac8612cc03b36a9a11/app/workers/generic_worker.rb#L5-L18

app/models/project_media.rb
vasconsaurus: Added the `unless.project_media.nil?` and the `CheckSentry.notify`, question: should I keep it like this or would it be better to use `begin/rescue`?
https://github.com/meedan/check-api/blob/353612139eca53830ed062ac8612cc03b36a9a11/app/models/project_media.rb#L488-L498

test/models/project_media_8_test.rb
melsawy: when we need to make sure that job created in background we should use `Sidekiq::Testing.fake!` and check the jobs size otherwise the both are same.

app/workers/generic_worker.rb
melsawy: Yes make sense to always set `User.current` and I suggest the following code for User.current

   if options
      user_id = options.delete(:user_id) if options.key?(:user_id)
      current_user = User.current
      User.current = User.find_by_id(user_id)
      klass.public_send(klass_method, **options)
      User.current = current_user
    else
      klass.public_send(klass_method)
    end

vasconsaurus:
@melsawy @caiosba

I have made some updates based on Sawy's review, and started to try to write a wrapper method.
Currently it works, and I added a couple of new tests, but it doesn't work as a drop-in replacement.

I think we could try to implement something like this
```
ProjectMedia.run_later_in(time).create_tags(project_media_id: self.id, tags_json: self.set_tags.to_json, user_id: self.user_id)
```

But, even then, we might not or want a drop in replacement? We could pass an object to `delay_for`, but that is, I think, one of the reasons they removed it. So we might not want to replicate that behavior, and update our code accordingly, what to you think? 

Besides that, I have other questions:
- What do you think about this so far? 
- Are there any things you think are missing or anything to change?
- How should I move forward?
- On errors: My tests pass and the tags get greated but I'm seeing at least a couple errors on Sidekiq:

Screenshot 2024-09-13 at 16 15 32

Let me know if you have any questions 🙂

app/models/project_media.rb
```diff
@@ -528,6 +528,19 @@ def add_nested_objects(ms)
     ms.attributes[:requests] = requests
   end
 
+  def self.create_tags(**params)
```
caiosba, line 531: In order to make the code more clear and more predictable, it's better to explicitly list all arguments that the method expects, instead of `**params` (`**params` is more common in meta-programming).

test/models/project_media_8_test.rb
```diff
@@ -0,0 +1,72 @@
+require_relative '../test_helper'
+
+class ProjectMedia8Test < ActiveSupport::TestCase
+  def setup
+    super
```
caiosba, line 5: Tests will run faster if you don't run everything that `super` runs. I suggest you start with empty `setup` and `teardown` methods and then add only the steps you actually need:

```
def setup
end

def teardown
end
```

test/models/project_media_8_test.rb
```diff
@@ -0,0 +1,72 @@
...
+      ProjectMedia.create_tags(project_media_id: nil, tags_json: ['one', 'two'].to_json)
+    end
+  end
+
+  test "when creating an item with tag/tags, after_create :create_tags_in_background callback should create tags in the background" do
+    Sidekiq::Testing.inline!
+
+    team = create_team
+    project = create_project team: team
+    pm = create_project_media project: project, tags: ['one']
```
caiosba, line 32: The setter `tags` works? I thought it was `set_tags`.

test/models/project_media_8_test.rb
```diff
@@ -0,0 +1,72 @@
...
+  end
+
+  test "when creating an item with multiple tags, after_create :create_tags_in_background callback should only schedule one job" do
+    Sidekiq::Testing.fake!
+
+    team = create_team
+    project = create_project team: team
+
+    assert_nothing_raised do
+      create_project_media project: project, tags: ['one', 'two', 'three']
```
caiosba, line 44: The setter `tags` works? I thought it was `set_tags`.

test/workers/generic_worker_test.rb
```diff
@@ -0,0 +1,28 @@
+require 'test_helper'
+
+class GenericWorkerTest < ActiveSupport::TestCase
+  def setup
+    super
```
caiosba, line 5: Tests will run faster if you don't run everything that `super` runs. I suggest you start with empty `setup` and `teardown` methods and then add only the steps you actually need:

```
def setup
end

def teardown
end
```

test/workers/generic_worker_test.rb
```diff
@@ -0,0 +1,28 @@
...
+    end
+  end
+
+  test "should schedule a job for Team creation" do
+    Sidekiq::Testing.inline!
+
+    assert_nothing_raised do
+      GenericWorker.perform_async('Team', 'create!', name: 'BackgroundTeam', slug: 'background-team')
+    end
+  end
```
caiosba, line 27: I suggest adding a test here for the `Sidekiq::Testing.fake!` case too.

test/workers/generic_worker_test.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    assert_nothing_raised do
+      GenericWorker.perform_async('Tag', 'create!', annotated_type: 'ProjectMedia' , annotated_id: pm.id, tag: 'test_tag', skip_check_ability: true, user_id: pm.user_id)
+    end
+  end
+
+  test "should schedule a job for Team creation" do
+    Sidekiq::Testing.inline!
+
+    assert_nothing_raised do
```
caiosba, line 24: Another good assertion to do here is:

```
assert_difference "Team.where(name: 'BackgroundTeam', slug: 'background-team').count" do
  ...
end
```

This way you also make it's doing what it's supposed to.

caiosba (changes requested):

caiosba:
Great stuff @vasconsaurus ! The error you're seeing in Sidekiq usually means that code was changed but the application was not restarted after that. Just remember to do `touch tmp/restart.txt` every time you chance code and so you should never see this error in Sidekiq.

I agree we should not do a drop-in replacement. Let's have our own implementation with our own signature. It will be easier to track during the migration too.

I added other comments directly to the code - nothing that changes the direction, and I also don't see anything missing. Once comments are addressed, all tests are passing and all the code is covered, I think this is ready for final review and tests. Great work!

test/models/project_media_8_test.rb
vasconsaurus: @caiosba this `create_project_media` is a method from `sample_data`, here is the relevant line:

https://github.com/meedan/check-api/blob/dffb66d1040398aecf78e43cec8be0840605a04e/lib/sample_data.rb#L520

caiosba (changes requested):
Thanks @vasconsaurus for addressing all the comments from my last review! It's looking great! Just one last ask here: please add a test for this case: https://github.com/meedan/check-api/blob/2c9c1d0e8c8e93d86657a7cc3b060059990f2a8e/app/workers/generic_worker.rb#L15.

app/workers/generic_worker.rb
```diff
@@ -0,0 +1,18 @@
...
+    options = method_args.extract_options!.with_indifferent_access
+    if options
```
caiosba, line 8: @vasconsaurus the tests you added in your last commit are still not covering line 15 - I found out that the problem is here... `options` is always a hash, so `if options` is always `true`... you should replace this conditional by something like `unless options.blank?`.

caiosba (changes requested):

app/workers/generic_worker.rb
vasconsaurus: @caiosba I updated this, but have a question regarding tests:
The two tests I added were passing before, so they are not really working as expected. How could I fix it?

Also, is there any extra logging or error handling that I should add?

caiosba: Thanks, Manu. No, I don't think any additional logging or error handling is needed here. But for the test, I suggest you actually assess that the method is executed. For now your test for the no-arguments case just verifies that no exception is raised.

caiosba:
@melsawy , since you requested changes before, please re-review this PR.

app/workers/generic_worker.rb
vasconsaurus: true 🤦‍♀️ still, do you think I should verify, assert anything else?

melsawy (commented):
LGTM, great work @vasconsaurus 
You can merge once apply Caio's requests and check test coverage.

app/workers/generic_worker.rb
vasconsaurus: I updated to use `assert_difference` both here and  for the test in line 28. 
But one last comment: even when the test wasn't hitting the `else`, because I was using `if options`, the test still succeeded. So blank was still created.

So I guess what I mean is, this test makes sure `Blank` is created, but it doesn't make sure of which branch we are hitting.

caiosba: Correct, I think that the other block was still able to handle it just fine. That said, maybe the if/else is not even needed, but it's more readable/clear to have them separate.

vasconsaurus: Yup, so I was wondering about that. I think it might be nice to have the option with just the one straightforward step, but it isn't really needed. Should we keep it or remove it (the conditional)?

caiosba: Yeah, that's what I said, I think it's more clear to have the conditional. Ruby Meta programming can be harder to read, otherwise :)

melsawy (approved):

PR #2012, opened 2024-08-28
Search for fact-checks by multiple keywords regardless the order.

caiosba:
## Description

Search for fact-checks by multiple keywords regardless the order.

Fixes: CV2-5129.

## How has this been tested?

Unit test added.

May not work for all languages.

## Checklist

### DIFF
--- a/app/models/team.rb
+++ b/app/models/team.rb
@@ -496,7 +496,7 @@ class Team < ApplicationRecord
 
     # Filter by text
     if filters[:text].to_s.size > 2
-      tsquery = Team.sanitize_sql_array(["websearch_to_tsquery(?)", filters[:text].split(/\s+/).map(&:strip).join(' & ')]) # FIXME: May not work for all languages
+      tsquery = Team.sanitize_sql_array(["websearch_to_tsquery(?)", filters[:text]]) # FIXME: May not work for all languages
       tsvector = "to_tsvector('simple', coalesce(title, '') || ' ' || coalesce(url, '') || coalesce(description, ''))"
       query = query.where(Arel.sql("#{tsvector} @@ #{tsquery}"))
     end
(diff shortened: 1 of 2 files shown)


app/models/team.rb
```diff
@@ -535,7 +535,11 @@ def filtered_fact_checks(filters = {})
     query = query.where('fact_checks.report_status' => filters[:report_status].to_a.map(&:to_s)) unless filters[:report_status].blank?
 
     # Filter by text
-    query = query.where('(fact_checks.title ILIKE ? OR fact_checks.url ILIKE ? OR fact_checks.summary ILIKE ?)', *["%#{filters[:text]}%"]*3) if filters[:text].to_s.size > 2
+    if filters[:text].to_s.size > 2
+      tsquery = Team.sanitize_sql_array(["to_tsquery(?)", filters[:text].split(/\s+/).map(&:strip).join(' & ')]) # FIXME: May not work for all languages
```
computermacgyver, line 539: Not sure if `plainto_tsquery` is any better for other languages? It does seem simpler, however as we don't have to split the query: https://www.postgresql.org/docs/current/textsearch-controls.html

computermacgyver: Or maybe even better is `websearch_to_tsquery`:
> websearch_to_tsquery creates a tsquery value from querytext using an alternative syntax in which simple unformatted text is a valid query. Unlike plainto_tsquery and phraseto_tsquery, it also recognizes certain operators. Moreover, this function will never raise syntax errors, which makes it possible to use raw user-supplied input for search. The following syntax is supported:

melsawy: yes @computermacgyver `websearch_to_tsquery`  is better I used it and verified that everything works as expected

caiosba: @melsawy , since we're using `websearch_to_tsquery`, do we still need to `split` the query? It's better if we don't need to.

app/models/team.rb
```diff
@@ -495,7 +495,11 @@ def filtered_explainers(filters = {})
     query = query.where(updated_at: Range.new(*format_times_search_range_filter(JSON.parse(filters[:updated_at]), nil))) unless filters[:updated_at].blank?
 
     # Filter by text
-    query = query.where('(title ILIKE ? OR url ILIKE ? OR description ILIKE ?)', *["%#{filters[:text]}%"]*3) if filters[:text].to_s.size > 2
+    if filters[:text].to_s.size > 2
+      tsquery = Team.sanitize_sql_array(["websearch_to_tsquery(?)", filters[:text].split(/\s+/).map(&:strip).join(' & ')]) # FIXME: May not work for all languages
```
computermacgyver, line 499: My reading of the `websearch_to_tsquery` documentation is that we can simply use:
`tsquery = Team.sanitize_sql_array(["websearch_to_tsquery(?)", filters[:text]])`
and do not need to split, strip, and join. I've tried the unit test with this alternative and it passes. I'm building the full app to try it end-to-end.

computermacgyver (approved):

computermacgyver (approved):

PR #2039, opened 2024-09-16
Sort tags alphabetically

amoedoamorim:
## Description

Sorting team.tag_texts and project_media.tags alphabetically

References: CV2-2502

## How has this been tested?

Verified sorted output is correct on check-web

- Etc.

## Checklist

### DIFF
--- a/app/graph/types/project_media_type.rb
+++ b/app/graph/types/project_media_type.rb
@@ -206,7 +206,7 @@ class ProjectMediaType < DefaultObject
   field :tags, TagType.connection_type, null: true
 
   def tags
-    object.get_annotations('tag').map(&:load)
+    object.get_annotations('tag').map(&:load).sort_by { |tag| tag.tag_text.downcase }
   end
 
   field :comments, CommentType.connection_type, null: true
--- a/app/graph/types/team_type.rb
+++ b/app/graph/types/team_type.rb
@@ -224,7 +224,7 @@ class TeamType < DefaultObject
   end
 
   def tag_texts(keyword: nil)
-    object.tag_texts_by_keyword(keyword).sort_by { |tag_text| tag_text.text.downcase }
+    object.tag_texts_by_keyword(keyword).order('text ASC')
   end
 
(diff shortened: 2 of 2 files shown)


app/graph/types/project_media_type.rb
```diff
@@ -206,7 +206,7 @@ def flags
   field :tags, TagType.connection_type, null: true
 
   def tags
-    object.get_annotations('tag').map(&:load)
+    object.get_annotations('tag').map(&:load).sort_by { |tag| tag.tag_text.downcase }
```
caiosba, line 209: This is OK for this case because this method already returns an array (see that `.map()` is used)... so `sort_by`, which sorts an array in memory, won't create a performance issue or break pagination.

app/graph/types/team_type.rb
```diff
@@ -224,7 +224,7 @@ def sources(keyword: nil)
   end
 
   def tag_texts(keyword: nil)
-    object.tag_texts_by_keyword(keyword)
+    object.tag_texts_by_keyword(keyword).sort_by { |tag_text| tag_text.text.downcase }
```
caiosba, line 227: In this case, we want an `ActiveRecord::Collection` to be returned, this way, the GraphQL pagination can be applied. If we use `sort_by`, it's going to convert the query into an array, which is going to load all tags in memory, creating a performance issue and breaking pagination. For this case, the right way is: `.order('text ASC')`. See that the underlying method returns an `ActiveRecord::Collection`: https://github.com/meedan/check-api/blob/develop/app/models/team.rb#L430. So, we can just append other `ActiveRecord` methods to the chain (order, limit, where, etc.).

caiosba (changes requested):

app/graph/types/team_type.rb
amoedoamorim: Oh, that makes sense, thanks! Silly me and copilot! :D

caiosba (approved):

PR #2054, opened 2024-09-26
5120 – Dont create duplicate tags and clean up `#`

vasconsaurus:
### Context
While looking into a tags issue I noticed a few things:
- when we made a request with duplicate tags:
    - we got an error, so the job was retried 
    - the tag was added twice to the FactCheck
    - the tag is added once to the ProjectMedia
- when we made a request with a tag with a `#`
   - we got an error, so the job was retried
       - there seem to have been two errors related to this:
           - ActiveRecord::RecordInvalid: Tag already exists
           - ActiveRecord::RecordInvalid: Text has already been taken 
   - the tag with the `#` is added to the FactCheck 
   - the tag is not added to the ProjectMedia

### What was happening
- There are a few things happening at the same time:
    - Creation of a `ProjectMedia` with tags 
    - Creation of a `FactCheck` with tags
    - Tags are an `Object` from the `Tag` Class for `ProjectMedia`, but are a simple array for `FactCheck`
- For the ProjectMedia:
    - It would create the tag, then it would try to create the same tag again, and then it would fail and retry again, and so on
    - This happen because we have validations in place for the `Tag` class
- For `FactCheck`:
   - It would just create the tags twice
   - Because we have no validations for the tags array

### How it was fixed
To make sure we don't have duplicate tags:
- I added a method to do a clean up step, to make sure the `FactCheck` tags are unique, and
- added uniq to project media tag creation in `Tag` 

To make sure we deal `#tags`
- I updated the clean up step in `FactCheck`, and
- added a clean up step to project media tag creation in `Tag`

### Notes
While running tests there were some cases where the creation of the `ProjectMedia`/`FactCheck` completely failed due to failed Tag's validations. Maybe if we fail to create the tags, we still want the `ProjectMedia`/`FactCheck` to be created (just without the tags). If that is the case, maybe we should look into it in a separate ticket.

References: 5120

## How has this been tested?

rails test test/workers/generic_worker_test.rb
rails test test/models/project_media_8_test.rb
rails test test/lib/generic_worker_helpers_test.rb
rails test test/models/tag_test.rb
rails test test/controllers/graphql_controller_12_test.rb:684

- Etc.

## Checklist

### DIFF
--- a/app/models/annotations/tag.rb
+++ b/app/models/annotations/tag.rb
@@ -96,7 +96,7 @@ class Tag < ApplicationRecord
 
     if !project_media.nil?
       tags = JSON.parse(tags_json)
-      clean_tags = tags.map { |tag| tag.strip.gsub(/^#/, '') }.uniq!
+      clean_tags = tags.map { |tag| tag.strip.gsub(/^#/, '') }.uniq
       clean_tags.each { |tag| Tag.create! annotated: project_media, tag: tag.strip, skip_check_ability: true }
     else
       error = StandardError.new("[ProjectMedia] Exception creating project media's tags in background. Project media is nil.")
--- a/app/models/annotations/tag.rb
+++ b/app/models/annotations/tag.rb
@@ -1,5 +1,6 @@
 class Tag < ApplicationRecord
   include AnnotationBase
+  extend TagHelpers
 
   # "tag" is a reference to a TagText object
   field :tag, GraphQL::Types::Int, presence: true
@@ -96,8 +97,7 @@ class Tag < ApplicationRecord
 
     if !project_media.nil?
       tags = JSON.parse(tags_json)
-      clean_tags = tags.map { |tag| tag.strip.gsub(/^#/, '') }.uniq
-      clean_tags.each { |tag| Tag.create! annotated: project_media, tag: tag.strip, skip_check_ability: true }
+      clean_tags(tags).each { |tag| Tag.create! annotated: project_media, tag: tag.strip, skip_check_ability: true }
     else
       error = StandardError.new("[ProjectMedia] Exception creating project media's tags in background. Project media is nil.")
       CheckSentry.notify(error, project_media_id: project_media_id)
--- a/app/models/annotations/tag.rb
+++ b/app/models/annotations/tag.rb
@@ -96,7 +96,7 @@ class Tag < ApplicationRecord
 
     if !project_media.nil?
       tags = JSON.parse(tags_json)
-      tags.each { |tag| Tag.create! annotated: project_media, tag: tag.strip, skip_check_ability: true }
+      tags.uniq.each { |tag| Tag.create! annotated: project_media, tag: tag.strip, skip_check_ability: true }
     else
       error = StandardError.new("[ProjectMedia] Exception creating project media's tags in background. Project media is nil.")
       CheckSentry.notify(error, project_media_id: project_media_id)
--- a/app/models/annotations/tag.rb
+++ b/app/models/annotations/tag.rb
@@ -96,7 +96,8 @@ class Tag < ApplicationRecord
 
     if !project_media.nil?
       tags = JSON.parse(tags_json)
-      tags.uniq.each { |tag| Tag.create! annotated: project_media, tag: tag.strip, skip_check_ability: true }
+      clean_tags = tags.map { |tag| tag.strip.gsub(/^#/, '') }.uniq!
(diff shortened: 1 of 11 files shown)


app/models/fact_check.rb
```diff
@@ -56,6 +57,11 @@ def self.get_exported_data(query, team)
     data
   end
 
+  def clean_tags
+    return if self.tags.blank?
+    self.tags = self.tags.map! { |tag| tag.strip.gsub(/^#/, '') }.uniq
```
caiosba, line 62: Not a blocker, but: `map!` already changes `self.tags` in place, so the `self.tags = self.tags.map!` is redundant (that's true for any method that ends on `!`)... but since at the end you use `uniq`, which doesn't change the attribute in place, I suggest you do:

```ruby
self.tags = self.tags.collect{ |tag| tag.strip.gsub(/^#/, '') }.uniq
```

caiosba (approved):
Thanks @vasconsaurus , it captures everything we discussed. I left one suggestion, though... just for clarity and to avoid redundancy, but very minor. Great work on the tests!

app/models/annotations/tag.rb
```diff
@@ -91,6 +91,19 @@ def hit_nested_objects_limit?
     ret
   end
 
+  def self.create_project_media_tags(project_media_id, tags_json)
+    project_media = ProjectMedia.find_by_id(project_media_id)
+
+    if !project_media.nil?
+      tags = JSON.parse(tags_json)
+      clean_tags = tags.map { |tag| tag.strip.gsub(/^#/, '') }.uniq
```
melsawy, line 99: I think we should also handle case sensitive to avoid duplication.
i.e  clean_tags = tags.map { |tag| tag.strip.gsub(/^#/, '') }.map(&:downcase).uniq

app/models/fact_check.rb
melsawy, line 62: I think we should also handle case sensitive to avoid duplication.
i.e self.tags = self.tags.map { |tag| tag.strip.gsub(/^#/, '') }.map(&:downcase).uniq

test/controllers/graphql_controller_12_test.rb
```diff
@@ -680,4 +680,55 @@ def teardown
...
+    create_team_user team: t, user: b
+    p = create_project team: t
+    authenticate_with_token(a)
+
+    query1 = ' mutation create {
+              createProjectMedia(input: {
+                project_id: ' + p.id.to_s + ',
+                media_type: "Blank",
+                channel: { main: 1 },
+                set_tags: ["science", "science", "#science"],
```
melsawy, line 698: also set_tags: ["science", "science", "#science", 'SCIENCE', '#SCIENCE'] should pass

test/models/fact_check_test.rb
```diff
@@ -596,4 +596,26 @@ def setup
       end
     end
   end
+
+  test "should not have duplicate tags" do
+    pm = create_project_media
+    cd = create_claim_description(project_media: pm)
+    fc = create_fact_check claim_description: cd, tags: ['one', 'one', '#one']
```
melsawy, line 603: also fc = create_fact_check claim_description: cd, tags: ['one', 'one', '#one', 'ONE', '#ONE'] should pass with same tags count

melsawy (changes requested):
I added a comment related to case sensitive  as I think creating tags should be case insensitive

vasconsaurus:
@caiosba @melsawy I have a couple questions:
- should we extract `tags.map { |tag| tag.strip.downcase.gsub(/^#/, '') }.uniq` to a concern? So we don't have to update it in two places?
- as a step after this work, is there any clean up work we should do?
    - If we are downcasing the tags: should we be able to create standalone tags that are capitalized?
    - If we are removing the `#` should we clean up fact-check tags that have the `#`?
    - Should we make sure the blank project media has the same tags as its related fact-check?

Screenshot 2024-09-30 at 10 48 38

caiosba:
I don't think we should downcase before saving, because it may change the intent of the user (for example, it can be an acronym, person name, brand, etc.). Again, I see it as an example of us interfering in an editorial decision, which we shouldn't do. So, I take @melsawy's point, but if we don't need to do it, I don't think we should.

melsawy (approved):

vasconsaurus:
@caiosba @melsawy I extracted the logic to a concern, can you see if it makes sense?

caiosba (approved):
Great refactoring, thanks Manu!

PR #2079, opened 2024-10-11
[WIP] 5441 – Change how we create imported fact-checks when an item already exists

vasconsaurus:
- **update how we create project media in sample data**
- **add tests**

## Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context – why has this been changed/fixed.

References: TICKET-ID, TICKET-ID, …

## How has this been tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can verify the changes. Please describe whether or not you implemented automated tests.

- Etc.

## Checklist

### DIFF
--- a/test/controllers/graphql_controller_12_test.rb
+++ b/test/controllers/graphql_controller_12_test.rb
@@ -731,4 +731,119 @@ class GraphqlController12Test < ActionController::TestCase
     assert_equal 1, pm_tags.count
     assert_equal 1, fc_tags.count
   end
+
+  test "should append FactCheck to ProjectMedia, if ProjectMedia already exists and does not have a FactCheck" do
+    Sidekiq::Testing.inline!
+    url = 'http://example.com'
+    t = create_team
+    p = create_project team: t
+    pm = create_project_media team: t, url: url
+
+    assert_not_nil pm
+
+    a = ApiKey.create!
+    b = create_bot_user api_key_id: a.id
+    create_team_user team: t, user: b
+    authenticate_with_token(a)
+
+    pender_url = CheckConfig.get('pender_url_private') + '/api/medias'
+    response_body = '{"type":"media","data":{"url":"' + url + '","type":"item"}}'
+    WebMock.stub_request(:get, pender_url).with({ query: { url: url } }).to_return(body: response_body)
+
+    query = <<~GRAPHQL
+      mutation {
+        createProjectMedia(input: {
+          project_id: #{p.id},
+          media_type: "Blank",
+          channel: { main: 1 },
+          set_tags: ["tag"],
+          set_status: "verified",
+          set_claim_description: "Lorem ipsum dolor sit amet, consectetur adipiscing elit.",
+          set_original_claim: "#{url}",
+          set_fact_check: {
+            title: "Title #1",
+            language: "en",
+          }
+        }) {
+          project_media {
+          dbid
+          full_url
+            claim_description {
+              fact_check {
+                dbid
+              }
+            }
+          }
+        }
+      }
+    GRAPHQL
+
+    post :create, params: { query: query, team: t.slug }
+    assert_response :success
+
+    response_pm = JSON.parse(@response.body)['data']['createProjectMedia']['project_media']
+    fact_check = response_pm['claim_description']['fact_check']
+
+    assert_not_nil fact_check
+    assert_equal response_pm['dbid'], pm.id
+  end
+
+  test "should create a FactCheck with a Blank ProjectMedia, if ProjectMedia already exists and has a FactCheck in a different language" do
+    Sidekiq::Testing.inline!
+    url = 'http://example.com'
+
+    t = create_team
+    p = create_project team: t
+    pm = create_project_media team: t, url: url
+    c = create_claim_description project_media: pm
+    fc_1 = create_fact_check claim_description: c
+
+    assert_not_nil fc_1
+
+    a = ApiKey.create!
+    b = create_bot_user api_key_id: a.id
+    create_team_user team: t, user: b
+    authenticate_with_token(a)
+
+    pender_url = CheckConfig.get('pender_url_private') + '/api/medias'
+    response_body = '{"type":"media","data":{"url":"' + url + '","type":"item"}}'
+    WebMock.stub_request(:get, pender_url).with({ query: { url: url } }).to_return(body: response_body)
+
+    query = <<~GRAPHQL
+      mutation {
+        createProjectMedia(input: {
+          project_id: #{p.id},
+          media_type: "Blank",
+          channel: { main: 1 },
+          set_tags: ["tag"],
+          set_status: "verified",
+          set_claim_description: "Lorem ipsum dolor sit amet, consectetur adipiscing elit.",
+          set_original_claim: "#{url}",
+          set_fact_check: {
+            title: "Title #1",
+            language: "pt",
+          }
+        }) {
+          project_media {
+          dbid
+          full_url
+            claim_description {
+              fact_check {
+                dbid
+              }
+            }
+          }
+        }
+      }
+    GRAPHQL
+
+    post :create, params: { query: query, team: t.slug }
+    assert_response :success
+
+    response_pm = JSON.parse(@response.body)['data']['createProjectMedia']['project_media']
+    fc_2 = response_pm['claim_description']['fact_check']
+
+    assert_not_nil fc_2
+    assert_not_equal response_pm['dbid'], pm.id
+  end
 end
--- a/lib/sample_data.rb
+++ b/lib/sample_data.rb
@@ -511,7 +511,7 @@ module SampleData
       options[:team] = options[:project].team
     end
     options[:team] = create_team unless options.has_key?(:team)
-    options[:media] = create_valid_media({team: options[:team]}) unless options.has_key?(:media)
+    options[:media] = create_valid_media({team: options[:team], url: options[:url]}) unless options.has_key?(:media)
     options.each do |key, value|
       pm.send("#{key}=", value) if pm.respond_to?("#{key}=")
     end
@@ -550,7 +550,7 @@ module SampleData
 
   def create_valid_media(options = {})
     pender_url = CheckConfig.get('pender_url_private') + '/api/medias'
-    url = random_url
+    url = options[:url] ? options[:url] : random_url
     params = { url: url }
     params[:archivers] = Team.current.enabled_archivers if !Team.current&.enabled_archivers.blank?
     WebMock.stub_request(:get, pender_url).with({ query: params }).to_return(body: '{"type":"media","data":{"url":"' + url + '","type":"item","archives":{}}}')


vasconsaurus:
@caiosba when you have some time, could you please take a look into the tests?

```
rails test test/controllers/graphql_controller_12_test.rb:735
rails test test/controllers/graphql_controller_12_test.rb:792
```

test/controllers/graphql_controller_12_test.rb
```diff
@@ -731,4 +731,119 @@ def teardown
...
+              fact_check {
+                dbid
+              }
+            }
+          }
+        }
+      }
+    GRAPHQL
+
+    post :create, params: { query: query, team: t.slug }
```
caiosba, line 781: I suggest we put a `assert_no_difference 'ProjectMedia.count'` and a `assert_difference 'FactCheck.count'` around this line.

test/controllers/graphql_controller_12_test.rb
```diff
@@ -731,4 +731,119 @@ def teardown
...
+
+    pender_url = CheckConfig.get('pender_url_private') + '/api/medias'
+    response_body = '{"type":"media","data":{"url":"' + url + '","type":"item"}}'
+    WebMock.stub_request(:get, pender_url).with({ query: { url: url } }).to_return(body: response_body)
+
+    query = <<~GRAPHQL
+      mutation {
+        createProjectMedia(input: {
+          project_id: #{p.id},
+          media_type: "Blank",
```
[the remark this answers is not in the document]
vasconsaurus: Hm, good question. When we make the request in Zapier we do send the `media_type`.

vasconsaurus:
@caiosba I'm going to start taking a look at the code today. Are there any specific steps you think I should take?

@jayjay-w Is there anything on the `original claim` side I should take into consideration? Do you have any suggestions on how to tackle this?

caiosba:
@vasconsaurus my suggestion to start is to `rescue` the specific exception _only_ for the `ProjectMedia` use case, and then call a method (defined in the model) that is going to handle the two special cases we outlined in the ticket... so, [here](https://github.com/meedan/check-api/blob/develop/app/graph/mutations/graphql_crud_operations.rb#L10), something like:

```ruby
begin
  obj.save_with_version!
rescue StandardError => e # I know it's not StandardError, let's rescue the exception we currently get... we can replace the current exception class by a specific one if it makes things easier
  if obj.is_a?(ProjectMedia) && !obj.set_fact_check.blank? && !obj.set_original_claim.blank?
    obj = obj.handle_fact_check_for_existing_claim # We can then define this method in the ProjectMedia model and implement the logic for the two cases we listed in the ticket - it will be much easier to unit-test too
  else
    raise e
  end
end
```

Let me know if it makes sense! :)

But a question before you proceed: Were you able to reproduce in the two automated tests the same error we get from Zapier? Would be good to document it (e.g., the test run and the output) in the ticket.

jayjay-w:
@vasconsaurus I do not have any special considerations that we need to take regarding `original_claim`, so Caio's rescue suggestion will be sufficient, in my opinion.

vasconsaurus:
@caiosba I think I'll be able to get back to this one today, and will look into:

>But a question before you proceed: Were you able to reproduce in the two automated tests the same error we get from Zapier? Would be good to document it (e.g., the test run and the output) in the ticket.

vasconsaurus:
> But a question before you proceed: Were you able to reproduce in the two automated tests the same error we get from Zapier? Would be good to document it (e.g., the test run and the output) in the ticket.

@caiosba it seems we do get the exact same error 🎉 

From Zapier logs:
```
{
  "errors": [
    {
      "message": "This item already exists",
      "code": 9,
      "data": { ... }
    }
  ]
}
```

From our tests
```
"errors":[{"message":"This item already exists", "code":9, "data":{...}}]
```

caiosba:
> > But a question before you proceed: Were you able to reproduce in the two automated tests the same error we get from Zapier? Would be good to document it (e.g., the test run and the output) in the ticket.
> 
> @caiosba it seems we do get the exact same error 🎉
> 
> From Zapier logs:
> 
> ```
> {
>   "errors": [
>     {
>       "message": "This item already exists",
>       "code": 9,
>       "data": { ... }
>     }
>   ]
> }
> ```
> 
> From our tests
> 
> ```
> "errors":[{"message":"This item already exists", "code":9, "data":{...}}]
> ```

That's perfect, Manu! :ok_hand:

vasconsaurus:
@caiosba @jayjay-w 
I realized I was only looking at the original claim as a link, and I noticed that it behaves differently when it's a link vs when it's another type. When it is a link if it is the same link the media creation fails. But if it is the same image file or quote it creates a new media. Is this the expected behavior?

I added new tests [here](https://github.com/meedan/check-api/pull/2079/commits/7dc85456a57122d1ce911afff6ab07b86b1e7d70)

And ran manual tests for quote and image
Screenshot 2024-10-22 at 15 40 41
Screenshot 2024-10-22 at 15 41 04

caiosba:
> @caiosba @jayjay-w I realized I was only looking at the original claim as a link, and I noticed that it behaves differently when it's a link vs when it's another type. When it is a link if it is the same link the media creation fails. But if it is the same image file or quote it creates a new media. Is this the expected behavior?
> 
> I added new tests [here](https://github.com/meedan/check-api/pull/2079/commits/7dc85456a57122d1ce911afff6ab07b86b1e7d70)
> 
> And ran manual tests for quote and image Screenshot 2024-10-22 at 15 40 41 Screenshot 2024-10-22 at 15 41 04

@vasconsaurus yes, that's expected. We have a filter for existing items for media submissions that happen through tiplines:

- Text: https://github.com/meedan/check-api/blob/develop/app/models/bot/smooch.rb#L834
- Link: https://github.com/meedan/check-api/blob/develop/app/models/bot/smooch.rb#L843
- Uploaded files: https://github.com/meedan/check-api/blob/develop/app/models/bot/smooch.rb#L920

I suggest you ignore it for now, and we can work on that in a separate, follow-up ticket.

app/graph/mutations/graphql_crud_operations.rb
```diff
@@ -92,7 +92,15 @@ def self.create(type, inputs, ctx, parents_mapping = {})
     attrs["annotation_type"] = type.gsub(/^dynamic_annotation_/, "") if type =~
       /^dynamic_annotation_/
 
-    self.safe_save(obj, attrs, parents_mapping.keys)
+    begin
+      self.safe_save(obj, attrs, parents_mapping.keys)
+    rescue StandardError => e
```
caiosba, line 97: Please remember to replace this by a more specific exception.

app/models/project_media.rb
```diff
@@ -430,6 +430,22 @@ def apply_rules_and_actions_on_update
     self.team.apply_rules_and_actions(self, rule_ids)
   end
 
+  def handle_fact_check_for_existing_claim
+    # find the project media with the original claim media
+    pm = ProjectMedia.find_by(media_id: self.media_id)
```
caiosba, line 435: Please filter by the `team_id` as well.

app/graph/mutations/graphql_crud_operations.rb
```diff
@@ -92,7 +92,15 @@ def self.create(type, inputs, ctx, parents_mapping = {})
...
+      if obj.is_a?(ProjectMedia) && obj.set_fact_check.present? && obj.set_original_claim.present?
+        obj = obj.handle_fact_check_for_existing_claim
```
caiosba, line 99: As per our conversation, let's turn this method into a class method:

```ruby
obj = ProjectMedia.handle_fact_check_for_existing_claim(obj)
```

app/models/project_media.rb
```diff
@@ -430,6 +430,22 @@ def apply_rules_and_actions_on_update
     self.team.apply_rules_and_actions(self, rule_ids)
   end
 
+  def handle_fact_check_for_existing_claim
```
caiosba, line 433: As per comment above, turn this into a class method.

app/models/project_media.rb
```diff
@@ -430,6 +430,22 @@ def apply_rules_and_actions_on_update
...
+    pm = ProjectMedia.find_by(media_id: self.media_id)
+    if pm.fact_check.blank?
+      pm.set_claim_description = self.set_claim_description
+      pm.set_fact_check = self.set_fact_check
+      pm.create_claim_description_and_fact_check
+    elsif pm.fact_check.present?
+      if pm.fact_check.language != self.set_fact_check['language']
+        m = Blank.create!
+        self.set_original_claim = nil
+        self.media_id = m.id
```
caiosba, line 444: Once this becomes a class method, we can create a new instance here:

```ruby
pm = obj.clone # If it doesn't work as expected, replace this by ProjectMedia.new
pm.set_original_claim = nil
pm.media_id = m.id
pm.save!
```

app/models/project_media.rb
```diff
@@ -430,6 +430,22 @@ def apply_rules_and_actions_on_update
...
+      end
+    end
```
caiosba, line 446: Then, at the end, we just need to return `pm`.

caiosba (changes requested):

vasconsaurus:
@caiosba I have questions! I think I might be missing something

1. Our first test ([here](https://github.com/meedan/check-api/blob/7dc85456a57122d1ce911afff6ab07b86b1e7d70/test/controllers/graphql_controller_12_test.rb#L735)) is currently passing. But I realized this [line](https://github.com/meedan/check-api/blob/7dc85456a57122d1ce911afff6ab07b86b1e7d70/app/models/project_media.rb#L439) returns the fact check and not the project media. If I return the project media I get an error. That seems to be the same I'm getting for the second test, so moving on to that...
2. The second test, I tried saving the project media but I get the error below:
```
Failure:
GraphqlController12Test#test_should_create_a_FactCheck_with_a_Blank_ProjectMedia,_if_ProjectMedia_already_exists_and_has_a_FactCheck_in_a_different_language [/app/test/controllers/graphql_controller_12_test.rb:847]:
Expected response to be a <2XX: success>, but was a <400: Bad Request>
Response body: {"errors":[{"message":"          Failed to implement CreateProjectMedia.project_media, tried:\n\n          - `#\u003cClass:0x0000aaaae52f44d8\u003e#project_media`, which did not exist\n          - `TrueClass#project_media`, which did not exist\n          - Looking up hash key `:project_media` or `\"project_media\"` on `true`, but it wasn't a Hash\n\n          To implement this field, define one of the methods above (and check for typos)\n","code":5,"data":{}}]}
```

I'm sure I'm missing a lot of things about how this part of the code works. What should I be returning from the rescue? What am I missing here?

**And one last question:**
I tried testing using the models directly instead of going through the graphql test (since it can be slow). But I couldn't get it to work. I think because it doesn't really go through the graphql controller and that is where we adding the begin/rescue. Is that correct? And if it is, why do we only want to add this to graphql layer?

caiosba:
@vasconsaurus hummm... please push what you have so I can take a look :)

> I tried testing using the models directly instead of going through the graphql test (since it can be slow). But I couldn't get it to work.

For that, you should just test the `handle_fact_check_for_existing_claim` method directly.

vasconsaurus:
@caiosba I finished the changes we talked about, could you take a look and see if this is ready for review or if there is something missing?

app/graph/mutations/graphql_crud_operations.rb
```diff
@@ -7,7 +7,20 @@ def self.safe_save(obj, attrs, parent_names = [])
       obj.send(method, value) if obj.respond_to?(method)
     end
     obj.disable_es_callbacks = Rails.env.to_s == "test"
-    obj.save_with_version!
+
+    begin
+      obj.save_with_version!
+    rescue RuntimeError => e
+      if e.message.include?("This item already exists") &&
```
caiosba, line 14: Request: @vasconsaurus , this is not a good way to check for the error because of two reasons:

(1) If `I18n.locale` is not English, then this error message will be localized, so won't match this condition (for example, for Portuguese it will be "Este item já existe").

(2) The copy for the error message may change in the future.

I suggest you check for the exception class or some other non-localized information.

app/models/concerns/project_media_creators.rb
```diff
@@ -11,6 +11,28 @@ def create_metrics_annotation
...
+        claim_description: cd,
+        report_status: (fact_check['publish_report'] ? 'published' : 'unpublished'),
+        rating: self.set_status,
+        tags: self.set_tags.to_a.map(&:strip),
+        skip_check_ability: true
+      })
+    end
+    fc
+  end
+
```
caiosba, line 35: Question: Here you're just moving this method from `private` to `public`, not really changing anything, right?

app/models/project_media.rb
```diff
@@ -430,6 +430,31 @@ def apply_rules_and_actions_on_update
     self.team.apply_rules_and_actions(self, rule_ids)
   end
 
+  def self.handle_fact_check_for_existing_claim(existing_pm,new_pm)
+    if existing_pm.fact_check.blank?
+      existing_pm.appended_fact_check_from(new_pm)
```
caiosba, line 435: Suggestion: Since this method takes an action, I suggest to name it with the verb name in imperative form, so, `append_fact_check_from`.

app/models/project_media.rb
```diff
@@ -430,6 +430,31 @@ def apply_rules_and_actions_on_update
...
+    elsif existing_pm.fact_check.present?
+      if existing_pm.fact_check.language != new_pm.set_fact_check['language']
+        new_pm.replaced_with_blank_media
```
caiosba, line 438: Suggestion: Since this method takes an action, I suggest to name it with the verb name in imperative form, so, `replace_with_blank_media`.

caiosba (changes requested):
Looks great @vasconsaurus , glad that the approach we discussed has worked as expected! I left a few comments, but the approach is correct. Did the tests pass?

app/models/concerns/project_media_creators.rb
vasconsaurus: Yes. Once I changed the instance method to be a class one I had problems accessing `create_claim_description_and_fact_check`

vasconsaurus:
@caiosba 
> Did the tests pass?

Yes 🎉

app/graph/mutations/graphql_crud_operations.rb
vasconsaurus: @caiosba, questions:
- The error we get is a `RuntimeError`, so it's not super specific. It does send the lapis code in the message, so I changed it to look for that. ([here](https://github.com/meedan/check-api/pull/2079/commits/cb97946cc90cf337402d2424899a20387f085a9a))
- However, would it be worth for us to create a custom error? We get the error from here:
https://github.com/meedan/check-api/blob/cb97946cc90cf337402d2424899a20387f085a9a/app/models/concerns/project_association.rb#L139-L155

caiosba: @vasconsaurus , I don't think we need a specific code, I think that `DUPLICATED` is good enough for this case. But please use the variable `LapisConstants::ErrorCodes::const_get('DUPLICATED')` instead of the hard-coded value "9".

app/graph/mutations/graphql_crud_operations.rb
```diff
@@ -7,7 +7,20 @@ def self.safe_save(obj, attrs, parent_names = [])
       obj.send(method, value) if obj.respond_to?(method)
     end
     obj.disable_es_callbacks = Rails.env.to_s == "test"
-    obj.save_with_version!
+
+    begin
+      obj.save_with_version!
+    rescue RuntimeError => e
+      if e.message.include?("\"code\":#{LapisConstants::ErrorCodes::const_get('DUPLICATED')}") &&
```
caiosba, line 14: You could also do a `JSON.parse` on the message and check for the code, but this is fine too.

app/graph/mutations/graphql_crud_operations.rb
```diff
@@ -7,7 +7,20 @@ def self.safe_save(obj, attrs, parent_names = [])
...
+      obj.is_a?(ProjectMedia) &&
+      obj.set_fact_check.present? &&
+      obj.set_original_claim.present?
+        existing_pm = ProjectMedia.find(JSON.parse(e.message)['data']['id'])
```
caiosba, line 18: Related to my other comment: this `JSON.parse` could be at the top and then the parsed `data` object could be used to check for the `id` here and for the error code at the top.

app/models/project_media.rb
```diff
@@ -430,6 +430,31 @@ def apply_rules_and_actions_on_update
...
+    self
+  end
+
+  def replace_with_blank_media
+    m = Blank.create!
+    self.set_original_claim = nil
+    self.media_id = m.id
+    self.save!
+    self
+  end
```
caiosba, line 456: Suggestion for this whole block... I think it's more clear to:

- Explicitly say, after line 435, that we're returning the existing item: `return existing_pm`
- Explicitly say, after line 438, that we're returning the new item: `return new_pm`

And then in the methods `append_fact_check_from` and `replace_with_blank_media`, you don't need to return `self` in their last lines (447 and 455).

caiosba (approved):
Great work @vasconsaurus ! Thanks for addressing the comments from the previous reviews. I added a couple of suggestions mostly for clarity.

app/graph/mutations/graphql_crud_operations.rb
vasconsaurus: I tried this before, but we ran into an issue with test `GraphqlController12Test#test_should_not_create_feed_invitation` / `rails test test/controllers/graphql_controller_12_test.rb:69`.

```
[11, 20] in /app/app/graph/mutations/graphql_crud_operations.rb
   11:     begin
   12:       obj.save_with_version!
   13:     rescue RuntimeError => e
   14:       require 'byebug'
   15:       byebug
=> 16:       if e.message.include?("\"code\":#{LapisConstants::ErrorCodes::const_get('DUPLICATED')}") &&
   17:       obj.is_a?(ProjectMedia) &&
   18:       obj.set_fact_check.present? &&
   19:       obj.set_original_claim.present?
   20:         existing_pm = ProjectMedia.find(JSON.parse(e.message)['data']['id'])
(byebug) e
#<RuntimeError: No permission to create FeedInvitation>
(byebug) JSON.parse(e)
*** TypeError Exception: no implicit conversion of RuntimeError into String

nil
(byebug) JSON.parse(e.message)
*** JSON::ParserError Exception: unexpected token at 'No permission to create FeedInvitation'

nil
```

PR #2082, opened 2024-10-11
Do not send report if search results were already received.

caiosba:
## Description

I noticed a regression introduced by CV2-5451 that we overlooked. We shouldn't set `confirmed_by` and `confirmed_at` for all relationships, but really keep setting them only for when a relationship created as a suggestion is confirmed... this is the purpose of these two fields. Instead, the logic for inhering the status from the parent item and sending a report is that should have been fixed. This is what this PR does. So:

- Child item should always inherit status from parent item for confirmed relationships, regardless if they are created as confirmed matches or a suggestion that is confirmed... also, should be the same logic if the relationship is created by a bot or by a human
- But when a child item matches a parent item, we should NOT send a report if the user already received a report as search results... this means that we should not send a report for confirmed relationships created by bots

Reference: CV2-5451.

## How has this been tested?

Existing unit tests should cover this. I tested manually as well:

- Get a tipline search result: You should not also receive a report for the match that is created between the existing item and your media, because you already received search results
- Manually-created similar item: Send report
- Confirmed suggestion: Send report

In all cases above, the child item should inherit the status from the parent.

## Checklist

### DIFF
--- a/app/models/bot/smooch.rb
+++ b/app/models/bot/smooch.rb
@@ -63,7 +63,7 @@ class Bot::Smooch < BotUser
       relationship = Relationship.find_by_id(rid)
       unless relationship.nil?
         # A relationship created by the Smooch Bot or Alegre Bot is related to search results, so the user has already received the report as a search result - unless it's a suggestion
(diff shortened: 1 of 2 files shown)


DGaffney (approved):
:shipit:

PR #2083, opened 2024-10-12
CV2-5348: refactor ES cached field calling and remove retry_on_conflict

melsawy:
## Description

- Set ElasticSearch’s retry_on_conflict: 0 (so we can rely only on the Sidekiq retry, which retrieves latest value)
- Pass class and ID instead of the full object

References: CV2-5348

## How has this been tested?

Re-run automated tests.

## Checklist

### DIFF
--- a/app/lib/check_cached_fields.rb
+++ b/app/lib/check_cached_fields.rb
@@ -97,17 +97,13 @@ module CheckCachedFields
           update_pg: options[:update_pg],
           pg_field_name: options[:pg_field_name],
         }
-        obj_data = { klass: obj.class.name, id: obj.id }
-        self.delay_for(1.second).index_cached_field_bg(index_options, value, name, YAML::dump(obj_data))
+        self.delay_for(1.second).index_cached_field_bg(index_options, value, name, obj.class.name, obj.id)
       end
     end
 
-    def index_cached_field_bg(index_options, value, name, obj_data)
-      obj_data = begin YAML::load(obj_data) rescue {} end
-      unless obj.blank?
-        obj = obj_data[:klass].constantize.find_by_id obj_data[:id]
-        self.index_and_pg_cached_field(index_options, value, name, obj) unless obj.nil?
-      end
+    def index_cached_field_bg(index_options, value, name, klass, id)
+      obj = klass.constantize.find_by_id id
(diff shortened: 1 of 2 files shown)


app/lib/check_cached_fields.rb
```diff
@@ -97,12 +97,15 @@ def index_cached_field(options, value, name, obj)
           update_pg: options[:update_pg],
           pg_field_name: options[:pg_field_name],
         }
-        self.delay_for(1.second).index_cached_field_bg(index_options, value, name, obj)
+        obj_data = { klass: obj.class.name, id: obj.id }
+        self.delay_for(1.second).index_cached_field_bg(index_options, value, name, YAML::dump(obj_data))
```
caiosba, line 101: Question: Why pass an argument `obj_data` instead of two arguments `klass` and `id`? Is it just to avoid that current jobs fail? If so, I agree it's a good idea! Just want to confirm the intent. And if so, I think you should add a comment to the code to explain the rationale.

app/lib/check_elastic_search.rb
```diff
@@ -63,7 +63,7 @@ def update_elasticsearch_doc_bg(options)
       create_doc_if_not_exists(options)
       sleep 1
       client = $repository.client
-      client.update index: CheckElasticSearchModel.get_index_alias, id: options[:doc_id], retry_on_conflict: 3, body: { doc: fields }
+      client.update index: CheckElasticSearchModel.get_index_alias, id: options[:doc_id], body: { doc: fields }
```
caiosba, line 66: Question: Did you confirm that the default is zero? So we don't need to set it explicitly?

caiosba (approved):
Good to go if the answers to my two questions are "yes".

app/lib/check_cached_fields.rb
melsawy: The current jobs will fail event I pass an argument `obj_data`  but I use `object_data`  to be consistent with other implementation.
I agreed with your suggestion to pass id and klass it'll be more clear.

app/lib/check_elastic_search.rb
melsawy: Yes it's optional and the default is zero.

app/lib/check_cached_fields.rb
melsawy: Done

PR #2132, opened 2024-11-19
Bot Preview: Initial backend implementation

jayjay-w:
## Description

Add a bot_query field to TeamType which will send queries to the backend simulating tipline bot queries.

Initially, this is returning the most recent Explainers/FactChecks as TiplineSearchResults, but eventually it will be hooked to the actual bot implementation.

References: CV2-5702

## How has this been tested?

 - Confirmed that the most recent explainers/fact checks

## Checklist

### DIFF
--- a/app/graph/types/team_type.rb
+++ b/app/graph/types/team_type.rb
@@ -400,4 +400,56 @@ class TeamType < DefaultObject
     return nil unless User.current&.is_admin
     TeamStatistics.new(object, period, language, platform)
   end
+
+  field :bot_query, [TiplineSearchResultType], null: true do
+    argument :search_text, GraphQL::Types::String, required: true
+  end
+
+  def bot_query(search_text:)
+    unless User.current&.is_admin
+      raise GraphQL::ExecutionError, "You do not have permission to perform this action"
+    end
+
+    return nil unless search_text
+
+    explainers = object.explainers.order(created_at: :desc).limit(3)
+
+    fact_checks = FactCheck.joins(:claim_description)
+                           .where(claim_descriptions: { team_id: object.id })
+                           .order('fact_checks.created_at DESC')
+                           .limit(3)
+
+    combined_records = explainers + fact_checks
+    sorted_records = combined_records.sort_by(&:created_at).reverse
+
+    top_three = sorted_records.first(3)
+
+    results = top_three.map do |record|
+      if record.is_a?(Explainer)
+        TiplineSearchResult.new(
+          team: object,
+          title: record.title,
+          body: record.description,
+          image_url: nil,
+          language: record.language,
+          url: record.url,
+          type: :explainer,
+          format: :text
+        )
+      elsif record.is_a?(FactCheck)
+        TiplineSearchResult.new(
+          team: object,
+          title: record.title,
+          body: record.summary,
+          image_url: nil,
+          language: record.language,
+          url: record.url,
+          type: :fact_check,
+          format: :text
+        )
+      end
+    end
+
+    results
+  end
 end
--- a/app/graph/types/team_type.rb
+++ b/app/graph/types/team_type.rb
@@ -406,48 +406,22 @@ class TeamType < DefaultObject
   end
 
   def bot_query(search_text:)
-    unless User.current&.is_admin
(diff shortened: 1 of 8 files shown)


app/graph/types/team_type.rb
```diff
@@ -400,4 +400,56 @@ def statistics(period:, language: nil, platform: nil)
...
   end
+
+  field :bot_query, [TiplineSearchResultType], null: true do
+    argument :search_text, GraphQL::Types::String, required: true
+  end
+
+  def bot_query(search_text:)
+    unless User.current&.is_admin
+      raise GraphQL::ExecutionError, "You do not have permission to perform this action"
+    end
```
caiosba, line 411: Suggestion: Make this logic consistent with this one: https://github.com/meedan/check-api/blob/develop/app/graph/types/team_type.rb#L400. I suggest to implement a helper method that returns `nil` if the user is not a super-admin, and use this method here and above. This method could be defined in the base controller: https://github.com/meedan/check-api/blob/develop/app/controllers/api/v1/graphql_controller.rb.

app/graph/types/team_type.rb
```diff
@@ -400,4 +400,56 @@ def statistics(period:, language: nil, platform: nil)
...
+
+    return nil unless search_text
```
caiosba, line 413: Not a blocker: Not needed, since `search_text` is a required argument. This method won't even be executed if `search_text` is not provided.

app/graph/types/team_type.rb
```diff
@@ -400,4 +400,56 @@ def statistics(period:, language: nil, platform: nil)
...
+
+    explainers = object.explainers.order(created_at: :desc).limit(3)
+
+    fact_checks = FactCheck.joins(:claim_description)
+                           .where(claim_descriptions: { team_id: object.id })
+                           .order('fact_checks.created_at DESC')
+                           .limit(3)
```
caiosba, line 420: Suggestion: Order by `updated_at` instead of `created_at` - it gives the frontend team more flexibility on how to prepare their data locally.

app/graph/types/team_type.rb
```diff
@@ -400,4 +400,56 @@ def statistics(period:, language: nil, platform: nil)
...
+          team: object,
+          title: record.title,
+          body: record.summary,
+          image_url: nil,
+          language: record.language,
+          url: record.url,
+          type: :fact_check,
+          format: :text
+        )
+      end
```
caiosba, line 450: Please use the methods `as_tipline_search_result` we have defined for explainers: https://github.com/meedan/check-api/blob/develop/app/models/explainer.rb#L32. And implement a similar one for fact-checks and call it here.

caiosba (changes requested):
Thanks @jayjay-w , I left a few comments. Two other things please:

- This PR is currently targeting `develop`, please create an epic branch (sync with @amoedoamorim in order to create one with the same name as the frontend one) and point this PR to it.
- Please update the GraphQL schema files.

amoedoamorim:
Hi Jay! I'm trying to experiment with the API but I need 2 things from you:
Can you please push the relay schema (I did it locally to try the API), and can you also define a field `id` in `TiplineSearchResultType`? Relay needs to request it implicitly as per the screenshot below:

image

app/graph/types/team_type.rb
jayjay-w: It seems that methods in the GrapqlController are not exposed to `TeamType`. I will continue checking this and implement it in the next PR for the actual implementation.

But I made the check consistent with the one above.

jayjay-w:
Thank you, @amoedoamorim,

I added the `id` field to `TiplineSearchResultType`, let me know if it is OK now?

---

Also, I based this branch on an epic branch called `epic/bot-preview`, let me know if this name is sufficient for both frontend and backend.

amoedoamorim:
Thanks, Jay! I tried the changes and seem to work as expected! I'll now work on formatting the response data and display in the frontend! :+1: 

I have just one more ask, can you please rename the target branch to `epic/5571-bot-preview` so it matches the frontend one? This way we can run the integration tests using the matching backend branch

caiosba (approved):
Thanks @jayjay-w for addressing the comments from the last review! Looks good!

PR #2142, opened 2024-11-27
CV2-5761 list most relevant articles fact check and explainer for project media item

melsawy:
## Description

List most relevant FactCheck and Explainer for a ProjectMedia.
- [x] Include FactCheck and explainer for a ProjectMedia item
- [x] Fix article sort to be with score
- [x] Update the limit to be 3 (get the value from config) for team query and 10 (get the value from config) for item query

References: CV2-5761

## How has this been tested?

Implemented automated tests.

## Checklist

### DIFF
--- a/app/models/concerns/smooch_search.rb
+++ b/app/models/concerns/smooch_search.rb
@@ -21,7 +21,7 @@ module SmoochSearch
 
         # Search for explainers if fact-checks were not found
         if reports.empty? && query['type'] == 'text'
-          explainers = self.search_for_explainers(uid, query['text'], team_id, language).first(3).select{ |explainer| explainer.as_tipline_search_result.should_send_in_language?(language) }
+          explainers = self.search_for_explainers(uid, query['text'], team_id, language).select{ |explainer| explainer.as_tipline_search_result.should_send_in_language?(language) }
           Rails.logger.info "[Smooch Bot] Text similarity search got #{explainers.count} explainers while looking for '#{query['text']}' for team #{team_id}"
           results = explainers.collect{ |explainer| explainer.project_medias.to_a }.flatten.uniq.reject{ |pm| pm.blank? }.first(3)
           reports = explainers.map(&:as_tipline_search_result)
@@ -100,9 +100,10 @@ module SmoochSearch
       end
     end
 
-    def parse_search_results_from_alegre(results, after = nil, feed_id = nil, team_ids = nil)
+    def parse_search_results_from_alegre(results, after = nil, feed_id = nil, team_ids = nil, limit = nil)
+      limit ||= CheckConfig.get(:most_relevant_team_limit, 3, :integer)
       pms = reject_temporary_results(results).sort_by{ |a| [a[1][:model] != Bot::Alegre::ELASTICSEARCH_MODEL ? 1 : 0, a[1][:score]] }.to_h.keys.reverse.collect{ |id| Relationship.confirmed_parent(ProjectMedia.find_by_id(id)) }
-      filter_search_results(pms, after, feed_id, team_ids).uniq(&:id).sort_by{ |pm| pm.report_status == 'published' ? 0 : 1 }.first(3)
+      filter_search_results(pms, after, feed_id, team_ids).uniq(&:id).first(limit)
     end
 
     def date_filter(team_id)
@@ -160,7 +161,8 @@ module SmoochSearch
 
     # "type" is text, video, audio or image
     # "query" is either a piece of text of a media URL
-    def search_for_similar_published_fact_checks_no_cache(type, query, team_ids, after = nil, feed_id = nil, language = nil)
+    def search_for_similar_published_fact_checks_no_cache(type, query, team_ids, after = nil, feed_id = nil, language = nil, limit = nil)
+      limit ||= CheckConfig.get(:most_relevant_team_limit, 3, :integer)
       results = []
       pm = nil
       pm = ProjectMedia.new(team_id: team_ids[0]) if team_ids.size == 1 # We'll use the settings of a team instead of global settings when there is only one team
@@ -182,7 +184,7 @@ module SmoochSearch
           results = self.search_by_keywords_for_similar_published_fact_checks(words, after, team_ids, feed_id, language)
         else
           alegre_results = Bot::Alegre.get_merged_similar_items(pm, [{ value: self.get_text_similarity_threshold }], Bot::Alegre::ALL_TEXT_SIMILARITY_FIELDS, text, team_ids)
-          results = self.parse_search_results_from_alegre(alegre_results, after, feed_id, team_ids)
+          results = self.parse_search_results_from_alegre(alegre_results, after, feed_id, team_ids, limit)
           Rails.logger.info "[Smooch Bot] Text similarity search got #{results.count} results while looking for '#{text}' after date #{after.inspect} for teams #{team_ids}"
         end
       else
@@ -192,7 +194,7 @@ module SmoochSearch
         media_url = self.save_locally_and_return_url(media_url, type, feed_id)
         threshold = Bot::Alegre.get_threshold_for_query(type, pm)[0][:value]
         alegre_results = Bot::Alegre.get_items_with_similar_media_v2(media_url: media_url, threshold: [{ value: threshold }], team_ids: team_ids, type: type)
-        results = self.parse_search_results_from_alegre(alegre_results, after, feed_id, team_ids)
+        results = self.parse_search_results_from_alegre(alegre_results, after, feed_id, team_ids, limit)
         Rails.logger.info "[Smooch Bot] Media similarity search got #{results.count} results while looking for '#{query}' after date #{after.inspect} for teams #{team_ids}"
       end
       results
@@ -303,7 +305,8 @@ module SmoochSearch
       end
     end
 
-    def search_for_explainers(uid, query, team_id, language = nil)
+    def search_for_explainers(uid, query, team_id, language = nil, limit = nil)
+      limit ||= CheckConfig.get(:most_relevant_team_limit, 3, :integer)
       results = nil
       begin
         text = ::Bot::Smooch.extract_claim(query)
@@ -312,7 +315,7 @@ module SmoochSearch
           results = results.where(language: language) if !language.nil? && should_restrict_by_language?([team_id])
           results = results.order('updated_at DESC')
         else
-          results = Explainer.search_by_similarity(text, language, team_id)
+          results = Explainer.search_by_similarity(text, language, team_id, limit)
         end
       rescue StandardError => e
         self.handle_search_error(uid, e, language) unless uid.blank?
--- a/app/models/concerns/smooch_search.rb
+++ b/app/models/concerns/smooch_search.rb
@@ -9,6 +9,7 @@ module SmoochSearch
     def search(app_id, uid, language, message, team_id, workflow, provider = nil)
       platform = self.get_platform_from_message(message)
       begin
+        limit = CheckConfig.get(:most_relevant_team_limit, 3, :integer)
         sm = CheckStateMachine.new(uid)
         self.get_installation(self.installation_setting_id_keys, app_id) if self.config.blank?
         RequestStore.store[:smooch_bot_provider] = provider unless provider.blank?
@@ -21,7 +22,7 @@ module SmoochSearch
 
         # Search for explainers if fact-checks were not found
         if reports.empty? && query['type'] == 'text'
-          explainers = self.search_for_explainers(uid, query['text'], team_id, language).select{ |explainer| explainer.as_tipline_search_result.should_send_in_language?(language) }
+          explainers = self.search_for_explainerssearch_for_explainers(uid, query['text'], team_id, limit, language).select{ |explainer| explainer.as_tipline_search_result.should_send_in_language?(language) }
           Rails.logger.info "[Smooch Bot] Text similarity search got #{explainers.count} explainers while looking for '#{query['text']}' for team #{team_id}"
           results = explainers.collect{ |explainer| explainer.project_medias.to_a }.flatten.uniq.reject{ |pm| pm.blank? }.first(3)
           reports = explainers.map(&:as_tipline_search_result)
@@ -100,8 +101,7 @@ module SmoochSearch
       end
     end
 
-    def parse_search_results_from_alegre(results, after = nil, feed_id = nil, team_ids = nil, limit = nil)
-      limit ||= CheckConfig.get(:most_relevant_team_limit, 3, :integer)
+    def parse_search_results_from_alegre(results, limit, after = nil, feed_id = nil, team_ids = nil)
       pms = reject_temporary_results(results).sort_by{ |a| [a[1][:model] != Bot::Alegre::ELASTICSEARCH_MODEL ? 1 : 0, a[1][:score]] }.to_h.keys.reverse.collect{ |id| Relationship.confirmed_parent(ProjectMedia.find_by_id(id)) }
       filter_search_results(pms, after, feed_id, team_ids).uniq(&:id).first(limit)
     end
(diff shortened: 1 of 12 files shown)


app/models/team.rb
```diff
@@ -564,8 +564,10 @@ def filter_by_keywords(query, filters, type = 'FactCheck')
   end
 
   def search_for_similar_articles(query, pm = nil)
-    # query expected to be text
-    result_ids = Bot::Smooch.search_for_similar_published_fact_checks_no_cache('text', query, [self.id]).map(&:id)
+    # query:  expected to be text
+    # pm: to request a most relevant to specific item and also include both FactCheck & Explainer
+    limit = pm.nil? ? CheckConfig.get(:most_relevant_team_limit, 3, :integer) : CheckConfig.get(:most_relevant_item_limit, 10, :integer)
```
caiosba, line 569: Do we have any case currently where this method is going to be called with a `nil` `pm`? I don't think so... and in that case, we can simplify this method and only handle the case where there is a `pm`, so we just need a single configuration key and we always return both fact-checks and explainers.

app/models/explainer.rb
```diff
@@ -107,7 +107,8 @@ def self.update_paragraphs_in_alegre(id, previous_paragraphs_count, timestamp)
     end
   end
 
-  def self.search_by_similarity(text, language, team_id)
+  def self.search_by_similarity(text, language, team_id, limit = nil)
+    limit ||= CheckConfig.get(:most_relevant_team_limit, 3, :integer)
```
caiosba, line 111: I suggest we always enforce a `limit` to be passed (see other comments).

app/models/concerns/smooch_search.rb
```diff
@@ -304,7 +306,8 @@ def ask_for_feedback_when_all_search_results_are_received(app_id, language, work
       end
     end
 
-    def search_for_explainers(uid, query, team_id, language = nil)
+    def search_for_explainers(uid, query, team_id, language = nil, limit = nil)
+      limit ||= CheckConfig.get(:most_relevant_team_limit, 3, :integer)
```
caiosba, line 310: As I said in another comment, I suggest we always enforce a `limit` to be passed, instead of accepting `nil` and falling back to a default value.

app/models/concerns/smooch_search.rb
```diff
@@ -160,7 +161,8 @@ def search_for_similar_published_fact_checks(type, query, team_ids, after = nil,
 
     # "type" is text, video, audio or image
     # "query" is either a piece of text of a media URL
-    def search_for_similar_published_fact_checks_no_cache(type, query, team_ids, after = nil, feed_id = nil, language = nil)
+    def search_for_similar_published_fact_checks_no_cache(type, query, team_ids, after = nil, feed_id = nil, language = nil, limit = nil)
+      limit ||= CheckConfig.get(:most_relevant_team_limit, 3, :integer)
```
caiosba, line 165: Like the other comments: I suggest we always enforce a `limit` to be passed.

app/models/concerns/smooch_search.rb
```diff
@@ -100,9 +100,10 @@ def reject_temporary_results(results)
       end
     end
 
-    def parse_search_results_from_alegre(results, after = nil, feed_id = nil, team_ids = nil)
+    def parse_search_results_from_alegre(results, after = nil, feed_id = nil, team_ids = nil, limit = nil)
+      limit ||= CheckConfig.get(:most_relevant_team_limit, 3, :integer)
```
caiosba, line 104: Like the other comments: I suggest we always enforce a `limit` to be passed.

app/models/concerns/smooch_search.rb
```diff
@@ -21,7 +21,7 @@ def search(app_id, uid, language, message, team_id, workflow, provider = nil)
 
         # Search for explainers if fact-checks were not found
         if reports.empty? && query['type'] == 'text'
-          explainers = self.search_for_explainers(uid, query['text'], team_id, language).first(3).select{ |explainer| explainer.as_tipline_search_result.should_send_in_language?(language) }
+          explainers = self.search_for_explainers(uid, query['text'], team_id, language).select{ |explainer| explainer.as_tipline_search_result.should_send_in_language?(language) }
```
caiosba, line 24: This line is only called by the tipline search. So here we could pass to `search_for_explainers` the configured value for the maximum number of results for tipline search (3 by default).

app/models/explainer.rb
```diff
@@ -121,7 +122,7 @@ def self.search_by_similarity(text, language, team_id)
     }
     response = Bot::Alegre.query_sync_with_params(params, "text")
     results = response['result'].to_a.sort_by{ |result| result['_score'] }
```
caiosba, line 124: Please replace this sorting by the same one we have for fact-checks (https://github.com/meedan/check-api/blob/7bac066b91f1dca8395af1c707520c980d44e7bf/app/models/concerns/smooch_search.rb#L105):

`sort_by{ |a| [a[1][:model] != Bot::Alegre::ELASTICSEARCH_MODEL ? 1 : 0, a[1][:score]] }`

Or, add this as an action item to CV2-5617.

caiosba (changes requested):
Sawy, I left a few comments, most of them related to the same thing:

1. Let's not have a default value for the `limit` argument: Enforcing a value will also force us be sure that we're using it consistently.
2. Then, there are two code paths:
  2.1 For tipline search, pass the `limit` as the configuration key value (default: 3) from `Bot::Smooch.search` method, to `search_for_explainers` and  to `search_for_similar_published_fact_checks`.
  2.2 For the "most recent to most relevant" feature, pass the `limit` as the configuration key value (default: 10) from `Team.search_for_similar_articles` to `search_for_similar_published_fact_checks_no_cache` and to `search_for_explainers`

I also suggested a refactoring for `Team.search_for_similar_articles` since AFAIK we just have one use case we need to support for this method for now. So, we can simplify and reduce the complexity there by supporting the only case we need.

One more thing: Where are the tests?

app/models/team.rb
melsawy: There are two cases for calling this method:

1. From `TeamType`  the one should called by Jay 
2. From `ProjectMediaType`  to get most relevant 

case 1: should return FactCheck and fallback to Explainer if no FactCheck with limit 3
case 2: should return both FactCheck & Explainer  with limit 10

caiosba: Ah, you're right, my bad!

melsawy:
> Sawy, I left a few comments, most of them related to the same thing:
> 
> 1. Let's not have a default value for the `limit` argument: Enforcing a value will also force us be sure that we're using it consistently.
> 2. Then, there are two code paths:
>    2.1 For tipline search, pass the `limit` as the configuration key value (default: 3) from `Bot::Smooch.search` method, to `search_for_explainers` and  to `search_for_similar_published_fact_checks`.
>    2.2 For the "most recent to most relevant" feature, pass the `limit` as the configuration key value (default: 10) from `Team.search_for_similar_articles` to `search_for_similar_published_fact_checks_no_cache` and to `search_for_explainers`
> 
> I also suggested a refactoring for `Team.search_for_similar_articles` since AFAIK we just have one use case we need to support for this method for now. So, we can simplify and reduce the complexity there by supporting the only case we need.
> 
Done
> One more thing: Where are the tests?
Tests were added in previous PR

- [Team test](https://github.com/meedan/check-api/blob/CV2-5761-list-most-relevant-articles-fact-check-and-explainer-for-project-media-item/test/models/team_2_test.rb#L1553-L1574)
- [ProjectMedia test](https://github.com/meedan/check-api/blob/CV2-5761-list-most-relevant-articles-fact-check-and-explainer-for-project-media-item/test/models/project_media_7_test.rb#L119-L142)

app/models/explainer.rb
melsawy: Done

app/models/concerns/smooch_search.rb
melsawy: Done

app/models/concerns/smooch_search.rb
melsawy: Done

app/models/concerns/smooch_search.rb
melsawy: Done

app/models/concerns/smooch_search.rb
melsawy: Done

app/models/explainer.rb
melsawy: will add it as action item to CV2-5617

jayjay-w (approved):

app/models/concerns/smooch_search.rb
```diff
@@ -9,19 +9,20 @@ module ClassMethods
     def search(app_id, uid, language, message, team_id, workflow, provider = nil)
       platform = self.get_platform_from_message(message)
       begin
+        limit = CheckConfig.get('most_relevant_team_limit', 3, :integer)
         sm = CheckStateMachine.new(uid)
         self.get_installation(self.installation_setting_id_keys, app_id) if self.config.blank?
         RequestStore.store[:smooch_bot_provider] = provider unless provider.blank?
         query = self.get_search_query(uid, message)
-        results = self.get_search_results(uid, query, team_id, language).collect{ |pm| Relationship.confirmed_parent(pm) }.uniq
+        results = self.get_search_results(uid, query, team_id, language, limit).collect{ |pm| Relationship.confirmed_parent(pm) }.uniq
```
caiosba, line 17: The limit here should be much higher, I think we can even hard code it to 100, or something like that. Because further down we apply the limit over the results that actually have published reports (see comment below).

app/models/concerns/smooch_search.rb
```diff
@@ -9,19 +9,20 @@ module ClassMethods
...
         reports = results.select{ |pm| pm.report_status == 'published' }.collect{ |pm| pm.get_dynamic_annotation('report_design') }.reject{ |r| r.nil? }.collect{ |r| r.report_design_to_tipline_search_result }.select{ |r| r.should_send_in_language?(language) }
 
         # Extract explainers from matched media if they don't have published fact-checks but they have explainers
         reports = results.collect{ |pm| pm.explainers.to_a }.flatten.uniq.first(3).map(&:as_tipline_search_result) if !results.empty? && reports.empty?
```
caiosba, line 21: This `first(3)` is the one that should be replaced by the configuration key (see comment above).

app/models/concerns/smooch_search.rb
```diff
@@ -9,19 +9,20 @@ module ClassMethods
...
 
         # Search for explainers if fact-checks were not found
         if reports.empty? && query['type'] == 'text'
-          explainers = self.search_for_explainers(uid, query['text'], team_id, language).select{ |explainer| explainer.as_tipline_search_result.should_send_in_language?(language) }
+          explainers = self.search_for_explainers(uid, query['text'], team_id, limit, language).select{ |explainer| explainer.as_tipline_search_result.should_send_in_language?(language) }
```
caiosba, line 25: Same as for fact-checks: The limit here should be much higher, I think we can even hard code it to 100, or something like that. Because further down we apply the limit over the results that actually match the criteria (see comment below).

app/models/concerns/smooch_search.rb
```diff
@@ -9,19 +9,20 @@ module ClassMethods
...
           Rails.logger.info "[Smooch Bot] Text similarity search got #{explainers.count} explainers while looking for '#{query['text']}' for team #{team_id}"
           results = explainers.collect{ |explainer| explainer.project_medias.to_a }.flatten.uniq.reject{ |pm| pm.blank? }.first(3)
```
caiosba, line 27: This first(3) is the one that should be replaced by the configuration key (see comment above).

app/models/team.rb
```diff
@@ -566,8 +566,8 @@ def filter_by_keywords(query, filters, type = 'FactCheck')
   def search_for_similar_articles(query, pm = nil)
     # query:  expected to be text
     # pm: to request a most relevant to specific item and also include both FactCheck & Explainer
-    limit = pm.nil? ? CheckConfig.get(:most_relevant_team_limit, 3, :integer) : CheckConfig.get(:most_relevant_item_limit, 10, :integer)
-    result_ids = Bot::Smooch.search_for_similar_published_fact_checks_no_cache('text', query, [self.id], nil, nil, nil, limit).map(&:id)
+    limit = pm.nil? ? CheckConfig.get('most_relevant_team_limit', 3, :integer) : CheckConfig.get('most_relevant_item_limit', 10, :integer)
+    result_ids = Bot::Smooch.search_for_similar_published_fact_checks_no_cache('text', query, [self.id], limit).map(&:id)
```
caiosba, line 570: For this one it's OK to pass the `limit` here because we don't do any other filtering.

caiosba (changes requested):
Almost there @melsawy ! There is still a change we need to make at the tipline search level because we need to be aware of the other filters that are applied.

app/models/concerns/smooch_search.rb
melsawy: @caiosba the limit work as the existing behavior for example the `get_search_results`  called one of the following methods

- [parse_search_results_from_alegre](https://github.com/meedan/check-api/blob/develop/app/models/concerns/smooch_search.rb#L103-L106) 
- [search_by_keywords_for_similar_published_fact_checks](https://github.com/meedan/check-api/blob/develop/app/models/concerns/smooch_search.rb#L103-L106)

and both methods in the current develop branch limit the results to 3 so in this refactor I passed the limit through the parent method `get_search_results`

so you suggest to send the limit to `get_search_results` with higher value (100 for example) and after apply all filters limit the result to 3 (the limit from config value) right?

my suggestion to keep the current implementation as this refactor run with old one and just read the limit from the config value and file a ticket with a refactor to make sure to return a limit after apply all filter as there is no issue reported that there is an expected result should return by tipline search for example.

caiosba: Understood, @melsawy ... I reviewed the changes since my last review, let me do a full PR re-review in order to be sure that everything is consistent.

app/models/explainer.rb
caiosba: You can click on "Resolve conversation" once you do so.

app/models/concerns/smooch_search.rb
melsawy: Done

caiosba (approved):

app/models/team.rb
```diff
@@ -564,8 +564,10 @@ def filter_by_keywords(query, filters, type = 'FactCheck')
   end
 
   def search_for_similar_articles(query, pm = nil)
-    # query expected to be text
-    result_ids = Bot::Smooch.search_for_similar_published_fact_checks_no_cache('text', query, [self.id]).map(&:id)
+    # query:  expected to be text
+    # pm: to request a most relevant to specific item and also include both FactCheck & Explainer
+    limit = pm.nil? ? CheckConfig.get('most_relevant_team_limit', 3, :integer) : CheckConfig.get('most_relevant_item_limit', 10, :integer)
+    result_ids = Bot::Smooch.search_for_similar_published_fact_checks_no_cache('text', query, [self.id], limit).map(&:id)
```
skyemeedan, line 570: If we are calling via the `_published_fact_checks` endpoint, how do we support the requirement to return also the unpublished fact_checks for the 'relevant' use case?

test/controllers/graphql_controller_5_test.rb
```diff
@@ -405,6 +405,25 @@ def setup
...
+  test "should get relevant articles for ProjectMedia item" do
+    t = create_team
+    pm = create_project_media quote: 'Foo Bar', team: t
+    ex = create_explainer language: 'en', team: t, title: 'Foo Bar'
+    pm.explainers << ex
+    cd = create_claim_description description: pm.title, project_media: pm
+    fc = create_fact_check claim_description: cd, title: pm.title
+    items = FactCheck.where(id: fc.id) + Explainer.where(id: ex.id)
+    ProjectMedia.any_instance.stubs(:get_similar_articles).returns(items)
+    query = "query { project_media(ids: \"#{pm.id}\") { relevant_articles_count, relevant_articles {  edges { node { ... on FactCheck { dbid }, ... on Explainer { dbid } } } } } }"
```
skyemeedan, line 417: @danielevalverde does this test align with the query used in the front end to pull the relevant_articles?

app/models/explainer.rb
melsawy: Added as subtask as I am going to merge CV2-5617 [https://meedan.atlassian.net/browse/CV2-5779](https://meedan.atlassian.net/browse/CV2-5779)

app/models/team.rb
melsawy: There is a filter for published report so we can skip it to return both published & unpublished fact-checks
we can move the conversation to jira ticket as I'll merge the PR to epic branch

test/controllers/graphql_controller_5_test.rb
melsawy: @skyemeedan I added this test as a reference for front-end and I'll add a comment to jira ticket and if there is a missing fields required by @danielevalverde I can add these fields to epic branch

PR #2143, opened 2024-11-27
5607 – Data dashboard table initial state should be an empty table

vasconsaurus:
## Description

Before we would return `nil`, meaning no table would be displayed.
Since we always want to display the table, even if it's empty, we need to return a structure.

### Notes
- I had to add the `team.default_language` so the first row would be displayed, and not only the header.
- Three hyphens are showing by the side of the export button (they are highlighted in the video below). I think I might need to fix it on the front-end side. But I'm not sure how we want to deal with it.

### Video
https://github.com/user-attachments/assets/184de1c5-7455-4186-a587-d14a878ce91b

References: CV2-5607

### Notes
- I added a new test to make sure we are returning all the columns
- I updated the older tests to check if Month has an hyphen, instead of `data_report` being `nil`. This still makes sure we are starting from an empty `data_report`.

### Tests
should return an existing empty data report structure if no statistics exist
`rails test test/models/team_test.rb:1133` 
 
should default to Rails cache for data report if monthly team statistics not present
`rails test test/models/team_test.rb:1164`
 
should return data report with chronologically ordered items, preferring the MonthlyTeamStatistics when present
`rails test test/models/team_test.rb:1172`

- Etc.

## Checklist


vasconsaurus:
@caiosba when you have some time, can you take a look at this and see if it makes sense?
If this makes sense, and is in the correct direction, a question: What do you think is the best way to deal with the three hyphens that show by the side of the export button? (image below)

Screenshot 2024-11-27 at 17 04 26

caiosba:
@vasconsaurus , let's look into it together in our 1:1 - I want to find a way where we don't need to repeat the column names. But yes, the direction is correct! For the three dots in the UI, it may require a simple fix on the frontend side.

vasconsaurus:
@caiosba 

> a way where we don't need to repeat the column names

I kept thinking about this yesterday after I was done for the day, there is a method in `MonthlyTeamStatistic` we can use. So I updated it to do that.

vasconsaurus:
@caiosba regarding tests, just to confirm:

- I think the tests are running twice?
    -  Build and Run Tests / unit-tests (pull_request) – This one had one failing test.
    - Build and Run Tests / unit-tests (push) – The tests failed to run.
- This test failed, but seems unrelated: `CheckS3Test#test_should_manage_files_in_S3 [/app/test/lib/check_s3_test.rb:22]`

Should we be running those two? If yes, then should I re-run the one that didn't run?
Is this failing test the flaky one you were talking about earlier, or should I worry about it?

caiosba:
> I think the tests are running twice?

Hum... good point... @chinelo-obitube is this expected?

> Is this failing test the flaky one you were talking about earlier, or should I worry about it?

Seems unrelated to your changes, but it's not the one I was talking about... so, since we're not in a hurry, I suggest you just re-run.

PR #2153, opened 2024-12-13
Ensure that null is not returned as the item when creating an imported fact-check with original media in the GraphQL API.

caiosba:
## Description

When handling existing media to determine whether the imported fact-check will be added to or appended to the existing media, there was a scenario where the media already contained a fact-check in the same language being imported. In that scenario, nil was being returned. This PR fixes the issue by ensuring that a new object is returned in such cases, allowing the validations to handle whether the object is persisted or fails.

Fixes: CV2-5804.

## How has this been tested?

TDD. I added a unit test that was able to reproduce the issue.

## Checklist

### DIFF
--- a/app/models/project_media.rb
+++ b/app/models/project_media.rb
@@ -430,7 +430,7 @@ class ProjectMedia < ApplicationRecord
     self.team.apply_rules_and_actions(self, rule_ids)
   end
 
-  def self.handle_fact_check_for_existing_claim(existing_pm,new_pm)
+  def self.handle_fact_check_for_existing_claim(existing_pm, new_pm)
     if existing_pm.fact_check.blank?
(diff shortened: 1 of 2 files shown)


caiosba:
@vasconsaurus can you please also review this since you worked on the feature CV2-5441?

vasconsaurus:
@caiosba so in this case a new object is returned, but in the end it's creation will eventually fail since a media with the fact-check in the same language already exists?

vasconsaurus (approved):

caiosba:
@vasconsaurus , an easier way to explain this, is: this method, at the model level, shouldn't know anything about the API where it's used. We can read its responsibility this way: "Given two items, choose which one to return given the existing and the new fact-check". Is it more clear put that way? :)

PR #2160, opened 2024-12-17
Smooch Bot should not create a relationship between Media and TiplineResource

vasconsaurus:
## Description
We saw an error in Sentry where the creation of a `Relationship` between a `ProjectMedia` and a `TiplineResource` was attempted. The creation fails, but it should not be attempted in the first place.

**Error**
- `StandardError`
- `Unable to create new relationship as requested: Related items must exist in the same workspace`
- exception_class: `ActiveRecord::RecordInvalid`
- exception_message: `Related items must exist in the same workspace`

**For this to happen**
- The media can't be of type `text`, because we do check if it's a `ProjectMedia` in that case,
- it needs to have a `caption`,
- there needs to be an existing `ProjectMedia` with the same `id` as the `TiplineResource`,
	- if there isn't a  `ProjectMedia` with the same `id` as the `TiplineResource` we get a different error: `#<ActiveRecord::RecordNotFound:0x7ab20>`.

**Note on the error**
- The error is rescued, Sentry is notified and it's logged,
- so we need to check if Sentry was notified or/and if the error was logged to confirm it happened.

### Solution
We return early, and not even attempt to relate items if it's not a `ProjectMedia`, since `Relationship's` are only for `ProjectMedia`.

References: CV2-5676

## How has this been tested?

test "should not try to create relationship between media and tipline resource": `rails test test/models/bot/smooch_7_test.rb:636`

1. We create a `ProjectMedia`,
2. we create a `TiplineResource` and make sure it has the same id as the `ProjectMedia`,
3. we create a message that sends a video with a caption, to make sure it hits the condition we need it to hit,
4. it shouldn't notify Sentry or log an error, since it shouldn't get that far,
5. it should not try to create a Relationship between `ProjectMedia` and a `TiplineResource`

- Etc.

## Checklist

### DIFF
--- a/app/models/concerns/smooch_messages.rb
+++ b/app/models/concerns/smooch_messages.rb
@@ -430,7 +430,6 @@ module SmoochMessages
     end
 
     def smooch_relate_items_for_same_message(message, associated, app_id, author, request_type, associated_obj)
-      return unless associated_obj.is_a?(ProjectMedia)
       if !message['caption'].blank?
         # Check if message contains caption then create an item and force relationship
         self.relate_item_and_text(message, associated, app_id, author, request_type, associated_obj, Relationship.confirmed_type)
@@ -457,6 +456,7 @@ module SmoochMessages
       target = self.create_project_media_from_message(message)
       unless target.nil?
         smooch_post_save_message_actions(message, target, app_id, author, request_type, associated_obj)
+        return unless associated_obj.is_a?(ProjectMedia)
         Relationship.create_unless_exists(associated.id, target.id, relationship_type)
       end
     end
--- a/app/models/concerns/smooch_messages.rb
+++ b/app/models/concerns/smooch_messages.rb
@@ -430,6 +430,7 @@ module SmoochMessages
     end
 
     def smooch_relate_items_for_same_message(message, associated, app_id, author, request_type, associated_obj)
+      return if associated.class.name != 'ProjectMedia'
       if !message['caption'].blank?
         # Check if message contains caption then create an item and force relationship
         self.relate_item_and_text(message, associated, app_id, author, request_type, associated_obj, Relationship.confirmed_type)
--- a/app/models/concerns/smooch_messages.rb
+++ b/app/models/concerns/smooch_messages.rb
@@ -430,6 +430,7 @@ module SmoochMessages
     end
 
     def smooch_relate_items_for_same_message(message, associated, app_id, author, request_type, associated_obj)
+      return unless associated.is_a?(ProjectMedia)
       if !message['caption'].blank?
         # Check if message contains caption then create an item and force relationship
         self.relate_item_and_text(message, associated, app_id, author, request_type, associated_obj, Relationship.confirmed_type)
@@ -456,7 +457,6 @@ module SmoochMessages
       target = self.create_project_media_from_message(message)
       unless target.nil?
         smooch_post_save_message_actions(message, target, app_id, author, request_type, associated_obj)
-        return unless associated_obj.is_a?(ProjectMedia)
         Relationship.create_unless_exists(associated.id, target.id, relationship_type)
       end
     end
--- a/app/models/concerns/smooch_messages.rb
+++ b/app/models/concerns/smooch_messages.rb
@@ -395,6 +395,8 @@ module SmoochMessages
     end
 
     def save_message(message_json, app_id, author = nil, request_type = 'default_requests', associated_id = nil, associated_class = nil)
+    # associated: is the the "first media" and will be the source of the Relationship
+    # associated_obj: is used for TiplineRequest (smooch_resource_id field)
       message = JSON.parse(message_json)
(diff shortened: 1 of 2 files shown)


vasconsaurus:
@caiosba I thought about returning early if it's not a `ProjectMedia`, are there other classes we can/should be able to create a `Relationship` for?

app/models/concerns/smooch_messages.rb
```diff
@@ -430,6 +430,7 @@ def smoooch_post_save_message_actions(message, associated, app_id, author, reque
     end
 
     def smooch_relate_items_for_same_message(message, associated, app_id, author, request_type, associated_obj)
+      return if associated.class.name != 'ProjectMedia'
```
vasconsaurus, line 433: @caiosba then it's best to change this to `return unless associated_obj.is_a?(ProjectMedia)`

caiosba: Yes, it's more Ruby-ish and easier to read :)

That's something I would say when I review :D

melsawy (approved):

vasconsaurus:
@caiosba there are two tests failing that seem related to this change. Will look into it.

Another unrelated question: Is there a reason why we need to run tests twice both on pull_request and push?

vasconsaurus:
@melsawy since you worked on the Tipline, could youanswer some questions? 🙏 

- We have two failing tests:
    1. rails test test/models/bot/smooch_3_test.rb:327
    2. rails test test/models/bot/smooch_3_test.rb:194

- For the first one: Why do we expect three TiplineRequests if we are sending two messages here: https://github.com/meedan/check-api/blob/b8a52086848e481caa1956ee7dbc7748e77a6c38/test/models/bot/smooch_3_test.rb#L376. And like the one below we also get `NilClass` here
- The second one: It fails because when it gets to `smooch_relate_items_for_same_message` the class is `NilClass`, so it returns early and does not create a `Relationship`. Do you have any idea on why that is happening?

caiosba:
> Another unrelated question: Is there a reason why we need to run tests twice both on pull_request and push?

Let me ask Chinel on Slack.

vasconsaurus:
> Let me ask Chinel on Slack.

I know we asked her before if this was expected, but I was wondering on why 😅

caiosba (changes requested):
Code looks good! Let's just fix the two tests that broke.

app/models/concerns/smooch_messages.rb
```diff
@@ -430,6 +430,7 @@ def smoooch_post_save_message_actions(message, associated, app_id, author, reque
     end
 
     def smooch_relate_items_for_same_message(message, associated, app_id, author, request_type, associated_obj)
+      return unless associated_obj.is_a?(ProjectMedia)
```
melsawy, line 433: I believe it's better to move the conditions before this line(https://github.com/meedan/check-api/blob/develop/app/models/concerns/smooch_messages.rb#L459) and check associated object type

melsawy:
> @melsawy since you worked on the Tipline, could youanswer some questions? 🙏
> 
> * We have two failing tests:
>   
>   1. rails test test/models/bot/smooch_3_test.rb:327
>   2. rails test test/models/bot/smooch_3_test.rb:194
> * For the first one: Why do we expect three TiplineRequests if we are sending two messages here: https://github.com/meedan/check-api/blob/b8a52086848e481caa1956ee7dbc7748e77a6c38/test/models/bot/smooch_3_test.rb#L376
>   . And like the one below we also get `NilClass` here
> * The second one: It fails because when it gets to `smooch_relate_items_for_same_message` the class is `NilClass`, so it returns early and does not create a `Relationship`. Do you have any idea on why that is happening?

@vasconsaurus 
This first test generated three TiplineRequests because one of the two messages includes a caption with long text which means creation of new media with the caption text and linking it to the media item.

For the second instance, I proposed relocating the condition to the relate_item_and_text method before [this line](https://github.com/meedan/check-api/blob/develop/app/models/concerns/smooch_messages.rb#L459).

vasconsaurus:
@caiosba and @melsawy, tests are finally passing, so updates!

### 1.
It seems like the issue was that I was looking at the wrong object. I should have been checking `associated` instead of `associated_obj`. So I added two comments explaining the difference:

```
# associated: is the the "first media" and will be the source of the Relationship
# associated_obj: is used for TiplineRequest (smooch_resource_id field) 
```

I think it would be nice if we could rename this to something more clear. However, since I have no suggestions I added the comments 😅 

### 2.
I moved `return unless associated.is_a?(ProjectMedia)` earlier in the code. Since `Relationships` should only be between `ProjectMedia`, I don't think there should be a circumstance where we are trying to relate something else, so we can stop early, but defer to the both of you.

caiosba (approved):
Great, thanks Manu! I'm glad we spent the time we needed to really understand what was going on here and to reproduce the issue as a unit test. Great work!

PR #2167, opened 2025-01-03
Explainers should be indexed using `team_id`, not `team_slug`

jayjay-w:
## Description

Currently, fact checks are indexed with `team_id`, but explainers are indexed with `team_slug`. Here we are changing the indexing for explainers to use `team_id`.

We are also adding a rake task to re-index  existing explainers.

References: CV2-5825

## How to test
Confirm that the rake task runs successfully:

```ruby
bundle exec rake check:migrate:reindex_explainers
```

## Checklist


caiosba:
Tagging @DGaffney and @computermacgyver to review as well.

app/models/explainer.rb
```diff
@@ -61,7 +61,7 @@ def self.update_paragraphs_in_alegre(id, previous_paragraphs_count, timestamp)
 
     base_context = {
       type: 'explainer',
-      team: explainer.team.slug,
+      team: explainer.team.id,
```
DGaffney, line 64: I believe this needs to be `team_id` instead?

app/models/explainer.rb
```diff
@@ -107,7 +107,7 @@ def self.search_by_similarity(text, language, team_id, limit)
     models_thresholds = Explainer.get_alegre_models_and_thresholds(team_id)
     context = {
       type: 'explainer',
-      team: Team.find(team_id).slug
+      team: Team.find(team_id).id
```
DGaffney, line 110: Same here - team_id instead of team I think.

caiosba:
@jayjay-w please rebase - @melsawy fixed that test failure in `develop`.

lib/tasks/data/reindex_explainers.rake
```diff
@@ -0,0 +1,60 @@
...
+                  next if field_value.to_s.empty?
+  
+                  doc = Bot::Alegre.send_to_text_similarity_index_package(
+                    pm,
+                    field,
+                    field_value,
+                    Bot::Alegre.item_doc_id(pm, field)
+                  )
+  
+                  Bot::Alegre.query_async_with_params(doc, Bot::Alegre.get_pm_type(pm))
```
DGaffney, line 37: Looks good!

DGaffney (changes requested):
Change team to team_id unless I'm missing something?

app/models/explainer.rb
caiosba: In this case, both should work, but `explainer.team_id` is "faster".

app/models/explainer.rb
caiosba: Yes, no need to load the team object from the ID and to then get the ID... just replace this whole thing by the `team_id` variable.

lib/tasks/data/reindex_explainers.rake
caiosba, line 1: For consistency, I think it's better to have this rake task under `lib/tasks/migrate` than `lib/tasks/data`. Please follow the file naming convention there.

lib/tasks/data/reindex_explainers.rake
```diff
@@ -0,0 +1,60 @@
+namespace :check do
+    namespace :data do
+      # bundle exec rake check:data:reindex_explainers
```
caiosba, line 3: When moving this file to lib/tasks/migrate, please update the namespace and comment accordingly.

lib/tasks/data/reindex_explainers.rake
```diff
@@ -0,0 +1,60 @@
...
+      desc 'Reindex all explainers in the system'
+      task :reindex_explainers => [:environment] do
+        old_logger = ActiveRecord::Base.logger
+        ActiveRecord::Base.logger = nil
+  
+        errors = []
+        current_time = Time.now
+        event_id = ENV['EVENT_ID'] || Digest::MD5.hexdigest("reindex_all_explainers_#{current_time.to_i}")
```
caiosba, line 11: Looks like you copy-and-pasted this line from some other rake task but it's not used for anything here. So please delete this line.

lib/tasks/data/reindex_explainers.rake
```diff
@@ -0,0 +1,60 @@
...
+  
+        puts "[#{current_time}] Starting reindex of all explainers"
+        
+        begin
+            explainers_query = ProjectMedia.joins("INNER JOIN annotations ON annotations.annotated_id = project_medias.id AND annotations.annotated_type = 'ProjectMedia'")
+            .where(annotations: { annotation_type: 'explainer' })
```
caiosba, line 17: Explainer is not an annotation type, it's a model... you made the change from team slug to team id in the `Explainer` model, you should iterate through `Explainer.all` (paginated, of course). In order for the rake task to be idempotent, I suggest that you store in Redis the last ID that was migrated and iterate from there. There are many examples of that in the codebase, under `lib/tasks/migrate`, but in a nutshell, the convention we use is something like:

```ruby
cache_key = 'check:migrate:reindex_explainers:last_migrated_id'

last_migrated_id = Rails.cache.read(cache_key).to_i

Explainer.where('id > ?', last_migrated_id).order('id ASC').find_each do |explainer|
  # Code to re-index explainer goes here
  Rails.cache.write(cache_key, explainer.id)
end

Rails.cache.delete(cache_key)
```

lib/tasks/data/reindex_explainers.rake
caiosba, line 37: This logic is _right_ for project media fields (that's what Devin means by "looks good"), but for explainers we just index the whole explainer object, and we do so paragraph by paragraph. So, this whole block needs to be replaced. 

Here is how an explainer is indexed: https://github.com/meedan/check-api/blob/b14402df83bdaece5f3174876cda8ebe3cfb86d8/app/models/explainer.rb#L55.

For each `explainer`, I think you can do:

```ruby
Explainer.update_paragraphs_in_alegre(explainer.id, 0, Time.now.to_f)
```

lib/tasks/data/reindex_explainers.rake
```diff
@@ -0,0 +1,60 @@
...
+                    field_value,
+                    Bot::Alegre.item_doc_id(pm, field)
+                  )
+  
+                  Bot::Alegre.query_async_with_params(doc, Bot::Alegre.get_pm_type(pm))
+                end
+              rescue StandardError => e
+                errors << e
+                puts "[#{Time.now}] Error reindexing ProjectMedia with ID #{pm.id}: #{e.message}"
+                CheckSentry.notify(e, project_media_id: pm.id)
```
caiosba, line 42: We don't need to report a rake task error to Sentry, please remove this line.

lib/tasks/data/reindex_explainers.rake
```diff
@@ -0,0 +1,60 @@
...
+              end
+            end
+          end
+  
+          if errors.any?
+            raise Check::ExplainerReindexing::IncompleteRunError.new("Failed to reindex #{errors.length} explainers")
```
caiosba, line 48: Is this constant defined anywhere? Since this is a one-off rake task, you can just `raise('Error message')`.

test/lib/tasks/reindex_explainers_test.rb
```diff
@@ -0,0 +1,46 @@
+require_relative '../../test_helper'
+require 'rake'
+
+class ReindexExplainersTest < ActiveSupport::TestCase
+  def setup
+    super
```
caiosba, line 6: We shouldn't need to call `super` here (it makes tests slower).

test/lib/tasks/reindex_explainers_test.rb
```diff
@@ -0,0 +1,46 @@
...
+    @explainer_pm = create_project_media
+    create_annotation(annotation_type: @explainer_annotation_type, annotated: @explainer_pm)
+
+    # Mock Alegre behavior
+    Bot::Alegre.stubs(:query_async_with_params).returns("done")
+    Bot::Alegre.stubs(:send_to_text_similarity_index_package).returns({})
+  end
+
+  def teardown
+    super
```
caiosba, line 26: We shouldn't need to call super here (it makes tests slower).

test/lib/tasks/reindex_explainers_test.rb
caiosba, line 1: Please update tests based on the changes to the rake task itself.

caiosba (changes requested):
@jayjay-w, there are some inconsistencies here. Please review my comments and let me know if you have any questions. In general, let's try to avoid cycles of code/review/deploy/test by testing things locally first. These issues could have been identified if you had tested the changes locally beforehand.

Please run the following tests in your local development environment, using your local Telegram tipline, before resubmitting for review:

1. Create an explainer using the old Check API code (from develop, prior to these changes).
2. Ensure it is indexed in Alegre (you should see the Alegre calls in the docker-compose output).
3. Send a query (more than 3 words) to the tipline and confirm that the explainer is returned as a search result.
4. Switch to the new version of the code (i.e., the changes in this branch).
5. Send the same query to the tipline and confirm that the explainer is not returned as a search result (since it was indexed with the team slug but is being queried with the team ID).
6. Run the rake task to re-index the explainers.
7. Send the same query to the tipline and confirm that the explainer is returned as a search result.

This means that you should make sure that Alegre/Presto is running fine in your local development environment, which is important for other tasks, so it's worth the effort. Devin can help with that setup.

Please keep me updated on your progress with each of these steps, and let me know if you need assistance at any point.

app/models/explainer.rb
jayjay-w: Thank you both, I updated this to `team_id`

lib/tasks/migrate/20250108070424_reindex_explainers.rake
```diff
@@ -0,0 +1,27 @@
...
+    puts "[#{Time.now}] Starting reindex of all explainers"
+
+    Explainer.where('id > ?', last_migrated_id).order('id ASC').find_each(batch_size: 100) do |explainer|
+      begin
+        Explainer.update_paragraphs_in_alegre(explainer.id, 0, Time.now.to_f)
+        Rails.cache.write(cache_key, explainer.id)
+        puts "[#{Time.now}] Successfully reindexed explainer with ID #{explainer.id}"
+      rescue StandardError => e
+        Rails.logger.error "[#{Time.now}] Error reindexing explainer with ID #{explainer.id}: #{e.message}"
+        CheckSentry.notify(e, explainer_id: explainer.id)
```
caiosba, line 19: Please remove this line, we don't need to notify Sentry from rake tasks.

lib/tasks/migrate/20250108070424_reindex_explainers.rake
```diff
@@ -0,0 +1,27 @@
...
+      end
+    end
+
+    Rails.cache.delete(cache_key)
+    puts "[#{Time.now}] Successfully reindexed all explainers"
```
caiosba, line 24: Not sure if it's the GitHub UI messing things up, but the indentation looks wrong here.

test/lib/tasks/migrate/20250108070424_reindex_explainers_test.rb
```diff
@@ -0,0 +1,34 @@
...
+    Rake::Task.define_task(:environment)
+    Rake::Task['check:migrate:reindex_explainers'].reenable
+
+    @explainer = create_explainer
+    Explainer.stubs(:update_paragraphs_in_alegre).returns(true)
+    Rails.cache.write('check:migrate:reindex_explainers:last_migrated_id', 0)
+  end
+
+  def teardown
+    [@explainer].each(&:destroy)
```
caiosba, line 16: Why iterate over an array of one?

caiosba (changes requested):

lib/tasks/data/reindex_explainers.rake
jayjay-w: Done!

caiosba:
@DGaffney , since you requested changes, kindly re-review.

caiosba (approved):
Thanks for applying comments from code review. Looks good now. I also tested it locally (according to the test plan I shared here before) and it works well. I'll add details to the ticket.

app/models/explainer.rb
```diff
@@ -61,7 +61,7 @@ def self.update_paragraphs_in_alegre(id, previous_paragraphs_count, timestamp)
 
     base_context = {
       type: 'explainer',
-      team: explainer.team.slug,
+      team: explainer.team_id,
```
computermacgyver, line 64: Let's rename `team` to `team_id` to be consistent with how we index fact-checks?
https://github.com/meedan/check-api/blob/819bc4a3d3a834de2be6f5bee5e50e69ef996b76/app/models/concerns/alegre_v2.rb#L264

app/models/explainer.rb
```diff
@@ -107,7 +107,7 @@ def self.search_by_similarity(text, language, team_id, limit)
     models_thresholds = Explainer.get_alegre_models_and_thresholds(team_id)
     context = {
       type: 'explainer',
-      team: Team.find(team_id).slug
+      team: team_id
```
computermacgyver, line 110: As above, I'd rename `team` to `team_id` to be consistent with how we index fact-checks

computermacgyver (changes requested):
Minor suggestion to better align with how we index fact-checks
(see https://github.com/meedan/check-api/blob/819bc4a3d3a834de2be6f5bee5e50e69ef996b76/app/models/concerns/alegre_v2.rb#L263-L267)

In a similar move, I also think we should also change `context: base_context.merge({ paragraph: count })` and `context: base_context.merge({ field: 'title' })` to include `explainer_id: explainer.id` similar to how fact-checks index a `project_media_id` in their context.

app/models/explainer.rb
caiosba: Since I was still here... done!

app/models/explainer.rb
caiosba: Since I was still here... done!

caiosba:
Thanks @computermacgyver , since I was still on this PR, I made this change and re-tested locally. Kindly re-review.

jayjay-w:
@computermacgyver I reviewed Caio's change and approved+merged this PR. Let me know in case you catch anything I might have missed! Thank you very much for your suggestion.

PR #2173, opened 2025-01-09
Bug – CV2-5913 – Cannot update a new record

vasconsaurus:
## Description

This showed up as a `ActiveRecord::ActiveRecordError: cannot update a new record` that was related to an error we saw on Zapier.

### Context

This error only happens in a very specific context:
- we have an existing `ProjectMedia` with a `LinkMedia`
- we make a `graphql` request with:
    - the same link in `set_original_claim`
    - the same language
    - the `title` field is requested as part of the response 

### Causes
This was a combination of different changes.

- `project_media_creators.rb:create_link_media`: When we create the `LinkMedia` from `original_claim`, because links are global, we check if it already exists. This caused an issue when the `LinkMedia` already existed for the current team. 
```
  def create_link_media(url)
    team = self.team || Team.current
    pender_key = team.get_pender_key if team
    url_from_pender = Link.normalized(url, pender_key)
    Link.find_by(url: url_from_pender) || Link.create!(url: url, pender_key: pender_key)
  end
```
- `project_media.rb:self.handle_fact_check_for_existing_claim(existing_pm, new_pm)` : The above wasn't an issue before, but we started to return an object that was not persisted:
```
  def self.handle_fact_check_for_existing_claim(existing_pm, new_pm)
    if existing_pm.fact_check.blank?
      existing_pm.append_fact_check_from(new_pm)
      return existing_pm
    elsif existing_pm.fact_check.present?
      if existing_pm.fact_check.language != new_pm.set_fact_check['language']
        new_pm.replace_with_blank_media
        return new_pm
      end
    end
    new_pm
  end
```
- Because of this combination it found a `LinkMedia`, so it returned an `original_title`, but it tried to update that title from the new object that was not created (because it was a duplicate). In other words: 
    - The `ProjectMedia` was not created because it was a duplicate, but because it had the `original_title` information from the `LinkMedia`, it tried to update the title of the non-existing item anyway.

### Fix
- ~~When checking if the `LinkMedia` exists, we make sure we only return it if it does not exist in the current team. Else we run `Link.create!` and let the validations deal with the duplicate.~~
- ~~For robustness: we make sure we return a persisted object (it may avoid other cases in the future).~~
- After running all the tests and some further digging I realized that we need the error to be raised later for our  rescuing in graphql_crud_operations.rb to work. So I only made sure we try to return a persisted object or an error is raised.

References: CV2-5913

## How has this been tested?
This was tested manually:
    1. You can make the same graphql request twice, or
    2. you can create a new cluster and add a fact check manually, then make a graphql request with the same same link in set_original_claim and the same language

We also wrote a test: `rails test test/controllers/graphql_controller_12_test.rb:859`

- Etc.

## Checklist


caiosba (approved):
We've been pairing on this, so this was an easy review at this point - good to go, thanks Manu!

PR #2189, opened 2025-01-22
feature – 6021 – Store original claim in Media object

vasconsaurus:
## Description

To debug items created using the API (e.g., through the Zapier integration), improve performance, and avoid duplications, it is important to store the `original_claim_url` in the `Media` object created when this attribute is present.

When a `ProjectMedia` is created (e.g., via a GraphQL mutation) and the `original_claim` attribute (a URL) is provided, a new `Media` instance is created from that URL and associated with the `ProjectMedia` instance. Currently, the reference between the original claim and the created media is lost. This reference needs to be stored.

References: CV2-6021

### `project_media_7_test.rb`

"should save the original_claim url and original_claim_hash when Link Media is created from original_claim"
`rails test test/models/project_media_7_test.rb:200`

"should save the original_claim text and original_claim_hash when Claim media is created from original_claim"
`rails test test/models/project_media_7_test.rb:222`

"should save the original_claim url and original_claim_hash when UploadedAudio Media is created from original_claim"
`rails test test/models/project_media_7_test.rb:231`

"should check if the original_claim text exists and return that instance when trying to create claim media"
`rails test test/models/project_media_7_test.rb:247`

"should check if the original_claim url exists and return that instance when trying to create UploadedAudio media"
`rails test test/models/project_media_7_test.rb:256`

### `media_test.rb`

"Claim Media: should save the original_claim and original_claim_hash when created from original claim"
`rails test test/models/media_test.rb:631`

"Claim Media: should not save original_claim and original_claim_hash when not created from original claim"
`rails test test/models/media_test.rb:640`

"Claim Media: should not create duplicate media if media with original_claim_hash exists"
`rails test test/models/media_test.rb:646`

"Link Media: should save the original_claim and original_claim_hash when created from original claim"
`rails test test/models/media_test.rb:652`

"Link Media: should not save original_claim and original_claim_hash when not created from original claim"
`rails test test/models/media_test.rb:674`

"Link Media: should not create duplicate media if media with original_claim_hash exists"
`rails test test/models/media_test.rb:693`

"Uploaded Media: should save the original_claim and original_claim_hash when created from original claim"
`rails test test/models/media_test.rb:713`

"Uploaded Media: should not save original_claim and original_claim_hash when not created from original claim"
`rails test test/models/media_test.rb:729`

"Uploaded Media: should not create duplicate media if media with original_claim_hash exists"
`rails test test/models/media_test.rb:736`

- Etc.

## Checklist

### DIFF
--- a/app/models/media.rb
+++ b/app/models/media.rb
@@ -85,16 +85,16 @@ class Media < ApplicationRecord
 
       case content_type
       when /^image\//
-        create_uploaded_file_media_from_original_claim('UploadedImage', claim, ext)
+        find_or_create_uploaded_file_media_from_original_claim('UploadedImage', claim, ext)
       when /^video\//
-        create_uploaded_file_media_from_original_claim('UploadedVideo', claim, ext)
+        find_or_create_uploaded_file_media_from_original_claim('UploadedVideo', claim, ext)
       when /^audio\//
-        create_uploaded_file_media_from_original_claim('UploadedAudio', claim, ext)
+        find_or_create_uploaded_file_media_from_original_claim('UploadedAudio', claim, ext)
       else
-        create_link_media_from_original_claim(claim, project_media_team)
+        find_or_create_link_media_from_original_claim(claim, project_media_team)
       end
     else
-      create_claim_media_from_original_claim(claim)
+      find_or_create_claim_media_from_original_claim(claim)
     end
   end
 
@@ -130,7 +130,7 @@ class Media < ApplicationRecord
     self.original_claim_hash = Digest::MD5.hexdigest(original_claim) unless self.original_claim.blank?
   end
 
-  def self.create_uploaded_file_media_from_original_claim(media_type, url, ext)
+  def self.find_or_create_uploaded_file_media_from_original_claim(media_type, url, ext)
     klass = media_type.constantize
     existing_media = klass.find_by(original_claim_hash: Digest::MD5.hexdigest(url))
     
@@ -152,11 +152,11 @@ class Media < ApplicationRecord
     file
   end
 
-  def self.create_claim_media_from_original_claim(text)
+  def self.find_or_create_claim_media_from_original_claim(text)
     Claim.find_by(original_claim_hash: Digest::MD5.hexdigest(text)) || Claim.create!(quote: text, original_claim: text)
   end
 
-  def self.create_link_media_from_original_claim(url, project_media_team)
+  def self.find_or_create_link_media_from_original_claim(url, project_media_team)
     pender_key = project_media_team.get_pender_key if project_media_team
     url_from_pender = Link.normalized(url, pender_key)
     Link.find_by(url: url_from_pender) || Link.find_by(original_claim_hash: Digest::MD5.hexdigest(url)) || Link.create!(url: url, pender_key: pender_key, original_claim: url)
--- a/app/models/media.rb
+++ b/app/models/media.rb
@@ -80,7 +80,7 @@ class Media < ApplicationRecord
   def self.find_or_create_from_original_claim(claim, project_media_team)
     if claim.match?(/\A#{URI::DEFAULT_PARSER.make_regexp(['http', 'https'])}\z/)
       uri = URI.parse(claim)
-      content_type = fetch_content_type(uri)
+      content_type = Net::HTTP.get_response(uri)['content-type']
       ext = File.extname(uri.path)
 
       case content_type
@@ -128,11 +128,6 @@ class Media < ApplicationRecord
 
   def set_original_claim_hash
     self.original_claim_hash = Digest::MD5.hexdigest(original_claim) unless self.original_claim.blank?
-  end
-
-  def self.fetch_content_type(uri)
-    response = Net::HTTP.get_response(uri)
-    response['content-type']
   end
 
   def self.create_uploaded_file_media_from_original_claim(media_type, url, ext)
--- a/app/models/media.rb
+++ b/app/models/media.rb
@@ -133,7 +133,7 @@ class Media < ApplicationRecord
   def self.find_or_create_uploaded_file_media_from_original_claim(media_type, url, ext)
     klass = media_type.constantize
     existing_media = klass.find_by(original_claim_hash: Digest::MD5.hexdigest(url))
-    
+
     if existing_media
       existing_media
     else
--- a/app/models/media.rb
+++ b/app/models/media.rb
@@ -84,16 +84,16 @@ class Media < ApplicationRecord
 
       case content_type
       when /^image\//
-        create_uploaded_file_media_from_url('UploadedImage', claim, ext)
+        create_uploaded_file_media_from_original_claim('UploadedImage', claim, ext)
       when /^video\//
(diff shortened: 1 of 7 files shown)


vasconsaurus:
@caiosba I moved the code from `ProjectMediaCreators` to `Media`, I think there is still work to be done, but I have some questions:

1. The code below was originally was in `ProjectMediaCreators`, so it was a concern. I think this makes more sense as a general concern instead of a Media method, but I'm not sure. Not sure where it would fit as a concern as well, maybe something related to http requests?
https://github.com/meedan/check-api/blob/384a84d569ca32ae2e0ecfb9e2adf793eee4cd40/app/models/media.rb#L129-L132

2. We moved creating claim, link and uploaded file medias to Media. But we still have a lot of very similar code still in `ProjectMediaCreators`. Does it make sense to keep both? Does it make sense to keep them in `Media` and `ProjectMediaCreators`?
https://github.com/meedan/check-api/blob/c115a52ab695f2f6062d394930f628b87deef6ab/app/models/concerns/project_media_creators.rb#L169-L206

3. Regarding `original_claim_hash`: right now every time I create a Media, I create a hash from the url or text. I'm wondering if there is a better way to do this. Maybe a callback that generates and add the hash when `original_claim` is present?

4. I'm wondering about the conceptual difference between a `ProjectMedia` that is created from an `original_claim` and a `ProjectMedia` that is created from a `Media` through the UI. Why are they handled differently? And, if a Media that wasn't created from an `original_claim` exists, then it won't have the `original_claim_hash`, then if we try to create it through zapier could we create a duplicate `Media`? I think I might be confusing `Media` and `ProjectMedia` again though.

caiosba:
@vasconsaurus , answers inline below:

> 1. The code below was originally was in `ProjectMediaCreators`, so it was a concern. I think this makes more sense as a general concern instead of a Media method, but I'm not sure. Not sure where it would fit as a concern as well, maybe something related to http requests?

I just saw that this method is used in only one place, do I don't think it should even be a method... I think you can just move the method body to the place where it's called.

> 2. We moved creating claim, link and uploaded file medias to Media. But we still have a lot of very similar code still in ProjectMediaCreators. Does it make sense to keep both? Does it make sense to keep them in Media and ProjectMediaCreators?

I see what you mean... let's leave those for later, in order to avoid confusion, and focus first on moving only the ones related to the `original_claim` workflow... it will be easier to compare after that.

> 3. Regarding original_claim_hash: right now every time I create a Media, I create a hash from the url or text. I'm wondering if there is a better way to do this. Maybe a callback that generates and add the hash when original_claim is present?

Sorry if I wasn't clear before! Yes, that should be a callback like `before_validate :generate_original_claim_hash, on: :create`. We have similar examples in the code base.

> 4. I'm wondering about the conceptual difference between a ProjectMedia that is created from an original_claim and a ProjectMedia that is created from a Media through the UI. Why are they handled differently? And, if a Media that wasn't created from an original_claim exists, then it won't have the original_claim_hash, then if we try to create it through zapier could we create a duplicate Media? I think I might be confusing Media and ProjectMedia again though.

That case will be harder to happen, because from the UI it's a file upload, which will probably have a different hash than the file hosted in a URL. Also, with the original claim hash approach, we can have a performance gain by avoiding an unnecessary HTTP request, while for a file upload, we need to receive the whole file anyway.

vasconsaurus:
@caiosba on my prior questions:
1. I moved the method body and removed the method
2. Moved the ones related to original_claim
3. Moved generating the original_claim_hash to a callback
4. Still want to talk about this in our 1:1

Besides that I also:
1. added uniqueness validation
2. added tests to Media
3. updated some tests in ProjectMedia

Do you think I can move this to ready for review? Or is there any work I forgot before doing that?

app/models/media.rb
```diff
@@ -103,4 +125,40 @@ def set_type
...
     self.update_column(:uuid, self.id)
   end
+
+  def set_original_claim_hash
+    self.original_claim_hash = Digest::MD5.hexdigest(original_claim) unless self.original_claim.blank?
+  end
+
+  def self.create_uploaded_file_media_from_original_claim(media_type, url, ext)
+    klass = media_type.constantize
+    file = download_file(url, ext)
```
caiosba, line 135: Please move this to inside the `else` block below - the idea of this method is to only download the file if the hash doesn't exist.

app/models/media.rb
```diff
@@ -103,4 +125,40 @@ def set_type
   def set_uuid
     self.update_column(:uuid, self.id)
   end
+
+  def set_original_claim_hash
+    self.original_claim_hash = Digest::MD5.hexdigest(original_claim) unless self.original_claim.blank?
+  end
+
+  def self.create_uploaded_file_media_from_original_claim(media_type, url, ext)
```
caiosba, line 133: I suggest you rename this method to `find_or_create_[...]`.

app/models/media.rb
```diff
@@ -103,4 +125,40 @@ def set_type
...
+    raise "Invalid URL when creating media from original claim attribute" unless url =~ /\A#{URI::DEFAULT_PARSER.make_regexp(['http', 'https'])}\z/
+
+    file = Tempfile.new(['download', ext])
+    file.binmode
+    file.write(URI(url).open.read)
+    file.rewind
+    file
+  end
+
+  def self.create_claim_media_from_original_claim(text)
```
caiosba, line 155: I suggest you rename this method to `find_or_create_[...]`.

app/models/media.rb
```diff
@@ -103,4 +125,40 @@ def set_type
...
+    Claim.find_by(original_claim_hash: Digest::MD5.hexdigest(text)) || Claim.create!(quote: text, original_claim: text)
+  end
+
+  def self.create_link_media_from_original_claim(url, project_media_team)
```
caiosba, line 159: I suggest you rename this method to `find_or_create_[...]`.

db/migrate/20250124155814_add_original_claim_to_media.rb
```diff
@@ -0,0 +1,7 @@
+class AddOriginalClaimToMedia < ActiveRecord::Migration[6.1]
+  def change
+    add_column :medias, :original_claim, :text, null: true
+	  add_column :medias, :original_claim_hash, :string, null: true
+	  add_index :medias, :original_claim_hash, unique: true
```
caiosba, line 5: Please fix the indentation.

test/models/project_media_7_test.rb
```diff
@@ -196,4 +196,76 @@ def setup
       ProjectMedia.handle_fact_check_for_existing_claim(pm1, pm2)
     end
   end
+
+  test "should save the original_claim url and original_claim_hash when Link Media is created from original_claim" do
+    setup_elasticsearch
```
caiosba, line 201: I don't think we need ElasticSearch in this test, right? Please remove if possible (it makes tests slower).

caiosba (changes requested):
Great stuff @vasconsaurus ! I left some comments, but it's great overall!

caiosba (approved):
Thanks for addressing all comments from the previous review, Manu!

:ship:

jayjay-w:
I reviewed this although it has already been merged and all the changes look good to me!

vasconsaurus:
Thank you @jayjay-w!

PR #2201, opened 2025-02-04
CV2-6041: Verify the relationship target before establishing a new one.

melsawy:
## Description

Verify the relationship target before establishing a new one to prevent triggering errors due to database unique validation.

References: CV2-6041

## How has this been tested?

Re-run automated tests.

## Checklist


app/models/relationship.rb
vasconsaurus, line 1: @melsawy maybe this is a dumb question: but why not have this logic after the `if r.nil?` in line 164? Why do we want to check for this in line 161 and then again in 164, but then do something else?

vasconsaurus: Thank you for the explanation @melsawy!

app/models/relationship.rb
```diff
@@ -158,6 +158,7 @@ def create_or_update_parent_id
 
   def self.create_unless_exists(source_id, target_id, relationship_type, options = {})
     r = Relationship.where(source_id: source_id, target_id: target_id).where('relationship_type = ?', relationship_type.to_yaml).last
+    r = Relationship.where(target_id: target_id).last if r.nil?
```
caiosba, line 161: @melsawy, there is a unique index for the `target_id` column ([here](https://github.com/meedan/check-api/blob/develop/db/schema.rb#L622))... so, these two lines are redundant. I think we can replace these two lines (160 and 161) but just this:

```ruby
r = Relationship.where(target_id: target_id).last
```

Right?

And please fix the [failing test](https://github.com/meedan/check-api/actions/runs/13140070412/job/36664836781?pr=2201) (`RelationshipTest#test_should_create_relationship_unless_exists`).

caiosba (changes requested):

app/models/relationship.rb
melsawy: @caiosba there is a relationship_type condition; for example, if the same `source_id` and `target_id` exist but with different types, we should ignore this one and attempt to create a new relationship with the new type, as there is a [callback](https://github.com/meedan/check-api/blob/develop/app/models/relationship.rb#L23) before creating to destroy the existing one.
In this case, when the target_id is assigned to another source, I prefer to maintain two conditions but ensure that the target is not assigned to another source and swap the two conditions.
I'll push the new changes before re-request your review

caiosba: OK Sawy, makes sense!

app/models/relationship.rb
```diff
@@ -157,7 +157,8 @@ def create_or_update_parent_id
   end
 
   def self.create_unless_exists(source_id, target_id, relationship_type, options = {})
-    r = Relationship.where(source_id: source_id, target_id: target_id).where('relationship_type = ?', relationship_type.to_yaml).last
+    r = Relationship.where(target_id: target_id).where.not(source_id: source_id).last
+    r = Relationship.where(source_id: source_id, target_id: target_id).where('relationship_type = ?', relationship_type.to_yaml).last unless r.nil?
```
caiosba, line 161: Sorry @melsawy , it's still not clear to me what's the idea here (and I think that we need a comment in this code to explain it, like we had in the other parts of this method).

melsawy: Done and added more comments

jayjay-w (approved):

caiosba (approved):

PR #2216, opened 2025-02-12
CV2-6152: inherit status to confirmed items

melsawy:
## Description

All confirmed items should inherit the status of their parent item. When an item matches, it should receive the parent item's status. Additionally, if the parent status changes, the new status should be applied to all confirmed or matched items..

References: CV2-6152

## How to test?

- Re-run automated stest.
- Manually:
- Change item status and check confirmed items
- Move item from suggestion to confirmed and check the child status

## Checklist


app/models/relationship.rb
```diff
@@ -26,12 +26,12 @@ class Relationship < ApplicationRecord
   after_update :reset_counters, prepend: true
   after_update :propagate_inversion
   after_save :turn_off_unmatched_field, if: proc { |r| r.is_confirmed? || r.is_suggested? }
-  after_save :move_explainers_to_source, if: proc { |r| r.is_confirmed? }
+  after_save :move_explainers_to_source, :apply_status_to_children, if: proc { |r| r.is_confirmed? }
   before_destroy :archive_detach_to_list
   after_destroy :update_counters, prepend: true
   after_destroy :turn_on_unmatched_field, if: proc { |r| r.is_confirmed? || r.is_suggested? }
   after_commit :update_counter_and_elasticsearch, on: [:create, :update]
-  after_commit :update_counters, :destroy_elasticsearch_relation, on: :destroy
```
melsawy, line 34: I removed `update_counters` callback as it's already exists in `after_destroy` L31

app/models/bot/smooch.rb
```diff
@@ -58,36 +58,26 @@ def suggestion_accepted?
...
     end
 
-    def self.inherit_status_and_send_report(rid)
+    def self.smooch_send_report(rid)
       relationship = Relationship.find_by_id(rid)
       unless relationship.nil?
         target = relationship.target
-        parent = relationship.source
-
         # Always inherit status for confirmed matches
```
caiosba, line 65: If you're moving the status inheritance part somewhere else, please update this comment accordingly.

app/models/relationship.rb
```diff
@@ -380,6 +398,10 @@ def move_explainers_to_source
     end
   end
 
+  def apply_status_to_children
+    Relationship.delay_for(1.second, { queue: 'smooch_priority' }).replicate_status_to_children(self.source_id, User.current&.id, Team.current&.id)
+  end
```
caiosba, line 403: Sawy, can you please explain why we need to replicate the status for all children when a single relationship is saved? Is it only because it's "easier" (and we ignore if the statuses are the same anyway)?

melsawy: Yes Caio as I am using same method but you are right I should update the target item status only

app/models/relationship.rb
```diff
@@ -205,6 +205,23 @@ def self.create_unless_exists(source_id, target_id, relationship_type, options =
...
+    return if pm.nil?
+    User.current = User.where(id: uid).last
+    Team.current = Team.where(id: tid).last
+    status = pm.last_verification_status
+    pm.source_relationships.confirmed.find_each do |relationship|
+      s = relationship.target.last_status_obj
+      next if s.nil? || s.status == status
+      s.status = status
+      s.bypass_status_publish_check = true
+      s.save!
```
caiosba, line 219: Sawy, I think this part can be refactored as its own instance method for the `Relationship` model, something like `sync_statuses`.

app/models/relationship.rb
```diff
@@ -380,6 +397,17 @@ def move_explainers_to_source
...
 
+  def apply_status_to_target
+    status = self.source.last_verification_status
+    target = self.target
+    s = target.last_status_obj
+    unless s.nil? || s.status == status
+      s.status = status
+      s.bypass_status_publish_check = true
+      s.save
+    end
```
caiosba, line 408: Sawy, I think this part can be refactored as its own instance method for the `Relationship` model, something like `sync_statuses`.

caiosba (changes requested):
Thanks, Sawy! I think this implementation is much more consistent and clean. The code looks good now - thanks for the changes. I left one suggestion for refactoring.

Since this PR affects some really core and critical features, I tried to list the things we need to test manually as part of this work. Please add the test plan to the ticket later (when deployed to QA) and include anything you think is missing:

- One relationship created
- Single suggestion confirmed
- Status change for parent item
- Status change for fact-check

I'm not approving it yet because the tests are still failing - please request a re-review once all tests pass.

Great work!

app/models/relationship.rb
melsawy: Done

app/models/concerns/smooch_messages.rb
```diff
@@ -518,25 +518,6 @@ def should_try_to_send_report?(request_type, annotated)
...
-        self.get_installation(self.installation_setting_id_keys, data['app_id']) if self.config.blank?
-        message = parent.team.get_status_message_for_language(status, data['language'])
-        unless message.blank?
-          response = self.send_message_to_user(data['authorId'], message, {}, false, true, 'status_change')
-          self.save_smooch_response(response, parent, data['received'].to_i, 'fact_check_status', data['language'], { message: message })
-          requestors_count += 1
-        end
-      end
-      CheckNotification::InfoMessages.send('sent_message_to_requestors_on_status_change', status: pm.status_i18n, requestors_count: requestors_count) if requestors_count > 0
-    end
```
caiosba, line 538: Why you removed this method, @melsawy ? This is not the one related to the Slack message... this is the one that sends a message to the tipline user when the status changes. We need to keep it.

app/models/concerns/smooch_status.rb
```diff
@@ -32,14 +28,6 @@ def get_status_message_for_language(status, language)
...
         end
       end
-
-      protected
-
-      def send_message
-        pm = self.annotation.annotated
-        return unless Bot::Smooch.team_has_smooch_bot_installed(pm)
-        ::Bot::Smooch.delay_for(1.second, { queue: 'smooch_priority', retry: 0 }).send_message_on_status_change(pm.id, self.value, self.class.actor_session_id)
-      end
```
caiosba, line 48: Why you removed this method, @melsawy ? This is not the one related to the Slack message... this is the one that sends a message to the tipline user when the status changes. We need to keep it.

caiosba (changes requested):

app/models/concerns/smooch_messages.rb
melsawy: Back this one

app/models/concerns/smooch_status.rb
melsawy: Back this method

test/models/bot/slack_test.rb
```diff
@@ -132,28 +132,29 @@ def teardown
...
+  #       s = pm.annotations.where(annotation_type: pm.default_project_media_status_type).last.load
+  #       s.status = 'in_progress'
+  #       s.disable_es_callbacks = true
+  #       s.save!
+  #     end
+  #   end
+  #   RequestStore.store[:disable_es_callbacks] = false
+  #   assert_equal 6, WebMock::RequestRegistry.instance.times_executed(stub.request_pattern)
+  #   WebMock.allow_net_connect!
+  # end
```
caiosba, line 157: @melsawy , is this going to affect coverage?

melsawy: No I checked the coverage and still 100%

caiosba (approved):
Thanks Sawy for applying the comments from all reviews :)

PR #2223, opened 2025-02-17
6064 – Refactor duplicated media creation methods

vasconsaurus:
## Description

The `Media` class and the `ProjectMediaCreators` concern, in Check API, have some duplicated logic around creating different types of media. The main differences were: 
- the ones in the `Media` class handle `original_claim`
- the ones in `ProjectMediaCreators` expected a `ProjectMedia` object, 
- and the ones in the `Media` class expected the claim/url/file 

References: CV2-6064

## How to test?

Because Media is such an important class, we should run the whole suite. But while working on this I mainly ran:
```ruby
# media tests
rails test test/models/media_test.rb

# has tests that call set_media, which call the code we are changing
rails test test/models/bot/smooch_3_test.rb

# have tests related to original claim
rails test test/models/project_media_7_test.rb
rails test test/controllers/graphql_controller_11_test.rb
rails test test/controllers/graphql_controller_12_test.rb
```

## Checklist


vasconsaurus:
Some notes on the current state of things

- I tried to merge the logic but there are slight differences, that I'm not too sure on what the best way to handle is,
    - so there is still some pretty obvious duplication,
    - which I thought was best to leave for now: I think it's easier to see the differences/similarities, so it should be easier for us to discuss possible solutions?
- `rails test test/models/concerns/project_media_creators_test.rb` is broken, as it should be, since we no longer have the `create_link` method inside `project_media_creators`. Not sure what the best course of action regarding this one is.
- I don't like how I'm sending the data from ProjectMedia to Media? 
    - We need the claim/link/file, for different types of Media we might need different data from ProjectMedia, we need to keep track if it is an original_claim, currently we are not setting the user.. I think there should be a better way to access this? - I kept only `create_media` and removed `create_original_claim` from `ProjectMediaCreators` which seems to make more conceptual sense to me, since in the end it is just creating a media.

I think this is the bulk of it, though for sure I'm forgetting something.

app/models/media.rb
```diff
@@ -77,25 +79,36 @@ def domain
...
+      case media_type
+      when 'UploadedImage', 'UploadedVideo', 'UploadedAudio'
+        m = find_or_create_uploaded_file_media(project_media.file, media_type)
+      when 'Claim'
+        m = find_or_create_claim_media(project_media.quote, project_media.quote_attributions)
+      when 'Link'
+        m = find_or_create_link_media(project_media.url, team)
+      when 'Blank'
+        m = Blank.create!
+      end
```
caiosba, line 109: Let's try to resolve this duplication by trying to send the media information as an argument. Suggested signature:

```ruby
find_or_create_media_from_content(type, content, additional_args = [])
```

caiosba:
@melsawy @jayjay-w please review this as well.

app/models/media.rb
```diff
@@ -152,13 +134,53 @@ def self.download_file(url, ext)
     file
   end
 
-  def self.find_or_create_claim_media_from_original_claim(text)
-    Claim.find_by(original_claim_hash: Digest::MD5.hexdigest(text)) || Claim.create!(quote: text, original_claim: text)
+  # we set it to UploadedImage by default, should we?
+  # def self.create_with_file(project_media, media_type = 'UploadedImage')
```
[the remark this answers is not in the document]
vasconsaurus: Yes! But should we set `media_type = 'UploadedImage'` by default?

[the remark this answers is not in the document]
caiosba: No, I don't think it makes sense to have any default.

app/models/concerns/project_association.rb
```diff
@@ -158,8 +158,8 @@ def is_unique
 
     def set_media
       unless self.url.blank? && self.quote.blank? && self.file.blank? && self.media_type != 'Blank'
-        m = self.create_media
-        self.media_id = m.id unless m.nil?
+        self.create_media!
+        self.media_id unless self.media_id.nil?
```
melsawy, line 162: Q: this line should return `self.media_id` or assign a value to `self.media_id`

melsawy (approved):

app/models/concerns/project_association.rb
vasconsaurus: The `create_media` method changes the object and assigns the value, if I'm not mistaken. So I think we should be able to just return `self.media_id`.

PR #2224, opened 2025-02-18
"All Articles" list

caiosba:
## Description

We need to add an "All Articles" list to make it easier for the users to see the entire collection of both Claim & Fact-Checks as well as Explainers under one list.

Checklist:

- [x] There is already an `articles` field under `TeamType`: We need to make sure that the `article_type` argument is not mandatory anymore.
- [x] Implement a `filtered_articles` method for the `Team` model, that uses SQL `UNION` and reuse the logic for explainers and fact-checks implemented by methods `filtered_explainers` and `filtered_fact_checks`, respectively.
  - [x] Guard against SQL injections.
  - [x] Test for  all filters.
  - [x] Test for all sorting options.
  - [x] Avoid N + 1 queries problem.

Reference: CV2-5007.

## How to test?

Automated test was added. But we need to make sure that, for the "all articles" list:

- [x] All fact-checks/explainers filters work
- [x] All sorting options work
- [x] Combined with the two above, pagination works

## Checklist

### DIFF
--- a/app/models/team.rb
+++ b/app/models/team.rb
@@ -577,9 +577,9 @@ class Team < ApplicationRecord
   def filter_by_keywords(query, filters, type = 'FactCheck')
     tsquery = Team.sanitize_sql_array(["websearch_to_tsquery(?)", filters[:text]])
     if type == 'FactCheck'
-      tsvector = "to_tsvector('simple', coalesce(title, '') || ' ' || coalesce(summary, '') || ' ' || coalesce(url, '') || ' ' || coalesce(claim_descriptions.description, '') || ' ' || coalesce(claim_descriptions.context, ''))"
-    else
-      tsvector = "to_tsvector('simple', coalesce(title, '') || ' ' || coalesce(description, '') || ' ' || coalesce(url, ''))"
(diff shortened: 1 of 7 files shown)


app/models/team.rb
```diff
@@ -487,6 +487,23 @@ def available_newsletter_header_types
     available
   end
 
+  def filtered_articles(filters = {}, limit = 10, offset = 0, order = 'created_at', order_type = 'DESC')
+    columns = [:id, :title, :language, :created_at, :updated_at]
```
caiosba, line 491: We should be able to get only id and type for the sub-queries.

test/models/team_2_test.rb
```diff
@@ -1590,4 +1590,18 @@ def setup
     t = create_team
     assert_equal [], t.search_for_similar_articles('Test')
   end
+
+  # FIXME: Test with all possible filters
```
caiosba, line 1594: Need to test also that objects of different types but same ID are returned.

app/models/team.rb
caiosba: Done!

caiosba: We need all columns for sorting.

melsawy (approved):
Great work Dr. Caio

jayjay-w (approved):

PR #2230, opened 2025-02-20
Adding a `statistics_platforms` field to GraphQL `TeamType`.

caiosba:
## Description

The field returns the platforms for which analytics are available. The logic is: there is at least one media request for that platform.

Reference: CV2-5821.

## How to test?

I implemented an automated test for that.

## Checklist

### DIFF
--- a/app/models/team.rb
+++ b/app/models/team.rb
@@ -635,6 +635,11 @@ class Team < ApplicationRecord
     data
   end
 
+  # Platforms for which statistics are available (e.g., at least one media request)
+  def statistics_platforms
+    TiplineRequest.joins(:project_media).where('project_medias.team_id' => self.id).group('platform').count.keys
+  end
+
   # private
   #
   # Please add private methods to app/models/concerns/team_private.rb
--- a/test/models/team_2_test.rb
+++ b/test/models/team_2_test.rb
@@ -1597,6 +1597,7 @@ class Team2Test < ActiveSupport::TestCase
     assert_equal [], t.statistics_platforms
     create_tipline_request team_id: t.id, platform: 'telegram', associated: pm
     create_tipline_request team_id: t.id, platform: 'whatsapp', associated: pm
+    create_tipline_request team_id: t.id, platform: 'whatsapp', associated: pm
     assert_equal ['telegram', 'whatsapp'], t.reload.statistics_platforms.sort
   end
 end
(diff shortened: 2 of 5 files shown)


test/models/team_2_test.rb
```diff
@@ -1590,4 +1590,13 @@ def setup
     t = create_team
     assert_equal [], t.search_for_similar_articles('Test')
   end
+
+  test "should return platforms for which statistics are available" do
+    t = create_team
+    pm = create_project_media team: t
+    assert_equal [], t.statistics_platforms
+    create_tipline_request team_id: t.id, platform: 'telegram', associated: pm
+    create_tipline_request team_id: t.id, platform: 'whatsapp', associated: pm
```
melsawy, line 1599: I suggest adding another TiplineRequest with the same platform so we can verify the group by platform.

melsawy (approved):
LGTM, just add a comment for unit test.

test/models/team_2_test.rb
caiosba: Thanks Sawy, good point - done!

app/models/team.rb
```diff
@@ -635,6 +635,11 @@ def get_dashboard_exported_data(filters, dashboard_type)
     data
   end
 
+  # Platforms for which statistics are available (e.g., at least one media request)
+  def statistics_platforms
+    TiplineRequest.joins(:project_media).where('project_medias.team_id' => self.id).group('platform').count.keys
```
vasconsaurus, line 640: just out of curiosity: Any reason not to use `.pluck(:platform).uniq` instead of `.group('platform').count.keys` ?

caiosba: Using `.pluck(:platform).uniq` would load all platforms into Ruby memory and calls `.uniq`, which removes duplicates at the Ruby level... it wouldn't have good performance and could even trigger a OOM killer event.

But now I'm realizing that the best alternative (performance-wise) is to use `.distinct.pluck(:platform)`, which filters unique values at the database value... let me make that change.

vasconsaurus: > it wouldn't have good performance and could even trigger a OOM killer event.
aaaaaa! That makes sense! Thank you 🙏

caiosba: Thank you! The improvement is now in this other PR: https://github.com/meedan/check-api/pull/2232

PR #2233, opened 2025-02-24
Fixed cache clearing for bulk-accepted suggestions.

caiosba:
## Description

Previously, the id was incorrectly used as target_ids. Also, now, all relevant fields are included, not just a subset.

Fixes: CV2-6191.

## How to test?

I implemented a unit test that reproduces the issue.

## Checklist

### DIFF
--- a/app/models/concerns/relationship_bulk.rb
+++ b/app/models/concerns/relationship_bulk.rb
@@ -20,7 +20,6 @@ module RelationshipBulk
         end
         relationships = Relationship.where(id: ids, source_id: source_id)
         relationships.update_all(update_columns)
-        delete_cached_field(pm_source.id, ids)
         # Run callbacks in background
         extra_options = {
           team_id: team&.id,
@@ -29,6 +28,7 @@ module RelationshipBulk
           action: updates[:action]
         }
         self.delay.run_update_callbacks(ids.to_json, extra_options.to_json)
+        delete_cached_fields(pm_source.id, relationships.map(&:target_id))
         { source_project_media: pm_source }
       end
     end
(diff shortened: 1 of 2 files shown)


jayjay-w (approved):

app/models/concerns/relationship_bulk.rb
```diff
@@ -20,7 +20,6 @@ def bulk_update(ids, updates, team)
         end
         relationships = Relationship.where(id: ids, source_id: source_id)
         relationships.update_all(update_columns)
-        delete_cached_field(pm_source.id, ids)
```
vasconsaurus, line 23: @caiosba: before we were passing the relationship ids instead of the target project medias id, but we want to `delete_cached_field` for source and target project medias, correct?

Could we go through this file in our 1:1? I have other questions, like: why did you change the `delete_cached_field` method, and sort of related, why on like 31 we pass `pm_source.id` instead of `source_id`.

caiosba: Let's go through it in our 1:1, but I'm going to answer here for documentation as well.

There was a bug: We wanted to clear the cache for all project media IDs (source and targets, a.k.a. parent and children), but by mistake, the relationship IDs were being passed.

There is no difference between `source_id` and `pm.source_id`, it was only Sawy's preference :)

I took the opportunity to simplify, improve and refactor some things:
- Rename the method, since we're clearing multiple cached fields instead of only one
- Use the built-in method to clear all cached fields instead of manually clearing a hard-coded list of fields, because the previous implementation had knowledge it shouldn't have about how the cached fields work internally (like, the cache key format and which fields are affected by relationship). Also, it would miss it if we added a new field (which happened here - it was missing the `media_cluster_origin` field, that was added after this implementation

PR #2234, opened 2025-02-25
CV2-6141: Update options for existing relationship

melsawy:
## Description

After debugging various cases on Live, I discovered that metadata already exists, but the target is already part of another source. Therefore, based on our [method for creating relationships](https://github.com/meedan/check-api/blob/develop/app/models/relationship.rb#L161-L162), it returns the existing one without setting metadata options.

- [x] Add relationship ID to the logged message.
- [x] Update options for existing relationship.

References: CV2-6141

## How to test?

Re-run automated tests

## Checklist

### DIFF
--- a/app/models/relationship.rb
+++ b/app/models/relationship.rb
@@ -159,6 +159,13 @@ class Relationship < ApplicationRecord
   def self.create_unless_exists(source_id, target_id, relationship_type, options = {})
     # Verify that the target is not part of another source; in this case, we should return the existing relationship.
     r = Relationship.where(target_id: target_id).where.not(source_id: source_id).last
+    # should update options if exists
+    unless r.nil? || !options.blank?
+      options.each do |key, value|
+        r.send("#{key}=", value) if r.respond_to?("#{key}=")
+      end
+      r.save!
+    end
     return r unless r.nil?
     # otherwise we should get the existing relationship based on source, target and type
     r = Relationship.where(source_id: source_id, target_id: target_id).where('relationship_type = ?', relationship_type.to_yaml).last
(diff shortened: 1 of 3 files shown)


app/models/relationship.rb
```diff
@@ -159,6 +159,13 @@ def create_or_update_parent_id
   def self.create_unless_exists(source_id, target_id, relationship_type, options = {})
     # Verify that the target is not part of another source; in this case, we should return the existing relationship.
     r = Relationship.where(target_id: target_id).where.not(source_id: source_id).last
+    # should update options if exists
+    unless r.nil? || !options.blank?
+      options.each do |key, value|
+        r.send("#{key}=", value) if r.respond_to?("#{key}=")
+      end
+      r.save!
+    end
```
caiosba, line 168: Sawy, shouldn't we update the options only if the current values for the `Relationship` object are blank?

caiosba: Hum, I see, good point. But we should at least make sure that we're not _erasing_ previous values. So I suggest you check if at least one of the `value`'s from `options` is not blank.

melsawy: Done and check against model field

app/models/relationship.rb
```diff
@@ -159,6 +159,13 @@ def create_or_update_parent_id
   def self.create_unless_exists(source_id, target_id, relationship_type, options = {})
     # Verify that the target is not part of another source; in this case, we should return the existing relationship.
     r = Relationship.where(target_id: target_id).where.not(source_id: source_id).last
+    # should update options if exists and at least on field is blank
+    if !r.nil? && !options.blank? && r.model.blank?
```
caiosba, line 163: Sawy, as we discussed in the meeting, please make sure that `options` contains at least one non-blank value.

caiosba (changes requested):

caiosba (approved):
Thanks Sawy for addressing the comments from the previous reviews!

jayjay-w (approved):

PR #2238, opened 2025-02-26
CV2-6041: update suggested relation if new one is confirmed

melsawy:
## Description

There is a case for `create_unless_exists` that attempts to establish a confirmed relationship and there is a suggested relationship for the same source and target, and also go through [this case](https://github.com/meedan/check-api/blob/develop/app/models/relationship.rb#L168-L173).

so it same relationship exists but suggested then I'll update existing one to be confirmed.

References: CV2-6041

## How to test?

Implemented unit test

## Checklist


caiosba (approved):
Good catch, Sawy!

PR #2243, opened 2025-03-10
(5995) Implement timeout and fallback for `original_claim` URL processing

vasconsaurus:
### Context
The current implementation for processing original_claim URLs in Check API does not account for situations where external URLs take an extended time to process. This can result in delays or failures if processing exceeds the gateway timeout (currently set at 60 seconds).

This mostly happens when we try to create an `UploadedFile` `Media` from an `original_claim`. So we see this a lot in our zapier integrations, where we might get `app took too long too respond` response. This happens somewhat frequently, and could obscure real issues.

### How
- We configured a Timeout (shorter than 60s), 
- if that is exceeded, we create a `ClaimMedia` instead of `Audio/Image/VideoMedia`.

### Notes
**Rethink how we `find_or_create_uploaded_file_media`**
I realized we would sometimes send the file, and sometimes send the url. Then the method would need to figure out what to do. With that implementation, we would need to add a `begin/rescue` to this method, so it could then call `find_or_create_claim_media` if the download failed.

1. I think we should always send the same thing to `find_or_create_uploaded_file_media`, so we should always send a file.
2. If `ProjectMedia` is downloading the file, it can decide from the start where to send it, instead of sending it to one method and that redirecting to another one.
3. So `ProjectMedia` tries to download the file, and decides which method to call, 
4. `find_or_create_uploaded_file_media` and `find_or_create_claim_media` only worry about creating the Media

**To be added in a future PR**
We added this fallback to the creation of files, but we probably should add it to the creation of links as well.

References: CV2-5995

## How to test?

## Checklist

### DIFF
--- a/test/models/project_media_7_test.rb
+++ b/test/models/project_media_7_test.rb
@@ -268,6 +268,7 @@ class ProjectMedia7Test < ActiveSupport::TestCase
   end
 
   test "should create a Claim media instead of UploadedAudio media if a Timeout is raised when creating from original claim" do
+  # this error is expected when trying to dowload the file
     Tempfile.create(['test_audio', '.mp3']) do |file|
       file.write(File.read(File.join(Rails.root, 'test', 'data', 'rails.mp3')))
       file.rewind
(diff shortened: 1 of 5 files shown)


caiosba (approved):
Very good, Manu... Nice and clean.

test/models/project_media_7_test.rb
```diff
@@ -266,4 +266,37 @@ def setup
       assert_equal audio.id, pm.media.id
     end
   end
+
+  test "should create a Claim media instead of UploadedAudio media if a Timeout is raised when creating from original claim" do
+  # this error is expected when trying to dowload the file
```
caiosba, line 271: Super minor, but please just fix this indentation before merging.

test/models/project_media_7_test.rb
```diff
@@ -266,4 +266,37 @@ def setup
...
+      pm = create_project_media(set_original_claim: audio_url)
+
+      assert_equal 'Claim', pm.media.type
+      assert_equal audio_url, pm.media.quote
+      assert_not_empty pm.media.original_claim
+    end
+  end
+
+  test "should create a Claim media instead of Link media if a Timeout is raised" do
+  # this error is expected when Pender tries to parse the link, so during media creation
```
caiosba, line 289: Super minor, but please just fix this indentation before merging.

caiosba (approved):

PR #2250, opened 2025-03-12
4916 – Change 'listed' attribute so we do not display Fetch

vasconsaurus:
## Description

Fetch is deprecated and should not be shown in the workspace. We have already removed Fetch specific code from Check-Web. But it still renders the generic parts, if the `api` sends it.

We have a boolean attribute called `listed`. Which was implemented [here](https://github.com/meedan/check-api/commit/d0cd2747bc035aebbd33a9bae3bf60fc5ebe15b8). `Listed` bots are the ones that are displayed to users so they can install in their workspaces.

Since we don't want to show Fetch we can update it's `listed` attribute to "false", so it won't be displayed anymore. Note that Fetch `listed` attribute is already set to `false` in check live. We want it to be set t false regardless of environment.

References: CV2-4916

## How to test?

## Checklist

### DIFF
--- a/db/migrate/20250312190444_set_listed_to_false_for_fetch_bot.rb
+++ b/db/migrate/20250312190444_set_listed_to_false_for_fetch_bot.rb
@@ -1,7 +1,7 @@
 class SetListedToFalseForFetchBot < ActiveRecord::Migration[6.1]
   def change
     # Fetch should not be marked as "listed", so it does not appear on the UI
-    BotUser.where(login: ['fetch']).each do |bot|
+    BotUser.where(login: 'fetch').each do |bot|
       bot.set_listed = false
       bot.save!
     end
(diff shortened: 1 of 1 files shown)


db/migrate/20250312190444_set_listed_to_false_for_fetch_bot.rb
```diff
@@ -0,0 +1,9 @@
+class SetListedToFalseForFetchBot < ActiveRecord::Migration[6.1]
+  def change
+    # Fetch should not be marked as "listed", so it does not appear on the UI
+    BotUser.where(login: ['fetch']).each do |bot|
```
caiosba, line 4: It didn't need to be an array, could be just `login: 'fetch'`.

caiosba (approved):

db/migrate/20250312190444_set_listed_to_false_for_fetch_bot.rb
vasconsaurus: Very true. I just removed the old items, and didn't update accordingly.

melsawy (approved):

PR #2227, opened 2025-02-19
DRAFT: Store, index and search by `channel` for articles

jayjay-w:
## Description

We are adding the ability to store, index and search by `channel` for articles

References: CV2-6039

## How to test?

Please describe how to test the changes (manually and/or automatically).

## Checklist

- [ ] I have performed a self-review of my code and ensured that it is safe and runnable, that code coverage has not decreased, and that there are no new Code Climate issues. I have also followed [Meedan's internal coding guidelines](https://meedan.atlassian.net/wiki/spaces/ENG/pages/1309605889/Coding+guidelines).

PR #2255, opened 2025-03-20
6039 – Store, index search by `channel` for Articles

vasconsaurus:
## Description

**Disclaimer** Continuing Jay's work from PR #2227.

We need to store and index the channel of creation for articles to allow filtering by `channel` in articles lists:
 - imported: Imported from CSV. Should be set by our importing script.
 - manual: Created from the UI. Should be the default value.
 - api: Should be the default value when the creator is a bot user.
 - zapier: Created from Zapier. Should be set by the API client (Zapier).

We are:
- adding the ability to store, index and search by channel for articles,
- defaulting channel to manual,
- creating a rake task to update migration rake task to set the channel value for existing articles.

References: CV2-6039

## How to test?

```ruby
# graphql_controller_12_test

# test "should return fact_check channel in team articles query when filtered by channel"
rails test test/controllers/graphql_controller_12_test.rb:933

# test "should create explainer with default channel via GraphQL mutation"
rails test test/controllers/graphql_controller_12_test.rb:969

# test "should create fact_check with explicit channel via GraphQL mutation"
rails test test/controllers/graphql_controller_12_test.rb:994
```

```ruby
# explainer_test

# test "should assign default channel 'manual' for a regular user"
rails test test/models/explainer_test.rb:180

# test "should assign default channel 'api' for a BotUser"
rails test test/models/explainer_test.rb:186

# test "should allow explicit override of channel"
rails test test/models/explainer_test.rb:193

# test "should not allow an invalid channel value"
rails test test/models/explainer_test.rb:199

# test "should not change channel on update if already set"
rails test test/models/explainer_test.rb:206
```

```ruby
# team_test.rb

# test "should filter explainers by channel"
rails test test/models/team_test.rb:1340

# test "should filter fact_checks by channel"
rails test test/models/team_test.rb:1351

# test "should count articles by channel"
rails test test/models/team_test.rb:1366
```

```ruby
# fact_check_test.rb

# test "should assign default channel 'manual' for a regular user"
rails test test/models/fact_check_test.rb:745

# test "should assign default channel 'api' for a BotUser"
rails test test/models/fact_check_test.rb:750

# test "should allow explicit override of channel"
rails test test/models/fact_check_test.rb:757

# test "should not allow an invalid channel value"
rails test test/models/fact_check_test.rb:763

# test "should not change channel on update if already set"
rails test test/models/fact_check_test.rb:769
```

## Checklist


vasconsaurus:
@melsawy @caiosba I made a few updates. I'm adding some notes here, but there are 'notes' in the commits as well. Can you please take a look, and let me know what you think?

**1. Tests for fact-checks**
I had some issues writing a test to confirm channel is set to `api` when created by a Bot.
Specially if we rely only on `User.current`. I'm not sure if the issue is with how we set `User.current`, or with the way we create sample data, or something else entirely. But I hit a lot of permissions issues.

I left some notes in the test file and in the commit.

**2. Removed logic to set channel from create_claim_description_and_fact_check method**
This is a project_media_creators method, that creates a fact-check from an existing fact-check. So it only needs to get the channel from fact-check, not add it.

`before_validation :set_channel, on: :create` is the only place we should set the channel, I think.

**3. Do not set_channel for ClaimDescription**
Self-explanatory, `ClaimDescription` does not have the channel attribute, so I updated accordingly.

**4. Should we rely only on `User.current`?**
Something I asked myself while writing the tests for fact-check, so I wanted to ask you.
For now, I added `self.user.is_a?(BotUser)`

caiosba:
Thanks, @vasconsaurus. 2 and 3 make sense... for 1 and 4, in order to make sure we don't make the wrong assumptions, can you please tag us in the exact lines?

app/models/concerns/article.rb
```diff
@@ -31,6 +35,15 @@ def set_user
     self.user = User.current unless User.current.nil?
   end
 
+  def set_channel
+    return if self.channel.present?
+    if self.user.is_a?(BotUser) || User.current && User.current.is_a?(BotUser)
```
melsawy, line 40: Should be `self.user.is_a?(BotUser) || (User.current && User.current.is_a?(BotUser))`
as we should `||` with `User.current && User.current is bot`

test/models/explainer_test.rb
```diff
@@ -176,4 +176,58 @@ def teardown
...
+
+  test "should assign default channel 'api' for a BotUser" do
+    bot = create_bot_user(default: true, approved: true)
+    # Prevent duplicate installation errors by stubbing out the default bot callback if needed.
+    Team.any_instance.stubs(:add_default_bots_to_team)
+    create_team_user(team: @t, user: bot, role: 'admin')
+    User.current = bot
+    ex = create_explainer(team: @t, channel: nil)
+    assert_equal "api", ex.channel
+    User.current = nil
```
melsawy, line 199: Also you can use ```
with_current_user_and_team(bot, @t) do
  ex = create_explainer(team: @t, channel: nil)
 assert_equal "api", ex.channel
end

test/models/fact_check_test.rb
```diff
@@ -741,4 +741,65 @@ def setup
...
     assert_equal u, fc.author
   end
+
+  test "should assign default channel 'manual' for a regular user" do
+    fc = create_fact_check
+    assert_equal 'manual', fc.channel
+  end
+
+  # this was the only way so far I got this test to work so we could set User.current
+  test "should assign default channel 'api' for a BotUser" do
```
vasconsaurus, line 751: @caiosba test using `User.current`.

test/models/fact_check_test.rb
```diff
@@ -741,4 +741,65 @@ def setup
...
+    fc = create_fact_check(claim_description: cd, channel: nil, skip_check_ability: true)
+
+    assert_equal 'api', fc.channel
+
+    User.current = nil
+  end
+
+  # I feel something like this should work
+  # But if we only rely on User.current this does not work
+  test "should assign default channel 'api' for a BotUser #2" do
```
vasconsaurus, line 772: @caiosba for this test to work we need to check `self.user` in set_channel. Will tag you there as well.

app/models/concerns/article.rb
```diff
@@ -37,7 +37,7 @@ def set_user
 
   def set_channel
     return if self.channel.present?
-    if User.current && User.current.is_a?(BotUser)
+    if self.user.is_a?(BotUser) || User.current && User.current.is_a?(BotUser)
```
vasconsaurus, line 40: @caiosba here we were only checking `User.current`. Should we also check `self.user`?

caiosba: I think we should only check for `self.user`. Rationale: We already have a `set_user` callback before this one that sets the user based on `User.current`.

test/models/fact_check_test.rb
caiosba: Another reason for the `set_channel` callback to rely only on `self.user` - the test can also only rely on `self.user`.

lib/tasks/migrate/20250326140624_add_channel_to_articles.rake
```diff
@@ -0,0 +1,30 @@
+# bundle exec rake check:migrate:add_channel_to_articles
```
vasconsaurus, line 1: @melsawy if/when you have some time, can you please take a look at this file and let me know what you think?

melsawy: LGTM. I tested it against a live dump with `345609` FactCheck and it finished in `0m`. I also checked the live data, and it has `639838` articles, so I believe we are on the safe side.
great work @vasconsaurus

app/models/concerns/article.rb
```diff
@@ -31,6 +35,15 @@ def set_user
     self.user = User.current unless User.current.nil?
   end
 
+  def set_channel
+    return if self.channel.present?
+    if self.user.is_a?(BotUser)
+      self.channel = "api"
+    else
+      self.channel = "manual"
+    end
```
melsawy, line 44: recommended to use conditional operator
i.e self.channel = self.user.is_a?(BotUser) ?  'api'  :  'manual'

melsawy (approved):
Great work @vasconsaurus 
I just add a small comment for conditional operator.

test/controllers/graphql_controller_12_test.rb
```diff
@@ -908,25 +908,93 @@ def teardown
     assert_equal 'This item already exists', JSON.parse(@response.body).dig('errors', 0, 'message')
   end
 
-  test "should get annotations count" do
```
caiosba, line 911: Request: I don't think that removing this test was intended? I noticed that this PR uncovers line 19 of `app/graph/types/inclusions/task_and_annotation_fields.rb`.

app/models/concerns/article.rb
```diff
@@ -11,11 +13,11 @@ module Article
 
     before_validation :set_author, on: :create
     before_validation :set_user
-    before_validation :set_channel, on: :create, unless: -> { self.class_name == "ClaimDescription" }
+    before_validation :set_channel, on: :create, unless: -> { self.is_a?(ClaimDescription) }
     validates_presence_of :user
 
-    validates :channel, inclusion: { in: %w[imported manual api zapier] }, unless: -> { self.class_name == "ClaimDescription" }
-    enum channel: { imported: 0, manual: 1, api: 2, zapier: 3 }
+    validates :channel, inclusion: { in: %w[imported manual api zapier] }, unless: -> { self.is_a?(ClaimDescription) }
```
caiosba, line 19: @vasconsaurus please use `ARTICLE_CHANNELS.keys` instead of `%w[imported manual api zapier]` here.

caiosba (changes requested):
Thanks for addressing the comments, @vasconsaurus ! There is still one minor change - and regarding the other comment I added before: is the `seed` script already creating articles of different types?

caiosba (changes requested):
Thanks, Manu! Only missing the `seeds` file now.

caiosba (approved):
Thanks @vasconsaurus for going through all comments! :rocket:

PR #2260, opened 2025-03-23
CV2-6146: Add a message/conversation_id placeholder

melsawy:
## Description

Create a placeholder variable like {{message_id}} that can be included in a tipline message.

- [x] Implement a method called replace_placeholders in smooch_messages concern.
- [x] Call that method in the send_message_to_user implementations we have for the Zendesk provider and for the WhatsApp Cloud API provider:
- [x] Implement automated tests for the model method and for the two cases above

References: CV2-6146

## How to test?

Implement unit test.

## Checklist

### DIFF
--- a/app/models/concerns/smooch_capi.rb
+++ b/app/models/concerns/smooch_capi.rb
@@ -228,8 +228,8 @@ module SmoochCapi
       payload = {}
       account, to = uid.split(':')
       return if account != self.config['capi_phone_number']
-      text = self.replace_placeholders(uid, text)
       if text.is_a?(String)
+        text = self.replace_placeholders(uid, text)
         payload = {
           messaging_product: 'whatsapp',
           recipient_type: 'individual',
--- a/app/models/concerns/smooch_capi.rb
+++ b/app/models/concerns/smooch_capi.rb
@@ -228,7 +228,7 @@ module SmoochCapi
       payload = {}
       account, to = uid.split(':')
       return if account != self.config['capi_phone_number']
-      text = self.replace_placeholders(text, uid)
+      text = self.replace_placeholders(uid, text)
       if text.is_a?(String)
         payload = {
           messaging_product: 'whatsapp',
--- a/app/models/concerns/smooch_capi.rb
+++ b/app/models/concerns/smooch_capi.rb
@@ -228,6 +228,7 @@ module SmoochCapi
       payload = {}
       account, to = uid.split(':')
       return if account != self.config['capi_phone_number']
+      text = self.replace_placeholders(text, uid)
       if text.is_a?(String)
         payload = {
           messaging_product: 'whatsapp',
--- a/app/models/concerns/smooch_messages.rb
+++ b/app/models/concerns/smooch_messages.rb
@@ -570,7 +570,7 @@ module SmoochMessages
     end
 
     def replace_placeholders(uid, text)
-      external_id = TiplineMessage.where(uid: uid).last.external_id
+      external_id = TiplineMessage.where(uid: uid).last&.external_id
       text.gsub('{{message_id}}', external_id)
     end
   end
--- a/app/models/concerns/smooch_messages.rb
+++ b/app/models/concerns/smooch_messages.rb
@@ -568,5 +568,10 @@ module SmoochMessages
       # Define a min number of words to create a media
       CheckConfig.get('min_number_of_words_for_tipline_long_text') || CheckConfig.get('min_number_of_words_for_tipline_submit_shortcut', 10, :integer)
     end
+
+    def replace_placeholders(uid, text)
+      external_id = TiplineMessage.where(uid: uid).last.external_id
+      text.gsub('{{message_id}}', external_id)
+    end
   end
 end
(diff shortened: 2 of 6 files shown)


melsawy:
@caiosba can you please test against `CAPI` already I verified against `Zendesk` provider 
Screenshot from 2025-03-24 14-32-50

app/models/concerns/smooch_capi.rb
```diff
@@ -228,6 +228,7 @@ def capi_send_message_to_user(uid, text, extra = {}, _force = false, preview_url
       payload = {}
       account, to = uid.split(':')
       return if account != self.config['capi_phone_number']
+      text = self.replace_placeholders(uid, text)
```
caiosba, line 231: @melsawy , it's better to move this line to inside the condition below, no?

app/models/concerns/smooch_messages.rb
```diff
@@ -568,5 +568,10 @@ def min_number_of_words_for_tipline_long_text
       # Define a min number of words to create a media
       CheckConfig.get('min_number_of_words_for_tipline_long_text') || CheckConfig.get('min_number_of_words_for_tipline_submit_shortcut', 10, :integer)
     end
+
+    def replace_placeholders(uid, text)
+      external_id = TiplineMessage.where(uid: uid).last&.external_id
```
caiosba, line 573: For performance reasons, I suggest you run this query only if the message contains the placeholder.

caiosba, line 573: Just to be super safe, I suggest you filter by `team_id` as well.

test/models/bot/smooch_7_test.rb
```diff
@@ -660,4 +660,13 @@ def teardown
...
   end
+
+  test "should replace message_id placeholder" do
+    uid = random_string
+    tm = create_tipline_message uid: uid
+    text = 'Foo {{message_id}} bar {{message_id}}'
+    external_id = tm.external_id
+    output = Bot::Smooch.replace_placeholders(uid, text)
+    assert_equal "Foo #{external_id} bar #{external_id}", output
+  end
```
caiosba, line 671: For the test, I suggest you set an explicit `external_id` and assert it at the end, for example:

```ruby
tm = create_tipline_message uid: uid, external_id: 'abc123'
[...]
assert_equal "Foo abc123 bar abc123", output
```

app/models/concerns/smooch_messages.rb
caiosba, line 573: Another suggestion after testing locally: Please also filter by `direction: 'incoming'` and `state: 'received'`, this way we can consistently get "the last message sent by that user".

caiosba (changes requested):

caiosba (changes requested):
Thanks Sawy, it works well locally for CAPI. I left some comments to change a few things, and I'm happy to re-review after that. Another thing I noticed it's missing: let's make sure to shorten URLs in the message if the placeholder is found.

app/models/concerns/smooch_messages.rb
melsawy: Done

test/models/bot/smooch_7_test.rb
melsawy: Done

app/models/concerns/smooch_messages.rb
melsawy: Done and also I apply changes to allow multiple placeholders so in future it'll be easy to add a new placeholders.

melsawy:
@caiosba please re-test against CAPI and also I applied a new changes to allow multiple placeholders in future

caiosba (changes requested):
Sawy, please re-request review once the URL shortening case is fixed (and I can test it locally for CAPI as well).

melsawy:
> Thanks Sawy for addressing all code comments from the previous review! There is still one thing missing though: if URL shortening is enabled, the URL will be shortened before the placeholders are replaced, which in practice will "remove" the placeholder from the message. So, can you please also call the `replace_placeholders` method in the [URL shortening method?](https://github.com/meedan/check-api/blob/develop/lib/url_rewriter.rb#L25)
I could not call replace_placeholders in `shorten_and_utmize_urls` method as I depend on `uid` so I called it before calling `shorten_and_utmize_urls` method

caiosba:
> > Thanks Sawy for addressing all code comments from the previous review! There is still one thing missing though: if URL shortening is enabled, the URL will be shortened before the placeholders are replaced, which in practice will "remove" the placeholder from the message. So, can you please also call the `replace_placeholders` method in the [URL shortening method?](https://github.com/meedan/check-api/blob/develop/lib/url_rewriter.rb#L25)
> > I could not call replace_placeholders in `shorten_and_utmize_urls` method as I depend on `uid` so I called it before calling `shorten_and_utmize_urls` method

OK, that's actually better.

caiosba (approved):
@melsawy, I pushed a commit to fix it for messages that have menus and buttons. Works well for CAPI. Please test for Zendesk as well.

melsawy:
> @melsawy, I pushed a commit to fix it for messages that have menus and buttons. Works well for CAPI. Please test for Zendesk as well.

Works well for Zendesk thanks @caiosba

PR #2276, opened 2025-04-03
Include manual responses when calculating average response time

caiosba:
## Description

Include manual responses when calculating average response time.

- [x] Add columns first_manual_response_at and last_manual_response_at to the tipline_requests table, default null.
- [x] Fill in these columns every time a [manual response is sent](https://github.com/meedan/check-api/blob/develop/app/models/concerns/smooch_messages.rb#L555).
- [x] Update the average_response_time [method](https://github.com/meedan/check-api/blob/develop/lib/check_data_points.rb#L88) to take manual responses into account.
- [x] Update automated tests accordingly.

Reference: CV2-5996.

## How to test?

Automated tests implemented.

## Checklist

### DIFF
--- a/test/controllers/graphql_controller_10_test.rb
+++ b/test/controllers/graphql_controller_10_test.rb
@@ -763,7 +763,7 @@ class GraphqlController10Test < ActionController::TestCase
   end
 
   test "should send custom message to user" do
-    Bot::Smooch.stubs(:send_message_to_user).returns(OpenStruct.new(code: 200)).once
+    Bot::Smooch.stubs(:send_message_to_user).returns(OpenStruct.new(code: 200)).twice
     u = create_user
     t = create_team
     create_team_user user: u, team: t, role: 'editor'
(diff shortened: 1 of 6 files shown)


test/controllers/graphql_controller_10_test.rb
```diff
@@ -769,13 +769,17 @@ def setup
...
     authenticate_with_user(u)
 
     query = "mutation { sendTiplineMessage(input: { clientMutationId: \"1\", message: \"Hello\", inReplyToId: #{tr.id} }) { success } }"
     post :create, params: { query: query, team: t.slug }
 
     assert_response :success
     assert JSON.parse(@response.body)['data']['sendTiplineMessage']['success']
+    assert tr.reload.first_manual_response_at > 0
+    assert tr.reload.last_manual_response_at > 0
   end
```
melsawy, line 783: I recommend adding another test in which first_manual_response_at does not update, while last_manual_response_at should be updated with a new value.

melsawy (approved):
LGTM Caio just added a comment to add a new test case.

test/controllers/graphql_controller_10_test.rb
caiosba: Done!

PR #2278, opened 2025-04-08
CV2-6314: Sync `sources_count` value between PG & ES

melsawy:
## Description

Re-implement sources_count login using cached field approach
- [x] Add a cached field for sources_count and updated both PG & ES after update the cached value
- [x] sources_count field should updated after create/update/destroy confirmed relationships for both source & target
- [x] Delete targets_count field

References: CV2-6314

## How to test?

Re-run automated tests

## Checklist

### DIFF
--- a/app/models/concerns/relationship_bulk.rb
+++ b/app/models/concerns/relationship_bulk.rb
@@ -29,6 +29,9 @@ module RelationshipBulk
         }
         self.delay.run_update_callbacks(ids.to_json, extra_options.to_json)
         delete_cached_fields(pm_source.id, relationships.map(&:target_id))
+        # Call the cached field sources_count to update its value, as the field is not called from the UI.
+        ids = [pm_source.id, relationships.map(&:target_id)].flatten
+        ProjectMedia.where(id: ids).map(&:sources_count)
         { source_project_media: pm_source }
       end
     end
@@ -108,8 +111,6 @@ module RelationshipBulk
           begin Relationship.smooch_send_report(r.id) rescue nil end
         end
       end
-      # Update un-matched field
-      ProjectMedia.where(id: target_ids).update_all(unmatched: 0)
       # Update ES docs
       $repository.client.bulk body: es_body unless es_body.blank?
       # Import versions
@@ -140,7 +141,8 @@ module RelationshipBulk
           }
         }.to_json
       end
-
+      # Update un-matched field
+      ProjectMedia.where(id: target_ids).update_all(unmatched: 1)
       # Update ES
       options = {
         index: CheckElasticSearchModel.get_index_alias,
--- a/app/models/concerns/relationship_bulk.rb
+++ b/app/models/concerns/relationship_bulk.rb
@@ -128,9 +128,6 @@ module RelationshipBulk
       source = ProjectMedia.find_by_id(source_id)
       version_metadata = nil
       unless source.nil?
-        source.skip_check_ability = true
-        source.targets_count = Relationship.where(source_id: source.id).where('relationship_type = ? OR relationship_type = ?', Relationship.confirmed_type.to_yaml, Relationship.suggested_type.to_yaml).count
-        source.save!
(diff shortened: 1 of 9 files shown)


app/models/concerns/relationship_bulk.rb
```diff
@@ -66,7 +69,7 @@ def run_update_callbacks(ids_json, extra_options_json)
       index_alias = CheckElasticSearchModel.get_index_alias
       es_body = []
       versions = []
-      callbacks = [:reset_counters, :update_counters, :propagate_inversion]
+      callbacks = [:turn_off_unmatched_field, :move_explainers_to_source, :apply_status_to_target]
```
caiosba, line 72: We didn't call these callbacks before? Or, are they here to replace the lines you deleted from this method?

app/models/concerns/relationship_bulk.rb
```diff
@@ -106,11 +109,8 @@ def run_update_callbacks(ids_json, extra_options_json)
         # Send report if needed
         if extra_options['action'] == 'accept'
           begin Relationship.smooch_send_report(r.id) rescue nil end
-          Relationship.replicate_status_to_children(r.source_id, whodunnit, team_id)
```
caiosba, line 109: Why we don't need this line anymore? Is it because we call `apply_status_to_target`?

app/models/concerns/relationship_bulk.rb
```diff
@@ -29,6 +29,9 @@ def bulk_update(ids, updates, team)
         }
         self.delay.run_update_callbacks(ids.to_json, extra_options.to_json)
         delete_cached_fields(pm_source.id, relationships.map(&:target_id))
+        # Call the cached field sources_count to update its value, as the field is not called from the UI.
+        ids = [pm_source.id, relationships.map(&:target_id)].flatten
+        ProjectMedia.where(id: ids).map(&:sources_count)
```
caiosba, line 34: If you really want to update the value, shouldn't you call `sources_count(true)`?

app/models/concerns/relationship_bulk.rb
```diff
@@ -106,11 +109,8 @@ def run_update_callbacks(ids_json, extra_options_json)
         # Send report if needed
         if extra_options['action'] == 'accept'
           begin Relationship.smooch_send_report(r.id) rescue nil end
-          Relationship.replicate_status_to_children(r.source_id, whodunnit, team_id)
         end
       end
-      # Update un-matched field
-      ProjectMedia.where(id: target_ids).update_all(unmatched: 0)
```
caiosba, line 113: Why you deleted this line? Is it because now you call `turn_off_unmatched_field`?

caiosba (commented):
Sawy, it looks good to me, but some things are not clear/obvious, so I left some questions before approving. Next time, it helps if you add these to the PR description or as comments :)

And I have two more things `sources_count`... but I'm going to add them to the ticket.

app/models/concerns/relationship_bulk.rb
melsawy: Yes, Caio, I rely on callbacks rather than replicating status to all children each time.

app/models/concerns/relationship_bulk.rb
melsawy: Cached already cleared before this line so both are same.

app/models/concerns/relationship_bulk.rb
melsawy: Yes I rely on `turn_off_unmatched_field` callback

app/models/concerns/relationship_bulk.rb
melsawy: Yes, these callbacks replace the delete line, as I believe it's better to run callbacks for more consistency with other implementations.

caiosba (approved):
Thanks Sawy for going through all my comments. :shipit:

PR #2283, opened 2025-04-11
CV2-6314: remove unneeded code related to sources_count field

melsawy:
## Description

The cached field approach has already updated the value in ES, so I removed the other code to prevent hitting ES twice for the same field.

References: CV2-6314

## How to test?

## Checklist

### DIFF
--- a/app/models/concerns/project_association.rb
+++ b/app/models/concerns/project_association.rb
@@ -90,9 +90,7 @@ module ProjectAssociation
       ['team_id', 'user_id', 'read', 'source_id', 'project_id', 'unmatched'].each do |fname|
         data[fname] = self.send(fname).to_i if self.send("saved_change_to_#{fname}?")
       end
-      ['archived'].each do |fname|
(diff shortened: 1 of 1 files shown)


app/models/concerns/project_association.rb
```diff
@@ -90,7 +90,7 @@ def update_elasticsearch_data
       ['team_id', 'user_id', 'read', 'source_id', 'project_id', 'unmatched'].each do |fname|
         data[fname] = self.send(fname).to_i if self.send("saved_change_to_#{fname}?")
       end
-      ['archived', 'sources_count'].each do |fname|
+      ['archived'].each do |fname|
```
caiosba, line 93: Sawy, now it doesn't make sense to have an array and a loop with one item :sweat_smile:

caiosba (changes requested):

app/models/concerns/project_association.rb
melsawy: Fixed

caiosba (approved):

PR #2284, opened 2025-04-11
Fixing Code Climate issue and fixing flaky test.

caiosba:
## Description

Fixing a Code Climate issue (deleting a useless line) and fixing flaky test.

For the Code Climate issue, I deleted a line that was left by mistake as part of the `sources_count` refactoring (PR #2278).

The fix for the flaky test is to use a fixed mocked date, instead of relying on the current date, which could cause the flakyness we were seeing.

No Jira ticket for this.

## How to test?

I literally executed the flaky test locally 20 times after the fix.

## Checklist

### DIFF
--- a/app/models/relationship.rb
+++ b/app/models/relationship.rb
@@ -269,7 +269,6 @@ class Relationship < ApplicationRecord
 
   def propagate_inversion
     if self.source_id_before_last_save == self.target_id && self.target_id_before_last_save == self.source_id
-      ids = Relationship.where(source_id: self.target_id).map(&:id).join(',')
       report = Dynamic.where(annotation_type: 'report_design', annotated_type: 'ProjectMedia', annotated_id: self.source_id_before_last_save).last
       unless report.nil?
         report.annotated_id = self.source_id
--- a/test/lib/check_data_points_test.rb
+++ b/test/lib/check_data_points_test.rb
@@ -176,11 +176,13 @@ class CheckDataPointsTest < ActiveSupport::TestCase
   end
 
   test "should calculate average response time based also on custom manual messages" do
-    pm = create_project_media team: @team
-    tr = create_tipline_request team_id: @team.id, associated: pm
-    tr.update_columns(first_manual_response_at: (tr.created_at + 1.week).to_i, smooch_report_sent_at: (tr.created_at + 2.weeks).to_i)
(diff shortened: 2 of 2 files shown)


vasconsaurus:
@caiosba why did using the current date cause the flakiness?

caiosba:
> @caiosba why did using the current date cause the flakiness?

The test was flaky because it relied on `Time.now` implicitly, and that made the test result vary depending on when it was run. For example:

- `tr.created_at` was set using the actual current time (`Time.now`).
- The test calculated the average response time between those two timestamps.

But `CheckDataPoints.average_response_time` filters data based on a date range, and that range is dependent on `@start_date` and `@end_date`, which were defined relative to `Time.now`.

So, sometimes the test data fell outside the expected date range due to the system time when the test was run - leading to unexpected results.

The fix freezes time at 2025-02-01, so:

- All `created_at`, `first_manual_response_at`, etc., are deterministic.
- The test's data always falls within the hardcoded range '2025-01-01' to '2025-03-01'.

The response time calculation now consistently includes the test data.

Using `travel_to` ensures the environment inside the test behaves like it’s Feb 1, 2025 - so time-based logic, filtering, and averages behave consistently, no matter when the test is actually run.

caiosba:
> @caiosba why did using the current date cause the flakiness?

@vasconsaurus , the test was flaky because it relied on `Time.now` implicitly, and that made the test result vary depending on when it was run. For example:

- `tr.created_at` was set using the actual current time (`Time.now`).
- The test calculated the average response time between those two timestamps.

But `CheckDataPoints.average_response_time` filters data based on a date range, and that range is dependent on `@start_date` and `@end_date`, which were defined relative to `Time.now`.

So, sometimes the test data fell outside the expected date range due to the system time when the test was run - leading to unexpected results.

The fix freezes time at 2025-02-01, so:

- All `created_at`, `first_manual_response_at`, etc., are deterministic.
- The test's data always falls within the hardcoded range '2025-01-01' to '2025-03-01'.

The response time calculation now consistently includes the test data.

Using `travel_to` ensures the environment inside the test behaves like it’s Feb 1, 2025 - so time-based logic, filtering, and averages behave consistently, no matter when the test is actually run.

vasconsaurus (approved):

PR #2289, opened 2025-04-15
6297 – Webhook management API

vasconsaurus:
## Description

To reduce the need for engineering support and allow teams to self-configure their integrations, we need to generate the specific “CRUD” operations to support the webhook configuration UI.

### Implementation notes
1. Implemented a new GraphQL type `WebhookType`, that returns webhooks installed in a team
2. Implemented the mutations: `createWebhook`, `updateWebhook`, and `deleteWebhook`
3. Right now, `Webhook` is not a real class. It is an abstraction for a `BotUser` that has `events`, `request_url` and is not `default`.
4. However, as a follow-up to this work, we will:
    - Discuss simpler architecture for webhooks/bot users/API keys (from smaller changes to bigger refactorings)
    - Implement the type: 'webhook' to the `User` class
5. **To pay attention**: We want to make sure that `events` is not an empty array when validating. But it can be nil, and there were some conflicts with our core bots. Some extra attention there when reviewing the [logic](https://github.com/meedan/check-api/blob/8eef48e87843c260d800cae0fff8488c39dc2c50/app/models/bot_user.rb#L381-L397) might be called for. (I used this test to help validate this: `rails test test/models/team_2_test.rb:1627`)

References: CV2-5935, CV2-6297

## How to test?

```ruby
# test "should get only webhooks installed in a team"
rails test test/controllers/graphql_controller_2_test.rb:131

# test "should create a webhook, and return team with webhook's list"
rails test test/controllers/graphql_controller_2_test.rb:157

# test "should update a webhook, and return team with webhook's list"
rails test test/controllers/graphql_controller_2_test.rb:200

# test "should delete a webhook, whether the id format is Webhook/ID or BotUser/ID"
rails test test/controllers/graphql_controller_2_test.rb:252
```

## Checklist


app/models/ability.rb
```diff
@@ -67,7 +67,11 @@ def admin_perms
       obj.team_id == @context_team.id || obj.feed.team_id == @context_team.id
     end
     can [:create, :update, :read, :destroy], ApiKey, :team_id => @context_team.id
-    can :destroy, [Dynamic, DynamicAnnotation::Field], annotation_type: 'metadata' # FIXME: Restrict by team
+    # can :destroy, [Dynamic, DynamicAnnotation::Field], annotation_type: 'metadata' # FIXME: Restrict by team
+    can :destroy, [Dynamic, DynamicAnnotation::Field] do |obj|
+      source = Source.find(obj.id)
+      source.team_id == @context_team.id
```
caiosba: @vasconsaurus , here you're using the `Dynamic` or `DynamicAnnotation::Field` ID to load a `Source`, assuming that the IDs are the same, but they're not (actually even in a 1:1 relationship, this is fragile). I noticed that looks like both `Dynamic` and `DynamicAnnotation::Field` have a `team` method, so I think you can simply do:

```ruby
can :destroy, [Dynamic, DynamicAnnotation::Field] do |obj|
  obj.team.present? && obj.team == @context_team
end
```

caiosba (changes requested):

app/models/ability.rb
vasconsaurus: Updated!

caiosba:
I'm going to review the code, and @melsawy as well... @amoedoamorim , can you please make sure it works locally for the frontend code you're working on?

app/graph/types/webhook_type.rb
```diff
@@ -0,0 +1,23 @@
+class WebhookType < DefaultObject
+  description "Represents a Webhook, a BotUser with events and request url"
+
+  implements GraphQL::Types::Relay::Node
+
+  field :dbid, GraphQL::Types::Int, null: true
+  field :name, GraphQL::Types::String, null: true
+  field :events, GraphQL::Types::String, null: true
```
melsawy, line 8: I recommend defining this field as `JsonStringType`

app/graph/types/webhook_type.rb
```diff
@@ -0,0 +1,23 @@
...
+  field :request_url, GraphQL::Types::String, null: true
+  field :headers, GraphQL::Types::String, null: true
```
melsawy, line 10: I recommend defining this field as `JsonStringType`

melsawy (changes requested):
Great work @vasconsaurus 
I added comments regarding the use of `JsonStringType` for JSON fields.

app/graph/mutations/webhook_mutations.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  class Create < Mutations::CreateMutation
+    include SharedCreateAndUpdateFields
+  end
+
+  class Update < Mutations::UpdateMutation
+    include SharedCreateAndUpdateFields
+
+    argument :id, GraphQL::Types::String, required: false
```
caiosba, line 23: This field should be required.

app/graph/types/webhook_type.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  def events
+    object.get_events.to_json
+  end
```
caiosba, line 14: As Sawy said above for the `BotUser` model, if the fields are defined as `JsonStringType`, you don't need to call `.to_json` here.

app/graph/types/webhook_type.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  def request_url
+    object.get_request_url
+  end
+
+  def headers
+    object.get_headers.to_json
```
caiosba, line 21: As Sawy said above for the `BotUser` model, if the fields are defined as `JsonStringType`, you don't need to call `.to_json` here.

caiosba (changes requested):
Great stuff, Manu! I also left a couple of comments around the GraphQL API, but minor changes. Can you please also add unit tests for these missing lines?

```
app/graph/types/webhook_type.rb: 13, 17, 21
app/models/bot_user.rb: 384
```

amoedoamorim:
All 4 CRUD operations LGTM! :white_check_mark:

amoedoamorim:
Yeah, I did try the latest

caiosba (approved):
Thanks for addressing the comments, Manu! There was still one uncovered line, but I added a test for it already.

:shipit:

vasconsaurus:
> Thanks for addressing the comments, Manu! There was still one uncovered line, but I added a test for it already.
> 
> :shipit:

I had added this one to make sure we get an error if events is empty: https://github.com/meedan/check-api/blob/59882e284915dcd31f8951fec61887445c27f15b/test/models/bot_user_test.rb#L127

Was that not enough?

caiosba:
> > Thanks for addressing the comments, Manu! There was still one uncovered line, but I added a test for it already.
> > :shipit:
> 
> @caiosba, I had added this one to make sure we get an error if events is empty:
> 
> https://github.com/meedan/check-api/blob/59882e284915dcd31f8951fec61887445c27f15b/test/models/bot_user_test.rb#L127
> 
> Was that not enough?

@vasconsaurus , should be! And I confirmed locally that the supposedly uncovered line was reached by that test... I don't know why the coverage report didn't get it, but I saw it happening with other `BotUser` tests... it's less likely to happen when the new tests are added to `test/models/bot_user_2_test.rb` :shrug:

melsawy (approved):

PR #2298, opened 2025-05-04
Exposing `dbid` in GraphQL `ExplainerItemType` and including child items when sending explainers

caiosba:
## Description

- Exposing `dbid` in GraphQL `ExplainerItemType`.
- De-duplicating requests, so the same user doesn't receive the same explainer more than once.
- Including child items when sending explainers to previous requests.

Reference: CV2-6335.

## How to test?

Existing tests should cover this.

## Checklist

### DIFF
--- a/app/models/explainer_item.rb
+++ b/app/models/explainer_item.rb
@@ -17,7 +17,7 @@ class ExplainerItem < ApplicationRecord
   end
 
   def send_explainers_to_previous_requests(range)
-    ids = ProjectMedia.where(id: self.project_media.related_items_ids).map(&:id) # Including child items
+    ids = ProjectMedia.where(id: self.project_media.related_items_ids).pluck(:id) # Including child items
     # Keep track of UIDs so we don't send the same explainer to the same user more than once.
     # We could use GROUP BY or DISTINCT ON, but it would be more complex for the average number of requests we have.
     uids = []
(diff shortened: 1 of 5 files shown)


melsawy (approved):

caiosba:
@melsawy please re-review - I pushed a couple of changes to de-duplicate requests and to include requests from child items.

melsawy (approved):

caiosba:
@melsawy, I fixed the case (3) you reported in the Jira ticket, please review one more time!

app/models/explainer_item.rb
```diff
@@ -17,7 +17,14 @@ def version_metadata(_changes)
   end
 
   def send_explainers_to_previous_requests(range)
-    TiplineRequest.no_articles_sent(self.project_media_id).where(created_at: Time.now.ago(range.days)..Time.now).find_each do |tipline_request|
+    ids = ProjectMedia.where(id: self.project_media.related_items_ids).map(&:id) # Including child items
```
vasconsaurus, line 20: @caiosba, a question, why not use `ProjectMedia.where(id: self.related_items_ids).pluck(:id)` here?

caiosba: @vasconsaurus , yes, we could use `pluck`... it would be faster, actually! I'll make that change.

caiosba: Done (for this line and for a couple of other similar lines). Thanks Manu!

PR #2303, opened 2025-05-13
CV2-6363: Add periodic check for MonthlyTeamStatistic

melsawy:
## Description

Add Sentry notification for missing tipline statistics updates in last 24 hours

References: CV2-6363

## How to test?

Implemented automated tests

## Checklist

### DIFF
--- a/app/workers/tipline_statistics_periodic_check_worker.rb
+++ b/app/workers/tipline_statistics_periodic_check_worker.rb
@@ -1,0 +1,14 @@
+class TiplineStatisticsPeriodicCheckWorker
+  include Sidekiq::Worker
+
+  sidekiq_options retry: 0
+
+  def perform
+    if CheckConfig.get('ENABLE_TIPLINE_STATS_MONITORING', true)
+      periodic_check = Time.now.ago(1.day)
+      count = MonthlyTeamStatistic.where('created_at > ? OR updated_at > ?', periodic_check, periodic_check).count
+      error_msg = StandardError.new('No MonthlyTeamStatistic updated or created in the last 24 hours')
+      CheckSentry.notify(error_msg) if count == 0
+    end
+  end
+end
--- a/app/workers/tipline_statistics_periodic_check_worker.rb
+++ b/app/workers/tipline_statistics_periodic_check_worker.rb
@@ -4,10 +4,8 @@ class TiplineStatisticsPeriodicCheckWorker
   sidekiq_options retry: 0
 
   def perform
-    if CheckConfig.get('ENABLE_TIPLINE_STATS_MONITORING', true)
-      periodic_check = Time.now.ago(1.day)
-      count = MonthlyTeamStatistic.where('created_at > ? OR updated_at > ?', periodic_check, periodic_check).count
-      CheckSentry.notify(StandardError.new('No MonthlyTeamStatistic updated or created in the last 24 hours')) if count == 0
-    end
+    periodic_check = Time.now.ago(26.hours)
+    count = MonthlyTeamStatistic.where('created_at > ? OR updated_at > ?', periodic_check, periodic_check).count
+    CheckSentry.notify(StandardError.new('No MonthlyTeamStatistic updated or created in the last 24 hours')) if count == 0
   end
 end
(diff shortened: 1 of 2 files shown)


app/workers/tipline_statistics_periodic_check_worker.rb
```diff
@@ -0,0 +1,13 @@
+class TiplineStatisticsPeriodicCheckWorker
+  include Sidekiq::Worker
+
+  sidekiq_options retry: 0
+
+  def perform
+    if CheckConfig.get('ENABLE_TIPLINE_STATS_MONITORING', true)
+      periodic_check = Time.now.ago(1.day)
```
caiosba, line 8: I would add some buffer here just to be safe (the batch job takes a while to start, etc.)

app/workers/tipline_statistics_periodic_check_worker.rb
```diff
@@ -0,0 +1,13 @@
+class TiplineStatisticsPeriodicCheckWorker
+  include Sidekiq::Worker
+
+  sidekiq_options retry: 0
+
+  def perform
+    if CheckConfig.get('ENABLE_TIPLINE_STATS_MONITORING', true)
```
caiosba, line 7: Is this needed? This job won't be scheduled by default, it will execute only if it's explicitly created.

caiosba (commented):
Sawy, please add a comment to the worker just to explain how to schedule the job.

app/workers/tipline_statistics_periodic_check_worker.rb
melsawy: Removed

app/workers/tipline_statistics_periodic_check_worker.rb
melsawy: Done make it 26 hours

app/workers/tipline_statistics_periodic_check_worker.rb
```diff
@@ -4,10 +4,8 @@ class TiplineStatisticsPeriodicCheckWorker
   sidekiq_options retry: 0
 
   def perform
-    if CheckConfig.get('ENABLE_TIPLINE_STATS_MONITORING', true)
-      periodic_check = Time.now.ago(1.day)
-      count = MonthlyTeamStatistic.where('created_at > ? OR updated_at > ?', periodic_check, periodic_check).count
-      CheckSentry.notify(StandardError.new('No MonthlyTeamStatistic updated or created in the last 24 hours')) if count == 0
-    end
+    periodic_check = Time.now.ago(26.hours)
```
caiosba, line 7: I think even 25 would be good, but I suggest you add a comment here just to make clear that it was not a typo, but an intended buffer (something like "adding a buffer of one hour to our expected 24 hours interval", something like that.

caiosba (approved):
Thanks Sawy for addressing all comments from the PR review! I just left a note about a minor comment that I think would be useful to add.

test/workers/tipline_statistics_periodic_check_worker_test.rb
```diff
@@ -0,0 +1,25 @@
+require_relative '../test_helper'
+
+class TiplineStatisticsPeriodicCheckWorkerTest < ActiveSupport::TestCase
+  test "should notify Sentry for missing tipline statistics updates in last 24 hours" do
+    t = create_team
+    # Verify both created/updated dates
```
vasconsaurus, line 6: @caiosba and @melsawy just out of curiosity: is there any advantage in using travel_to instead of stubbing the time?

something like:

```ruby
  travel_to 3.days.ago do
    @ms = create_monthly_team_statistic team: t
  end

  CheckSentry.expects(:notify).once
  TiplineStatisticsPeriodicCheckWorker.new.perform
```

caiosba: Good point, @vasconsaurus ... I also prefer and use `travel_to`, for two main reasons: it's part of standard Rails, and works with any time library, while `Time.stubs(:now)` is only affecting specifically `Time.now` and it's part of `mocha`... @melsawy , I suggest you make this change.

app/workers/tipline_statistics_periodic_check_worker.rb
melsawy: Done

test/workers/tipline_statistics_periodic_check_worker_test.rb
melsawy: Done

vasconsaurus (approved):

PR #2305, opened 2025-05-13
CV2-3659: Remove Project & ProjectGroup

melsawy:
## Description

Remove Project & ProjectGroup models

- [x] Remove project.rb & project_groub.rb models
- [x] Remove graphql fields and mutations
- [x] Remove permissions related to deleted models
- [x] Remove `project_id` reference from ProjectMedia model
- [x] Remove related modes from Team model (rule actions, default folder)
- [x] Remove cached fields for Project
- [x] Remove Project bulk-action (move_to)
- [x] Fix tests and ensure test coverage is 100%
- [x] Fix seed script

References: CV2-3659

## How to test?

- Re-run automated test
- Test manually
- [x] Add new items
- [x] Update status
- [x] Assign/Un-assign
- [x] Assign/un-assign mail notifications
- [x] Create new list
- [x] Update existing list (title/filters)
- [x] Tipline requests
- [x] suggest/match items
- [x] Team rules (Verify conditions/actions )

## Checklist


app/graph/types/query_type.rb
```diff
@@ -115,18 +115,15 @@ def find_public_team(slug:)
   field :project_media,
         ProjectMediaType,
         description:
-          'Information about a project media, The argument should be given like this: "project_media_id,project_id,team_id"',
+          'Information about a project media, The argument should be given like this: "project_media_id,team_id"',
```
caiosba, line 118: @melsawy , this requires changes to Check Web and other GraphQL clients, no? I remember that in different places of Check Web and other Check API clients we used to call like `1,,2`, `1,null,2`, `1,0,2`, etc.

app/graph/types/query_type.rb
```diff
@@ -115,18 +115,15 @@ def find_public_team(slug:)
...
 
   def project_media(ids:)
-    objid, pid, tid = ids.split(",").map(&:to_i)
+    objid, tid = ids.split(",").map(&:to_i)
     tid = (Team.current.blank? && tid.nil?) ? 0 : (tid || Team.current.id)
-    project = Project.where(id: pid, team_id: tid).last
-    pid = project.nil? ? 0 : project.id
-    Project.current = project
-    objid = ProjectMedia.belonged_to_project(objid, pid, tid) || 0
+    objid = ProjectMedia.belonged_to_team(objid, tid) || 0
```
caiosba, line 126: @melsawy , this requires changes to Check Web and other GraphQL clients, no? I remember that in different places of Check Web and other Check API clients we used to call like `1,,2`, `1,null,2`, `1,0,2`, etc. As I suggested before, for changes like this, we may need to keep some temporary code in the GraphQL layer with some deprecation warning.

app/lib/check_permissions.rb
caiosba, line 1: @melsawy I suggest we also force a cache clear for the permissions object in this file:

```ruby
cache_key = "team_permissions_#{self.private.to_i}_#{role}_role_20250526174300"
```

app/graph/types/query_type.rb
caiosba: Ah, something like:

```ruby
all_ids = ids.split(",").map(&:to_i)
objid = all_ids.first
tid = all_ids.last
```

Right? Yeah, that's a good idea!

app/models/concerns/assignment_concern.rb
```diff
@@ -66,13 +66,9 @@ def project_media_assigned_to_user(user, select = nil)
       pids = []
```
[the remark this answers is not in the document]
melsawy: Done

caiosba (approved):
Thanks Sawy for the changes! Code looks good. I'm going to try it locally with a WhatsApp tipline and answer in the ticket. But let's merge this once when integration tests pass from Check Web.

caiosba (changes requested):
@melsawy , when testing it locally using Check Web, I noticed that clicking on an item from the "All Items" page leads to "page not found".

melsawy:
> @melsawy , when testing it locally using Check Web, I noticed that clicking on an item from the "All Items" page leads to "page not found".

Fixed this one related to latest changes for objid and tid

caiosba (approved):
Thanks Sawy, tipline submission on WhatsApp works fine now.

test/models/project_media_5_test.rb
```diff
@@ -84,9 +84,6 @@ def setup
   end
 
   test "should have a media not not necessarily a project" do
```
vasconsaurus, line 86: @melsawy, I think you should probably remove this whole test ("should have a media not not necessarily a project")

app/graph/types/query_type.rb
```diff
@@ -121,7 +121,10 @@ def find_public_team(slug:)
   end
 
   def project_media(ids:)
-    objid, tid = ids.split(",").map(&:to_i)
+    # Get the first and last id to handle the ids: in format '1,,2', '1,null,2' or '1,0,2'
```
vasconsaurus, line 124: Question: Do we always get the ids with one element that might be empty, 0 or null in between them? Or does the quantity of elements vary, e.g.: '1,2,null,3'? Why do we always want the first and the last?

melsawy: The format for IDs is `ProjectMediaId,ProjectId,TeamId`, with both `ProjectId` and `TeamId` expected to be `nil`. Since `ProjectId` was removed but the external GraphQL still uses the old format, we now only need `ProjectMediaId` and `TeamId`. Therefore, the behavior changes to retrieving the first and last values to ensure we get `ProjectMediaId` and `TeamId` in case GraphQL submits three arguments.

test/models/project_media_5_test.rb
melsawy: Removed this test

PR #2313, opened 2025-06-04
CV2-6395: Truncate title field to stay safely under maximum.

melsawy:
## Description

Ensure the field value does not exceed ~32KB  so truncate at 30_000 to stay safely under 32KB as the maximum size should be 32 * 1024 bytes.

References: CV2-6395

## How to test?

Manually create a new item with same [title here]( https://checkmedia.org/rumour-tracking-chequeabolivia-/media/3338473?tab=articles)

## Checklist

### DIFF
--- a/app/models/concerns/project_media_cached_fields.rb
+++ b/app/models/concerns/project_media_cached_fields.rb
@@ -639,7 +639,9 @@ module ProjectMediaCachedFields
       else
         self.update_column(:custom_title, title)
       end
-      title.to_s
+      # Ensure the field value does not exceed ~32KB (ref: CV2-6395);
+      # truncate at 30_000 to stay safely under 32KB as the maximum size should be 32 * 1024 bytes.
+      title.to_s.truncate(30_000)
     end
 
     def recalculate_status


app/models/concerns/project_media_cached_fields.rb
```diff
@@ -639,7 +639,9 @@ def recalculate_title
       else
         self.update_column(:custom_title, title)
       end
-      title.to_s
+      # Ensure the field value does not exceed ~32KB (ref: CV2-6395);
+      # truncate at 30_000 to stay safely under 32KB as the maximum size should be 32 * 1024 bytes.
+      title.to_s.truncate(30_000)
```
vasconsaurus, line 644: @melsawy I might be misunderstanding, but I tried to run this with the title example you added to the ticket and the original and truncated title were the same size:

```
> title.size
=> 17147
> title2 = title.to_s.truncate(30_000)
=> "https://vm.tiktok.com/ZMSrt46Xj/\\n\\ndata:image/jpeg;base64,/9j/4AAQSkZJRgABAQAAA..."
> title2.length
=> 17147
```

And a couple of questions: Don't we have a maximum character size for titles? If we do, couldn't we truncate using that as a reference?

caiosba (approved):

vasconsaurus (approved):

vasconsaurus:
> @vasconsaurus maybe you got something different... but the original one has 40.349 characters:
> 
> Captura de tela de 2025-06-04 09-07-04

It probably lost something along the way, because I tried copy/pasting and should have used `File.read`. Sorry for the confusion @melsawy

caiosba:
@vasconsaurus. regarding your other questions:

> Don't we have a maximum character size for titles?

Actually, no... but we should. The item title can be one of the following:

- The claim (which has a long limit, since it's long text)
- The fact-check title (which has a short limit)
- The original media title (which has a long limit)
- A custom title (which has a short limit)
- Etc.

Each of these has a different character limit. But the item title itself should have a defined limit - and in my opinion, that limit should be well below 30K... @melsawy, what do you think? Could you please file a refactoring ticket in the backlog for that?

> If we do, couldn't we truncate using that as a reference?

Absolutely (as per the point above).

PR #2323, opened 2025-06-26
3942 – Avoid unwanted TeamBotInstallations when duplicating workspaces

vasconsaurus:
## Description

When duplicating teams, we were copying the bots from the original team into the new team. But that was causing issues, e.g.:
- We create a team with a bot that has an API key to import items
- We duplicate the team, which duplicates the bot and API key
- Now we might think we are importing items to the original team, but we are actually importing them to the duplicated team

In order to avoid different teams with same API keys, instead of copying the original team's bots, we want to just install any default bots. To do that, we just need to run the `add_default_bots_to_team` `after_create` callback even if `is_being_copied`.

References: CV2-3942 , CV2-5876

## How to test?

```ruby
#   test "should not duplicate original team bots, should install default bots when duplicating a team"
rails test test/models/team_test.rb:903
```

Note: I replaced `setup_smooch_bot` with `create_bot_user`, since the former is slower and we are only testing the installation, and not the bot itself.

## Checklist

### DIFF
--- a/app/models/concerns/team_duplication.rb
+++ b/app/models/concerns/team_duplication.rb
@@ -23,7 +23,6 @@ module TeamDuplication
           @clones << { original: original, clone: copy }
           self.alter_copy_by_type(original, copy)
         }
-        self.process_team_bot_installations(t, team)
         team = self.modify_settings(t, team)
         team = self.update_team_rules(team)
         self.adjust_team_tasks(team)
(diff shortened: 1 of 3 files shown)


caiosba (approved):
Pair-reviewed with Manu in our 1:1.

vasconsaurus:
One test keeps failing, but I think it's unrelated:

```ruby
Failure:
ProjectMedia8Test#test_should_have_a_fallback_when_trying_to_create_a_blank_item_with_original_claim_but_a_timeout_error_happens [/app/test/models/project_media_8_test.rb:181]:
Expected: "UploadedImage"
  Actual: "Claim"

rails test test/models/project_media_8_test.rb:167
```

I have seen this one be flaky before, but testing locally, it has consistently failed each time I tried. But, if it is indeed unrelated, I think this is a separate ticket.

What do you think @caiosba? Should we make a ticket to look into it? Can I go ahead and merge this PR?

vasconsaurus:
> @vasconsaurus if you see the same behavior on `develop`, it confirms it's not related to this PR so it would be OK to merge.

Yup, ran the test from `develop` and got the same error.

PR #2326, opened 2025-07-01
6248 teambotinstallation permissions issue

vasconsaurus:
- **reproduce error**
- **add skip_check_ability**

## Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context, type of change (bug fix, performance, feature, etc.) and things to pay attention to during review.

References: TICKET-ID-1, TICKET-ID-2, …, TICKET-ID-N

## How to test?

## Checklist

### DIFF
--- a/app/models/bot_user.rb
+++ b/app/models/bot_user.rb
@@ -141,7 +141,9 @@ class BotUser < User
   end
 
   def install_to!(team)
-    TeamBotInstallation.create! team_id: team.id, user_id: self.id
+    team_bot_installation = TeamBotInstallation.new team_id: team.id, user_id: self.id
+    team_bot_installation.skip_check_ability = true
+    team_bot_installation.save!
   end
 
   def uninstall_from!(team)
--- a/test/models/team_test.rb
+++ b/test/models/team_test.rb
@@ -907,6 +907,8 @@ class TeamTest < ActiveSupport::TestCase
     team_bot = create_bot_user(name: 'Team Bot', login: 'team_bot', default: false, approved: true)
 
     original_team = create_team(name: 'original_team')
+    team_user = create_team_user team: original_team, role: 'admin'
+    User.current = team_user.user
 
     # Install team_bot manually, since it is not a default bot
     team_bot.install_to!(original_team)


vasconsaurus:
@caiosba using `team_bot_installation.skip_check_ability`, but why should we do that instead of updating `check_basic_abilities.rb`?

caiosba:
@vasconsaurus , I thought we were calling `TeamBotInstallation.new` straight from the `Team` `after_create` callback that installs the default bots... But since we're calling `install_to!`, it's better to either fix the permission, as you suggested, or replace `install_to!` by a manual new and save that skips the permission.

lib/check_basic_abilities.rb
```diff
@@ -20,6 +20,7 @@ def global_admin_perms
 
   def authenticated_perms
     can :create, Team
+    can [:create, :update], TeamBotInstallation
```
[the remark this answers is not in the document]
caiosba: Yes, that's too broad - it allows any authenticated user to create any `TeamBotInstallation`. We should restrict by role and team. Please add a new rule here under `admin_perms`: https://github.com/meedan/check-api/blob/5b14c8933b93c2e347c5941e450619baf95f8a66/app/models/ability.rb#L56. Something like: `    can :create, TeamBotInstallation, :team_id => @context_team.id`.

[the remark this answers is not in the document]
vasconsaurus: @caiosba, I tried that, but it only works if we don't check the context:

```
def admin_perms 
    can [:create, :update], TeamBotInstallation
```

Because the `@context_team` is the original_team, but the `team_bot_installation.team` is the duplicated team.

app/models/bot_user.rb
```diff
@@ -141,7 +141,7 @@ def subscribed_to?(event)
   end
 
   def install_to!(team)
-    TeamBotInstallation.create! team_id: team.id, user_id: self.id
+    TeamBotInstallation.create! team_id: team.id, user_id: self.id, skip_check_ability: team.is_being_copied
```
caiosba, line 144: Ah, that's a nice solution!

test/models/project_media_8_test.rb
```diff
@@ -165,6 +165,7 @@ def teardown
   end
 
   test "should have a fallback when trying to create a blank item with original claim but a timeout error happens" do
+    skip "This test is flaky and needs to be fixed. Ticket for it: CV2–CV2-6425"
```
caiosba, line 168: Minor: `CV2-CV2-`, but thanks for flagging it.

test/models/team_test.rb
```diff
@@ -907,6 +907,8 @@ def setup
     team_bot = create_bot_user(name: 'Team Bot', login: 'team_bot', default: false, approved: true)
 
     original_team = create_team(name: 'original_team')
+    team_user = create_team_user team: original_team, role: 'admin'
+    User.current = team_user.user
```
caiosba, line 911: Before the fix, was the test failing with this change?

caiosba (approved):
Good to go once tests pass. @danielevalverde, as part of your review, please confirm is these changes fix the integration test.

PR #2336, opened 2025-07-30
CV2-4795: Add a deprecation for createProjectMedia mutation

melsawy:
## Description

Add a deprecation warning about this endpoint (for new partners, we should now use the createFactCheck GraphQL mutation)

References: CV2-4795

## How to test?

Re-run automated tests

## Checklist

### DIFF
--- a/app/graph/types/mutation_type.rb
+++ b/app/graph/types/mutation_type.rb
@@ -25,7 +25,7 @@ class MutationType < BaseObject
   field :createAccountSource, mutation: AccountSourceMutations::Create
   field :destroyAccountSource, mutation: AccountSourceMutations::Destroy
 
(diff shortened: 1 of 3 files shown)


caiosba (changes requested):
Sawy, like I said in the ticket - the deprecation warning is still applied to the whole mutation. If a conditional deprecation warning is not supported, I suggest you at least update the message to something like "If you're creating an item of type 'blank' with an attached fact-check, please use the `createFactCheck` mutation instead. This mutation won't support that case in the near future.", something like that.

melsawy:
> Sawy, like I said in the ticket - the deprecation warning is still applied to the whole mutation. If a conditional deprecation warning is not supported, I suggest you at least update the message to something like "If you're creating an item of type 'blank' with an attached fact-check, please use the `createFactCheck` mutation instead. This mutation won't support that case in the near future.", something like that.

Done

caiosba (approved):

PR #2358, opened 2025-09-16
CV2-6552:  N+1 query (Annotation & TagText)

melsawy:
## Description

Fix N+1 query for the cached field `tags_as_sentence` by running direct query instead of using `&:map` method to load Tag and TagText

References: CV2-6552, CV2-6552

## How to test?

Re-run automated tests

## Checklist

### DIFF
--- a/app/models/concerns/project_media_cached_fields.rb
+++ b/app/models/concerns/project_media_cached_fields.rb
@@ -644,7 +644,8 @@ module ProjectMediaCachedFields
     end
 
     def recalculate_tags_as_sentence
-      self.get_annotations('tag').map(&:load).map(&:tag_text).uniq.join(', ')
+      tag_text_ids = Tag.where(annotation_type: 'tag', annotated_type: 'ProjectMedia', annotated_id: self.id).collect{|t| t.data['tag']}
+      TagText.where(id: tag_text_ids).map(&:text).join(', ')
     end
 
     def recalculate_sources_as_sentence


app/models/concerns/project_media_cached_fields.rb
```diff
@@ -644,7 +644,8 @@ def recalculate_report_status
     end
 
     def recalculate_tags_as_sentence
-      self.get_annotations('tag').map(&:load).map(&:tag_text).uniq.join(', ')
+      tag_text_ids = Tag.where(annotation_type: 'tag', annotated_type: 'ProjectMedia', annotated_id: self.id).collect{|t| t.data['tag']}
+      TagText.where(id: tag_text_ids).map(&:text).join(', ')
```
vasconsaurus, line 648: Suggestion: changing `map(&:text)` to `pluck(:text)`, this way we only load the value for text instead, instead of the whole `TagText` object.

app/models/concerns/project_media_cached_fields.rb
```diff
@@ -644,7 +644,8 @@ def recalculate_report_status
     end
 
     def recalculate_tags_as_sentence
-      self.get_annotations('tag').map(&:load).map(&:tag_text).uniq.join(', ')
+      tag_text_ids = Tag.where(annotation_type: 'tag', annotated_type: 'ProjectMedia', annotated_id: self.id).collect{|t| t.data['tag']}
```
vasconsaurus, line 647: Question: Would it be worth it to extract this to a method like `get_tag_annotations`, or is this silly? I ask because I like how it was written before, but it is still pretty clear what we are doing now.

test/controllers/registrations_controller_test.rb
```diff
@@ -151,7 +151,7 @@ def teardown
     assert_no_difference 'User.count' do
       post :create, params: { api_user: { password: p1, password_confirmation: p1, email: existing_user.email, login: 'test', name: 'Test' } }
       assert_response 401
-      assert_equal 'Please check your email. If an account with that email doesn’t exist, you should have received a confirmation email. If you don’t receive a confirmation e-mail, try to reset your password or get in touch with our support.', response.parsed_body.dig("errors", 0, "message")
+      assert_equal 'Please check your email. If an account with that email doesn’t exist, you should have received a confirmation email. If you don’t receive a confirmation e-mail, try to reset your password or get in touch with our support.', JSON.parse(response.parsed_body).dig("errors", 0, "message")
```
vasconsaurus, line 154: Dumb question: Why did you have to change this test to use JSON.parse?

vasconsaurus (approved):
@melsawy I had a couple questions, and one suggestion. But, it looks good to me.

app/models/concerns/project_media_cached_fields.rb
melsawy: No need for this, as get_annotations('tag') performs the same logic; however, I want to avoid the N+1 query.

test/controllers/registrations_controller_test.rb
melsawy: This is the correct behavior; I do not know how this one worked before. :smile:

PR #2415, opened 2026-03-25
CV2-6690: Slow DB Query

melsawy:
## Description

Based on the EXPLAIN analysis of the slow query, I implemented the following improvements:

- [x] Added a new data_json column to enable querying using structured JSON instead of YAML
- [x] Created an index on data_json to improve performance
- [x] Added an index on project_medias.team_id 
- [x] Migrate data_json for existing Tag annotations

References: CV2-6690

## How to test?

Re-run automated tests

## Checklist

### DIFF
--- a/app/models/annotations/tag.rb
+++ b/app/models/annotations/tag.rb
@@ -49,7 +49,7 @@ class Tag < ApplicationRecord
     tag_pms = {}
     TagText.where(text: texts, team_id: team.id).each do |tag_text|
       texts_to_ids[tag_text.text] = tag_text.id
-      existing_tags = Tag.where(annotated_id: ids, annotated_type: 'ProjectMedia').where("data = ?", { tag: tag_text.id }.with_indifferent_access.to_yaml)
+      existing_tags = Tag.where(annotation_type: 'tag', annotated_id: ids, annotated_type: 'ProjectMedia').where("data_json ->> 'tag' = ?", tag_text.id.to_s)
       tag_pms[tag_text.id] = existing_tags.map(&:annotated_id)
     end
 
@@ -58,7 +58,7 @@ class Tag < ApplicationRecord
     inputs.each do |input|
       tag = texts_to_ids[input[:tag]] || input[:tag]
       if ids.include?(input[:annotated_id].to_i) && !tag_pms[tag].include?(input[:annotated_id].to_i)
-        inserts << input.to_h.with_indifferent_access.reject{ |k, _v| k.to_sym == :tag }.merge({ annotation_type: 'tag', data: { tag: tag } })
+        inserts << input.to_h.with_indifferent_access.reject{ |k, _v| k.to_sym == :tag }.merge({ annotation_type: 'tag', data: { tag: tag }, data_json: { tag: tag} })
       end
     end
     result = Annotation.import inserts, validate: false, recursive: false, timestamps: true
--- a/app/models/annotations/tag.rb
+++ b/app/models/annotations/tag.rb
@@ -16,6 +16,7 @@ class Tag < ApplicationRecord
   after_commit :destroy_elasticsearch_tag, on: :destroy
   after_commit :apply_rules_and_actions, on: [:create]
   after_commit :update_tags_count
+  after_save :update_data_json
 
   def content
     { tag: self.tag_text, tag_text_id: self.tag }.to_json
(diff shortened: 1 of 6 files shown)


lib/tasks/migrate/20260325113748_add_data_json_field_to_annotation.rake
```diff
@@ -0,0 +1,25 @@
...
+    task migrate_tags_data_json: :environment do
+      started = Time.now.to_i
+      last_team_id = Rails.cache.read('check:migrate:migrate_tags_data_json') || 0
+      Team.where('id > ?', last_team_id).find_each do |team|
+        puts "\nProcessing team #{team.slug} ....\n"
+        team.project_medias.find_in_batches(batch_size: 1000) do |pms|
+          print '.'
+          ids = pms.pluck(:id)
+          tag_items = []
+          Tag.where(annotation_type: 'tag', annotated_type: 'ProjectMedia', annotated_id: ids)
```
caiosba, line 13: To make the migration even more idempotent, you can skip the tags where `data_json` is not null.

caiosba (approved):

lib/tasks/migrate/20260325113748_add_data_json_field_to_annotation.rake
melsawy: For Tags the data always include a value to refer to TagText id

caiosba: Sorry, I mean, the new column.

caiosba:
Under different circumstances, it would be good to migrate all data from YAML to JSON, but given the maintenance-mode situation and how many records we have, I think the scope of the change here is sufficient.

app/models/tag_text.rb
```diff
@@ -27,7 +27,7 @@ def calculate_tags_count
 
   # Performance here could be much better, and we are also considering only ProjectMedia tags
```
vasconsaurus, line 28: @melsawy, out of curiosity, do you think this will be an issue in the context of explainers and fact-checks? Or not at all, since they are different?

lib/tasks/migrate/20260325113748_add_data_json_field_to_annotation.rake
```diff
@@ -0,0 +1,25 @@
+namespace :check do
+  namespace :migrate do
+    # bundle exec rails check:migrate:migrate_tags_data_json
+    task migrate_tags_data_json: :environment do
+      started = Time.now.to_i
```
vasconsaurus, line 5: I think you didn't use started

app/models/annotations/tag.rb
```diff
@@ -16,6 +16,7 @@ class Tag < ApplicationRecord
   after_commit :destroy_elasticsearch_tag, on: :destroy
   after_commit :apply_rules_and_actions, on: [:create]
   after_commit :update_tags_count
+  after_save :update_data_json
```
vasconsaurus, line 19: We are running this as a callback, will we have issues with bulk tag creation?

vasconsaurus (commented):
Left a few questions, and one final one, in tag.rb, in line 52, inside `bulk_create`:
https://github.com/meedan/check-api/blob/5e97be610cd1201af5ebb4d34b31b4df984f546b/app/models/annotations/tag.rb#L52

Should we update this query to use the new column as well?

vasconsaurus:
> Under different circumstances, it would be good to migrate all data from YAML to JSON, but given the maintenance-mode situation and how many records we have, I think the scope of the change here is sufficient.

Was wondering about this!

app/models/annotations/tag.rb
melsawy: Yes so I update the bulk insert to include the data_json

lib/tasks/migrate/20260325113748_add_data_json_field_to_annotation.rake
melsawy: Used it to calculate the execution time

melsawy:
> Left a few questions, and one final one, in tag.rb, in line 52, inside `bulk_create`:
> 
> https://github.com/meedan/check-api/blob/5e97be610cd1201af5ebb4d34b31b4df984f546b/app/models/annotations/tag.rb#L52
> 
> Should we update this query to use the new column as well?

Both queries return same result but it's better to use new column

app/models/tag_text.rb
melsawy: No issue in the context of explainers and fact-checks

vasconsaurus (approved):
Thanks for looking into my questions!

PR #2430, opened 2026-05-20
CV2-6177: Sentry error PG::UntranslatableCharacter Error

melsawy:
## Description

Remove `'\u0000'` from message text to avoid PG::Untranslatable error

References: CV2-6177

## How to test?

Re-run automated tests

## Checklist

### DIFF
--- a/app/models/concerns/smooch_messages.rb
+++ b/app/models/concerns/smooch_messages.rb
@@ -292,7 +292,7 @@ module SmoochMessages
       # Define a text variable to hold short text
       text = []
       list.collect{ |m| JSON.parse(m) }.sort_by{ |m| m['received'].to_f }.each do |message|
-        message['text'].gsub!('\u0000', '') # Avoid PG::UntranslatableCharacter exception
+        message['text']&.gsub!('\u0000', '') # Avoid PG::UntranslatableCharacter exception
         if message['type'] == 'text'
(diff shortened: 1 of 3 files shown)


caiosba (approved):

app/models/concerns/smooch_messages.rb
```diff
@@ -292,6 +292,7 @@ def bundle_list_of_messages_to_items(list, last)
       # Define a text variable to hold short text
       text = []
       list.collect{ |m| JSON.parse(m) }.sort_by{ |m| m['received'].to_f }.each do |message|
+        message['text']&.gsub!('\u0000', '') # Avoid PG::UntranslatableCharacter exception
```
caiosba: @melsawy I just remembered that I recently worked on it for Suwali, and you need to use double-quotes instead of single-quotes (so, `"\u0000"` instead of `'\u0000'`), otherwise it won't match the zero-byte chracter, it will match the literal string `\u0000`.

caiosba (changes requested):

app/models/concerns/smooch_messages.rb
caiosba: Captura de tela de 2026-05-20 18-03-16

caiosba (approved):

PR #2435, opened 2026-06-10
CV2-6172: Prevent duplicate Sidekiq jobs

melsawy:
## Description

Added a new argument to ElasticSearchWorker  `enqueued_at` so I can skip job execution if the job is outdated.

References: CV2-6172

## How to test?

Added unit test

## Checklist

### DIFF
--- a/app/workers/elastic_search_worker.rb
+++ b/app/workers/elastic_search_worker.rb
@@ -6,7 +6,7 @@ class ElasticSearchWorker
 
   sidekiq_retry_in { |_count, _e| 3 }
 
-  def perform(model_data, options, type, enqueued_at = Time.now.utc.to_i)
+  def perform(model_data, options, type, enqueued_at = 0)
     model_data = begin YAML::load(model_data) rescue nil end
     unless model_data.nil?
       model = model_data[:klass].constantize.find_by_id model_data[:id]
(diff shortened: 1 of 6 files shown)


app/workers/elastic_search_worker.rb
```diff
@@ -15,7 +15,6 @@ def perform(model_data, options, type)
         ops = {
           'create_doc' => 'create_elasticsearch_doc_bg',
           'update_doc' => 'update_elasticsearch_doc_bg',
-          'update_doc_team' => 'update_elasticsearch_doc_team_bg',
```
melsawy, line 18: I found that this callback has already been removed from the code, so there's no need to include it in the list.

caiosba (approved):
Approving, assuming this only applies when all fields are updated, not just a single field, right? That’s how I understood it. Otherwise, it could miss some field updates and only apply the latest ones. Please confirm, @melsawy.

melsawy:
> Approving, assuming this only applies when all fields are updated, not just a single field, right? That’s how I understood it. Otherwise, it could miss some field updates and only apply the latest ones. Please confirm, @melsawy.

This applies when single or some fields update but the model not ProjectMedia 
for example update status, tasks or reports fields as in this case I check against the Annotation.updated_at field not the ProjectMedia.updated_at as the ProjectMedia object can update different fields so I avoid checking ProjectMedia.update_at

PR #2447, opened 2026-07-29
CV2-6724 : Sunset Check notification

melsawy:
## Description

Send notification emails to workspace admins based on workspace usage, with separate messaging for high-usage and low-usage workspaces, using the copy from the [Jessie shared documents](https://docs.google.com/document/d/1pCA4oDRCRL0TzESRxmuPQTQn_qy_TKBWRvBJx3fC7Lw/edit?tab=t.jlfz6hyyulfs).

References: CV2-6724 

## How to test?

Sent mails to Caio and Jessie

## Checklist


melsawy:
@caiosba FYI, I forwarded the emails to Jessie to get her feedback before merging the PR.

app/mailers/sunset_mailer.rb
```diff
@@ -0,0 +1,26 @@
+class SunsetMailer < ApplicationMailer
+  layout nil
+
+  def notify(type, user, workspace, workspace_url)
+    @name = user.name
+    # Set subject based on type (notify or download)
+    subject_type = type == 'download' ? 'download' : 'notify'
+    subject = I18n.t("mail_sunset.#{subject_type}_subject", app_name: CheckConfig.get('app_name'), workspace: workspace)
```
caiosba: Since the email body is not translated, I don't think we need to translate the title.

caiosba (commented):
Sawy, while reviewing this, I was thinking about one thing... I don't think we should automatically send the download link to all admins. My suggestion is to send only the notification, and then send the download link only to users who request it (or even better, display the link on Check upon login), once we confirm that the requester is still a legit admin of that workspace.

My main concern is exposing the data more than necessary when probably only a few users will actually require it.

Let's discuss this with Jessie.

app/mailers/sunset_mailer.rb
melsawy: Done

melsawy:
> Sawy, while reviewing this, I was thinking about one thing... I don't think we should automatically send the download link to all admins. My suggestion is to send only the notification, and then send the download link only to users who request it (or even better, display the link on Check upon login), once we confirm that the requester is still a legit admin of that workspace.
> 
> My main concern is exposing the data more than necessary when probably only a few users will actually require it.
> 
> Let's discuss this with Jessie.

Filed a new ticket for this one [CV2-6740](https://meedan.atlassian.net/browse/CV2-6740)

[CV2-6740]: https://meedan.atlassian.net/browse/CV2-6740?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

caiosba (approved):
Thanks for the fix, please re-validate the copy with Jessie since we agreed to change the workflow.

PR #2454, opened 2026-08-24
CV2-6723: add export rake task with readme file

melsawy:
## Description

Add rake tasks to export workspace data

- rails check:sunset:export_workspace_data[team-slug]
- rails check:sunset:export_workspace_user_data[team-slug]
- rails check:sunset:export_workspace_articles_data[team-slug]
- rails check:sunset:export_workspace_annotations_data[team-slug]
- rails check:sunset:export_workspace_tipline_requets_data[team-slug]
- rails check:sunset:export_workspace_tipline_newsletter_data[team-slug]
- rails check:sunset:export_workspace_item_data[team-slug]

There is a main rake task to do the export and sent an email with download link `check:sunset:export_and_upload_workspace_data[team-slug]`

Sample of README.txt file
[README.txt](https://github.com/user-attachments/files/31515891/README.txt)

## How to test?

## Checklist

### DIFF
--- a/lib/check_s3.rb
+++ b/lib/check_s3.rb
@@ -45,16 +45,17 @@ class CheckS3
     end
   end
 
-  def self.write(path, content_type, content)
+  def self.write(path, content_type, content, bucket=nil)
+    bucket ||= CheckConfig.get('storage_bucket')
     client = Aws::S3::Client.new
     client.put_object(
       acl: 'public-read',
       key: path,
       body: content,
-      bucket: CheckConfig.get('storage_bucket'),
+      bucket: bucket,
       content_type: content_type
     )
-    begin client.put_object_acl(acl: 'public-read', key: path, bucket: CheckConfig.get('storage_bucket')) rescue nil end
+    begin client.put_object_acl(acl: 'public-read', key: path, bucket: bucket) rescue nil end
   end
 
   def self.delete(*paths)
@@ -66,9 +67,9 @@ class CheckS3
     client.delete_objects(bucket: CheckConfig.get('storage_bucket'), delete: { objects: objects })
   end
 
-  def self.write_presigned(path, content_type, content, expires_in)
-    self.write(path, content_type, content)
-    bucket = CheckConfig.get('storage_bucket')
+  def self.write_presigned(path, content_type, content, expires_in, bucket=nil)
+    bucket ||= CheckConfig.get('storage_bucket')
+    self.write(path, content_type, content, bucket)
     client = Aws::S3::Client.new
     s3 = Aws::S3::Resource.new(client: client)
     obj = s3.bucket(bucket).object(path)
--- a/lib/check_s3.rb
+++ b/lib/check_s3.rb
@@ -45,8 +45,7 @@ class CheckS3
     end
   end
 
-  def self.write(path, content_type, content, bucket=nil)
-    is_public = bucket.nil?
+  def self.write(path, content_type, content, bucket=nil, acl = 'public-read')
     bucket ||= CheckConfig.get('storage_bucket')
     client = Aws::S3::Client.new
     object = {
@@ -55,9 +54,9 @@ class CheckS3
       bucket: bucket,
       content_type: content_type
     }
-    object[:acl] = 'public-read' if is_public
+    object[:acl] = acl unless acl.nil?
     client.put_object(object)
-    begin client.put_object_acl(acl: 'public-read', key: path, bucket: bucket) rescue nil end if is_public
+    begin client.put_object_acl(acl: 'public-read', key: path, bucket: bucket) rescue nil end unless acl.nil?
   end
 
   def self.delete(*paths)
@@ -69,9 +68,9 @@ class CheckS3
     client.delete_objects(bucket: CheckConfig.get('storage_bucket'), delete: { objects: objects })
   end
 
-  def self.write_presigned(path, content_type, content, expires_in, bucket=nil)
+  def self.write_presigned(path, content_type, content, expires_in, bucket=nil, acl = 'public-read')
     bucket ||= CheckConfig.get('storage_bucket')
-    self.write(path, content_type, content, bucket)
+    self.write(path, content_type, content, bucket, acl)
     client = Aws::S3::Client.new
     s3 = Aws::S3::Resource.new(client: client)
     obj = s3.bucket(bucket).object(path)
--- a/lib/check_s3.rb
+++ b/lib/check_s3.rb
@@ -56,7 +56,7 @@ class CheckS3
     }
     object[:acl] = acl unless acl.nil?
     client.put_object(object)
-    begin client.put_object_acl(acl: 'public-read', key: path, bucket: bucket) rescue nil end unless acl.nil?
+    begin client.put_object_acl(acl: acl, key: path, bucket: bucket) rescue nil end unless acl.nil?
   end
 
   def self.delete(*paths)
--- a/lib/check_s3.rb
+++ b/lib/check_s3.rb
@@ -46,16 +46,18 @@ class CheckS3
   end
 
   def self.write(path, content_type, content, bucket=nil)
+    is_public = bucket.nil?
     bucket ||= CheckConfig.get('storage_bucket')
     client = Aws::S3::Client.new
-    client.put_object(
-      acl: 'public-read',
+    object = {
       key: path,
       body: content,
       bucket: bucket,
       content_type: content_type
-    )
-    begin client.put_object_acl(acl: 'public-read', key: path, bucket: bucket) rescue nil end
+    }
+    object[:acl] = 'public-read' if is_public
+    client.put_object(object)
+    begin client.put_object_acl(acl: 'public-read', key: path, bucket: bucket) rescue nil end if is_public
   end
 
   def self.delete(*paths)
--- a/lib/tasks/data/check_sunset.rake
+++ b/lib/tasks/data/check_sunset.rake
@@ -199,7 +199,7 @@ namespace :check do
           'ID' => 'Unique identifier of the user.',
           'Name' => 'Name of the user.',
           'Email' => 'Email address of the user.',
-          'Joined data' => 'Date when the user joined the workspace.',
+          'Joined date' => 'Date when the user joined the workspace.',
           'Last access date' => 'Date when the user most recently accessed the workspace.',
           'Role' => 'User\'s role in the workspace.',
         }
@@ -349,7 +349,7 @@ namespace :check do
         headers = {
           'ID' => 'Unique identifier for the TiplineNewsletter.',
           'Header type' => 'Type of header used in the newsletter.',
-          'Hheader file' => 'File used as the newsletter header.',
+          'Header file' => 'File used as the newsletter header.',
           'Header overlay text' => 'Text displayed as an overlay on the newsletter header.',
           'Header media url' => 'URL of the media used in the newsletter header.',
           'Introduction' => 'Introduction text displayed in the newsletter.',
@@ -363,7 +363,7 @@ namespace :check do
           'Last sent at' => 'Date and time when the newsletter was last sent.',
           'Language' => 'Language of the TiplineNewsletter in two-letter format, such as en/ar/fr.',
           'Enabled' => 'Indicates whether the newsletter is enabled',
-          'Creation data' => 'Date and time when the newsletter was created.',
+          'Creation date' => 'Date and time when the newsletter was created.',
         }
         write_to_csv(file, headers.keys, data_csv)
         append_export_documentation(export_dir.join('README.txt'), 'workspace_tipline_newsletter_data.csv', 'Contains information about Tipline newsletters configured for the workspace.', headers)
@@ -437,13 +437,19 @@ namespace :check do
           compress_folder(folder_path, zip_path)
           # Save to S3
           bucket_name = ENV.fetch('EXPORT_OUTPUT_BUCKET')
-          zip_content = File.binread(zip_path)
-          s3_url = CheckS3.write_presigned("#{team.slug}/#{SecureRandom.hex(16)}/#{team.slug}.zip", 'application/zip', zip_content, 7.days.to_i, bucket_name)
-          key = Shortener::ShortenedUrl.generate!(s3_url).unique_key
-          download_url = CheckConfig.get('short_url_host') + '/' + key
-          puts "Download link (valid for 7 days): #{download_url}"
-          # Delete a directory for exported data and zip file
-          FileUtils.rm_rf(Rails.root.join('tmp', 'sunset'))
+          begin
+            zip_content = File.binread(zip_path)
+            s3_url = CheckS3.write_presigned("#{team.slug}/#{SecureRandom.hex(16)}/#{team.slug}.zip", 'application/zip', zip_content, 7.days.to_i, bucket_name, nil)
+            key = Shortener::ShortenedUrl.generate!(s3_url).unique_key
+            download_url = CheckConfig.get('short_url_host') + '/' + key
+            puts "Download link (valid for 7 days): #{download_url}"
+          rescue StandardError => e
+            puts "Failed to upload exported data #{e.message}"
+          ensure
+            # Delete a directory for exported data and zip file
+            FileUtils.rm_rf(folder_path)
+            File.delete(zip_path) if File.exist?(zip_path)
+          end
         end
       end
       minutes = ((Time.now.to_i - started) / 60).to_i
--- a/lib/tasks/data/check_sunset.rake
+++ b/lib/tasks/data/check_sunset.rake
@@ -439,7 +439,7 @@ namespace :check do
           bucket_name = ENV.fetch('EXPORT_OUTPUT_BUCKET')
           begin
             zip_content = File.binread(zip_path)
-            s3_url = CheckS3.write_presigned("#{team.slug}/#{SecureRandom.hex(16)}/#{team.slug}.zip", 'application/zip', zip_content, 7.days.to_i, bucket_name, nil)
+            s3_url = CheckS3.write_presigned("#{team.slug}/#{SecureRandom.hex(16)}/#{team.slug}.zip", 'application/zip', zip_content, 7.days.to_i, bucket_name, 'private')
             key = Shortener::ShortenedUrl.generate!(s3_url).unique_key
             download_url = CheckConfig.get('short_url_host') + '/' + key
             puts "Download link (valid for 7 days): #{download_url}"
--- a/lib/tasks/data/check_sunset.rake
+++ b/lib/tasks/data/check_sunset.rake
@@ -61,6 +61,11 @@ namespace :check do
           <<~README
             CSV Export Documentation
             ========================
+            This export contains information and data from the workspace in CSV format. Each CSV file contains a specific type of workspace data, such as workspace information, users, items, annotations, FactChecks, Explainers, and Tipline requests.
+
+            Media Files
+            ===========
+            Media files are not included directly in the CSV export. For items that contain media, the corresponding Media data or media URL in workspace_item_data.csv should be used to download the actual media files.
 
           README
         )
@@ -411,10 +416,10 @@ namespace :check do
         download_url = CheckS3.write_presigned("export/workspaces_data/#{team.slug}/#{Time.now.to_i}/#{team.slug}.zip", 'application/zip', zip_path, CheckConfig.get('export_csv_expire', 7.days.to_i, :integer))
         puts "Download link (valid 7 days): #{download_url}"
         # Send download link to workspace admins
-        # team.team_users.where(status: 'member').find_each do |tu|
-        #   puts "Sending email to #{tu.user.email}\n"
-        #   SunsetMailer.delay.notify(tu.user, tu.team.name)
-        # end
+        team.team_users.where(status: 'member').find_each do |tu|
+          puts "Sending email to #{tu.user.email}\n"
+          SunsetMailer.delay.notify(tu.user, tu.team.name)
+        end
       end
       minutes = ((Time.now.to_i - started) / 60).to_i
       puts "[#{Time.now}] Done in #{minutes} minutes."
--- a/lib/tasks/data/check_sunset.rake
+++ b/lib/tasks/data/check_sunset.rake
@@ -391,15 +391,16 @@ namespace :check do
       end
     end
 
-    # bundle exec rails check:sunset:export_upload_and_send_workspace_data[team-slug, email]
+    # bundle exec rails check:sunset:export_upload_and_send_workspace_data[team-slug, email] EXPORT_OUTPUT_BUCKET=XXXXX
     task :export_upload_and_send_workspace_data,[:slug, :email] => :environment do |_t, args|
       started = Time.now.to_i
       slug = args[:slug].to_s
       team = Team.find_by_slug slug
       unless team.nil?
         email = args[:email].to_s
+        puts "email:: #{email}"
         user = User.where(email: email).first
-        if team.team_users.where(user_id: user.id, role: 'admin', status: 'member').exists?
+        if user && team.team_users.where(user_id: user.id, role: 'admin', status: 'member').exists?
           # Call all exported tasks
           Rake::Task['check:sunset:export_workspace_init_readme'].invoke(args[:slug])
           Rake::Task['check:sunset:export_workspace_data'].invoke(args[:slug])
@@ -411,20 +412,20 @@ namespace :check do
           Rake::Task['check:sunset:export_workspace_item_data'].invoke(args[:slug])
           print_task_title 'Uploading workspace data'
           # compress & upload
-          folder_path = "#{Rails.root}/tmp/#{team.slug}"
-          zip_path = "#{Rails.root}/tmp/#{team.slug}.zip"
+          folder_path = Rails.root.join('tmp', team.slug)
+          zip_path = Rails.root.join('tmp', "#{team.slug}.zip")
           puts "Compressing #{folder_path} -> #{zip_path}"
           compress_folder(folder_path, zip_path)
           # Save to S3
+          bucket_name = ENV.fetch('EXPORT_OUTPUT_BUCKET')
           zip_content = File.binread(zip_path)
-          s3_url = CheckS3.write_presigned("export/workspaces_data/#{team.slug}.zip", 'application/zip', zip_content, CheckConfig.get('check_sunset_download_expire_days', 7, :integer).days.to_i)
+          s3_url = CheckS3.write_presigned("#{team.slug}/#{team.slug}.zip", 'application/zip', zip_content, 7.days.to_i, bucket_name)
           key = Shortener::ShortenedUrl.generate!(s3_url).unique_key
           download_url = CheckConfig.get('short_url_host') + '/' + key
-          puts "Download link (valid #{CheckConfig.get('check_sunset_download_expire_days', 7, :integer)} days): #{download_url}"
-          # Deleting exported files
-          puts "#{folder_path} -- #{zip_path}"
-          # File.delete(folder_path)
-          # File.delete(zip_path)
+          puts "Download link (valid for 7 days): #{download_url}"
+          # Delete a directory for exported data and zip file
+          FileUtils.rm_rf(folder_path)
+          File.delete(zip_path) if File.exist?(zip_path)
         end
       end
       minutes = ((Time.now.to_i - started) / 60).to_i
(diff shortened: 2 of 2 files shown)


lib/tasks/data/check_sunset.rake
```diff
@@ -0,0 +1,431 @@
...
+        zip_content = File.binread(zip_path)
+        s3_url = CheckS3.write_presigned("export/workspaces_data/#{team.slug}/#{Time.now.to_i}/#{team.slug}.zip", 'application/zip', zip_content, CheckConfig.get('check_sunset_download_expire_days', 7, :integer).days.to_i)
+        key = Shortener::ShortenedUrl.generate!(s3_url).unique_key
+        download_url = CheckConfig.get('short_url_host') + '/' + key
+        puts "Download link (valid #{CheckConfig.get('check_sunset_download_expire_days', 7, :integer)} days): #{download_url}"
+        # Send download link to workspace admins
+        # team.team_users.where(status: 'member').find_each do |tu|
+        #   puts "Sending email to #{tu.user.email}\n"
+        #   SunsetMailer.delay.notify('download', tu.user, tu.team.name)
+        # end
```
melsawy: I'll uncomment these lines after merge into epic branch as `SunsetMailer` exists in epic branch

lib/tasks/data/check_sunset.rake
```diff
@@ -0,0 +1,431 @@
...
+        Rake::Task['check:sunset:export_workspace_item_data'].invoke(args[:slug])
+        print_task_title 'Uploading workspace data'
+        # compress & upload
+        folder_path = "#{Rails.root}/public/#{team.slug}"
+        zip_path = "#{Rails.root}/public/#{team.slug}-#{Time.now.to_i}.zip"
+        puts "Compressing #{folder_path} -> #{zip_path}"
+        compress_folder(folder_path, zip_path)
+        # Save to S3
+        zip_content = File.binread(zip_path)
+        s3_url = CheckS3.write_presigned("export/workspaces_data/#{team.slug}/#{Time.now.to_i}/#{team.slug}.zip", 'application/zip', zip_content, CheckConfig.get('check_sunset_download_expire_days', 7, :integer).days.to_i)
```
caiosba: Is this S3 bucket public? I think so (it doesn't list, but the files inside are, if I'm not mistaken)... I suggest we use random names (e.g., `SecureRandom.hex`) for files, Sawy (the timestamp you added to the path helps, but we can be paranoid).

lib/tasks/data/check_sunset.rake
caiosba: As per [my comment in the other PR](https://github.com/meedan/check-api/pull/2447#pullrequestreview-5041001077), let's discuss this.

lib/tasks/data/check_sunset.rake
```diff
@@ -0,0 +1,431 @@
...
+      end
+    end
+
+    # Export workspace data
+    task :export_workspace_init_readme, [:slug] => :environment do |_t, args|
+      print_task_title 'CSV Export Documentation'
+      slug = args[:slug].to_s
+      team = Team.find_by_slug slug
+      unless team.nil?
+        export_dir = Rails.root.join("public", team.slug)
```
caiosba: Let's not save to the `public/` directory, not even temporarily... better save into `/tmp`.

lib/tasks/data/check_sunset.rake
caiosba: Actually, I think we should even create a new bucket only for the exports, with more restrict access rules (Jacopo can help if needed).

lib/tasks/data/check_sunset.rake
```diff
@@ -0,0 +1,431 @@
...
+      slug = args[:slug].to_s
+      team = Team.find_by_slug slug
+      unless team.nil?
+        # Call all exported tasks
+        Rake::Task['check:sunset:export_workspace_init_readme'].invoke(args[:slug])
+        Rake::Task['check:sunset:export_workspace_data'].invoke(args[:slug])
+        Rake::Task['check:sunset:export_workspace_user_data'].invoke(args[:slug])
+        Rake::Task['check:sunset:export_workspace_articles_data'].invoke(args[:slug])
+        Rake::Task['check:sunset:export_workspace_annotations_data'].invoke(args[:slug])
+        Rake::Task['check:sunset:export_workspace_tipline_requets_data'].invoke(args[:slug])
```
caiosba: Typo: `requets`

lib/tasks/data/check_sunset.rake
```diff
@@ -0,0 +1,431 @@
...
+          team_tasks.each do |tt|
+            headers[tt.label] = tt.description
+          end
+          write_to_csv(file, headers.keys, data_csv)
+          append_export_documentation(export_dir.join('README.txt'), 'workspace_annotations_data.csv', 'Contains information about workspace annotations. Each column represents an annotation label, and each row contains the corresponding annotation value for an item.', headers)
+        end
+      end
+    end
+    # bundle exec rails check:sunset:export_workspace_tipline_requets_data[team-slug]
+    task :export_workspace_tipline_requets_data,[:slug] => :environment do |_t, args|
```
caiosba: Typo: `requets`

lib/tasks/data/check_sunset.rake
```diff
@@ -0,0 +1,431 @@
...
+          data_csv << [
+            tr.id,
+            tr.language,
+            tr.created_at,
+            tr.tipline_user_uid,
+            tr.smooch_data['text']
+          ]
+        end
+        export_dir = Rails.root.join('public', team.slug)
+        file = export_dir.join('workspace_tipline_requets_data.csv')
```
caiosba: Typo: `requets`

lib/tasks/data/check_sunset.rake
```diff
@@ -0,0 +1,431 @@
...
+        headers = {
+          'ID' => 'Unique identifier of the TiplineRequest.',
+          'Language' => 'Language of the TiplineRequest in two-letter format, such as en/ar/fr.',
+          'Platform' => 'Platform through which the TiplineRequest was submitted.',
+          'Creation date' => 'Date and time when the TiplineRequest was created.',
+          'User UID' => 'Unique identifier of the tipline user who submitted the request.',
+          'Text' => 'Text submitted as part of the TiplineRequest.'
+        }
+        write_to_csv(file, headers.keys, data_csv)
+        append_export_documentation(export_dir.join('README.txt'), 'workspace_tipline_requets_data.csv', 'Contains information about requests submitted through workspace tiplines.', headers)
```
caiosba: Typo: `requets`

lib/tasks/data/check_sunset.rake
caiosba: And even if we were to send to all admins, that loop is iterating over all members, not filtering by admin role.

lib/tasks/data/check_sunset.rake
```diff
@@ -0,0 +1,431 @@
...
+      print_task_title 'Exporting workspace TiplineRequests data'
+      slug = args[:slug].to_s
+      team = Team.find_by_slug slug
+      unless team.nil?
+        data_csv = []
+        TiplineRequest.where(team_id: team.id).find_each do |tr|
+          print '.'
+          data_csv << [
+            tr.id,
+            tr.language,
```
caiosba, line 297: `tr.platform` is missing here

lib/tasks/data/check_sunset.rake
```diff
@@ -0,0 +1,431 @@
...
+            DynamicAnnotation::Field.select("a.annotated_id AS tid, dynamic_annotation_fields.value AS answer")
+            .joins("INNER JOIN annotations a ON a.id = dynamic_annotation_fields.annotation_id INNER JOIN annotations a2 ON a2.id = a.annotated_id")
+            .where("field_name LIKE 'response_%'")
+            .where('a.annotated_type' => 'Task', 'a2.annotated_type' => 'ProjectMedia', 'a2.annotated_id' => pm.id).each do |f|
+              a_answer[f.tid] = begin JSON.parse(f.answer) rescue f.answer end
+            end
+            next if a_answer.empty?
+            pm_raw = [pm.id]
+            headers_id.each do |ttid|
+              pm_raw << a_answer[a_ttid[ttid]] || '-'
```
caiosba: This is not working as intended - this parses as `(pm_raw << a_answer[a_ttid[ttid]]) || '-'`, so `'-'` is dead code... should be `pm_raw << (a_answer[a_ttid[ttid]] || '-')`.

lib/tasks/data/check_sunset.rake
```diff
@@ -0,0 +1,431 @@
...
+            end
+            data_csv << pm_raw
+          end
+          export_dir = Rails.root.join('public', team.slug)
+          file = export_dir.join('workspace_annotations_data.csv')
+          headers = { 'Item ID' => 'Unique identifier of the item associated with the annotation.' }
+          team_tasks.each do |tt|
+            headers[tt.label] = tt.description
```
caiosba: We should guard against two tasks with the same label... maybe `headers["#{tt.label} (#{tt.id})"]`?

lib/tasks/data/check_sunset.rake
```diff
@@ -0,0 +1,431 @@
...
+    end
+    # bundle exec rails check:sunset:export_workspace_articles_data[team-slug]
+    task :export_workspace_articles_data,[:slug] => :environment do |_t, args|
+      print_task_title 'Exporting workspace articles data (FactChecks & Explainers)'
+      slug = args[:slug].to_s
+      team = Team.find_by_slug slug
+      unless team.nil?
+        export_dir = Rails.root.join('public', team.slug)
+        # Export FactChecks
+        data_csv = FactCheck.get_exported_data({}, team).drop(1)
```
caiosba, line 218: Make sure the headers are the same (at least same count)

lib/tasks/data/check_sunset.rake
```diff
@@ -0,0 +1,431 @@
...
+          'Summary' => 'Summary of the FactCheck',
+          'URL' => 'URL of the FactCheck',
+          'Language' => 'Language of the FactCheck in two-letter format, such as en/ar/fr.',
+          'Report Status' => 'Current status of the FactCheck report.',
+          'Imported?' => 'Indicates whether the FactCheck was imported into the workspace'
+        }
+        write_to_csv(file, headers.keys, data_csv)
+        append_export_documentation(export_dir.join('README.txt'), 'workspace_fact_checks_data.csv', 'Contains information about FactChecks in the workspace.', headers)
+        # Export Explainers
+        data_csv = Explainer.get_exported_data({}, team).drop(1)
```
caiosba, line 232: Make sure the headers are the same (at least same count)

lib/tasks/data/check_sunset.rake
```diff
@@ -0,0 +1,431 @@
...
+        # compress & upload
+        folder_path = "#{Rails.root}/public/#{team.slug}"
+        zip_path = "#{Rails.root}/public/#{team.slug}-#{Time.now.to_i}.zip"
+        puts "Compressing #{folder_path} -> #{zip_path}"
+        compress_folder(folder_path, zip_path)
+        # Save to S3
+        zip_content = File.binread(zip_path)
+        s3_url = CheckS3.write_presigned("export/workspaces_data/#{team.slug}/#{Time.now.to_i}/#{team.slug}.zip", 'application/zip', zip_content, CheckConfig.get('check_sunset_download_expire_days', 7, :integer).days.to_i)
+        key = Shortener::ShortenedUrl.generate!(s3_url).unique_key
+        download_url = CheckConfig.get('short_url_host') + '/' + key
```
caiosba: We should clean up once the file is uploaded to S3 (e.g., delete it from disk - ZIP and CSVs).

caiosba (changes requested):

caiosba:
In summary, in terms of security, most important things here are:

- [x] Don't save temporary files to `public/`
- [x] Save to S3 in a new bucket with more restrict access rules and with random names
- [x] Clean up files once uploaded to S3

@melsawy, after that, please generate an export and ask for Jessie's review.

lib/tasks/data/check_sunset.rake
```diff
@@ -0,0 +1,431 @@
...
+        puts "Download link (valid #{CheckConfig.get('check_sunset_download_expire_days', 7, :integer)} days): #{download_url}"
+        # Send download link to workspace admins
+        # team.team_users.where(status: 'member').find_each do |tu|
+        #   puts "Sending email to #{tu.user.email}\n"
+        #   SunsetMailer.delay.notify('download', tu.user, tu.team.name)
+        # end
+      end
+      minutes = ((Time.now.to_i - started) / 60).to_i
+      puts "[#{Time.now}] Done in #{minutes} minutes."
```
caiosba, line 456: As we just discussed in our 1:1, let's also save a record (new model) to a new table where we map workspace to shortened URL, plus any other metadata we need, like the user who asked for it, when it was requested (so we can stop showing it after the expiration), etc.

lib/tasks/data/check_sunset.rake
```diff
@@ -0,0 +1,431 @@
...
+          'Media type' => 'Type of the item, such as Claim, Link, Image, Video, or Audio.',
+          'Media data' => 'Content or reference associated with the item, such as a Quote, URL, or file path.',
+        }
+        write_to_csv(file, headers.keys, data_csv)
+        append_export_documentation(export_dir.join('README.txt'), 'workspace_item_data.csv', 'Contains information about workspace items, including Claims, Links, and Media.', headers)
+      end
+    end
+
+    # bundle exec rails check:sunset:export_and_upload_workspace_data[team-slug]
+    task :export_and_upload_workspace_data,[:slug] => :environment do |_t, args|
```
caiosba: Looks like it should be `export_upload_and_send_workspace_data`. My suggestion is that this rake task also takes a single destination email as argument, and validates that there is an admin user in that workspace with that email.

lib/tasks/data/check_sunset.rake
melsawy: Removed this part and updated rake task to accept `email` as an argument and verify that email is an admin member and should handle the download ticket on another ticket [CV2-6740](https://meedan.atlassian.net/browse/CV2-6740)

lib/tasks/data/check_sunset.rake
melsawy: Done

lib/tasks/data/check_sunset.rake
melsawy: Done

lib/tasks/data/check_sunset.rake
melsawy: Done

lib/tasks/data/check_sunset.rake
melsawy: Done!, Good catch Dr.Caio

lib/tasks/data/check_sunset.rake
melsawy: Done

lib/tasks/data/check_sunset.rake
melsawy: Filed a new ticket to handle download link [CV2-6740](https://meedan.atlassian.net/browse/CV2-6740)

lib/tasks/data/check_sunset.rake
melsawy: Done

lib/tasks/data/check_sunset.rake
melsawy: I depend on passing the bucket name as ENV variable to rake task `EXPORT_OUTPUT_BUCKET`

lib/tasks/data/check_sunset.rake
melsawy: I got the header from the [export method](https://github.com/meedan/check-api/blob/develop/app/models/fact_check.rb#L65)

lib/tasks/data/check_sunset.rake
melsawy: I got the header from the [export method](https://github.com/meedan/check-api/blob/develop/app/models/explainer.rb#L52)

lib/tasks/data/check_sunset.rake
melsawy: Done

lib/tasks/data/check_sunset.rake
caiosba: Still missing

lib/tasks/data/check_sunset.rake
caiosba: Still missing

lib/tasks/data/check_sunset.rake
caiosba: Still missing

lib/tasks/data/check_sunset.rake
```diff
@@ -17,5 +65,389 @@ namespace :check do
...
+          print_task_title 'Uploading workspace data'
+          # compress & upload
+          folder_path = Rails.root.join('tmp', team.slug)
+          zip_path = Rails.root.join('tmp', "#{team.slug}.zip")
+          puts "Compressing #{folder_path} -> #{zip_path}"
+          compress_folder(folder_path, zip_path)
+          # Save to S3
+          bucket_name = ENV.fetch('EXPORT_OUTPUT_BUCKET')
+          zip_content = File.binread(zip_path)
+          s3_url = CheckS3.write_presigned("#{team.slug}/#{team.slug}.zip", 'application/zip', zip_content, 7.days.to_i, bucket_name)
```
caiosba: This path is still easy to guess, please add some random argument to the path, like `SecureRandom.hex(16)`.

lib/check_s3.rb
```diff
@@ -45,16 +45,17 @@ def self.get(path)
...
     client.put_object(
       acl: 'public-read',
       key: path,
       body: content,
-      bucket: CheckConfig.get('storage_bucket'),
+      bucket: bucket,
       content_type: content_type
     )
-    begin client.put_object_acl(acl: 'public-read', key: path, bucket: CheckConfig.get('storage_bucket')) rescue nil end
+    begin client.put_object_acl(acl: 'public-read', key: path, bucket: bucket) rescue nil end
```
caiosba: You're still saving the file with `public-read`.

caiosba: You're still saving the file with `public-read`.

lib/tasks/data/check_sunset.rake
```diff
@@ -17,5 +65,389 @@ namespace :check do
...
+          Rake::Task['check:sunset:export_workspace_data'].invoke(args[:slug])
+          Rake::Task['check:sunset:export_workspace_user_data'].invoke(args[:slug])
+          Rake::Task['check:sunset:export_workspace_articles_data'].invoke(args[:slug])
+          Rake::Task['check:sunset:export_workspace_annotations_data'].invoke(args[:slug])
+          Rake::Task['check:sunset:export_workspace_tipline_requests_data'].invoke(args[:slug])
+          Rake::Task['check:sunset:export_workspace_tipline_newsletter_data'].invoke(args[:slug])
+          Rake::Task['check:sunset:export_workspace_item_data'].invoke(args[:slug])
+          print_task_title 'Uploading workspace data'
+          # compress & upload
+          folder_path = Rails.root.join('tmp', team.slug)
```
caiosba: Let's avoid directly to `tmp` root... unlikely, but let's say we have a team slug `tmp` or `cache`... it would collide with Rails' defaults. In any case, is better to make sure it's more self contained, for example, `tmp/sunset`.

caiosba (changes requested):

lib/tasks/data/check_sunset.rake
melsawy: Done

lib/tasks/data/check_sunset.rake
melsawy: Done

lib/tasks/data/check_sunset.rake
melsawy: Done

lib/tasks/data/check_sunset.rake
melsawy: Done

lib/tasks/data/check_sunset.rake
melsawy: Done and the path is `tmp/sunset`

lib/check_s3.rb
melsawy: Since this method already exists and is used elsewhere, I kept the ACL behavior based on the bucket configuration. If we’re using the default bucket (the existing behavior), I’ll set the ACL to public-read; otherwise, I’ll leave it unset.

melsawy: Since this method already exists and is used elsewhere, I kept the ACL behavior based on the bucket configuration. If we’re using the default bucket (the existing behavior), I’ll set the ACL to public-read; otherwise, I’ll leave it unset.

caiosba: I think it's better to pass an `acl` parameter that is `public-read` by default and then the sunset task can explicitly pass `acl = 'private'` instead of leaving it `nil`.

lib/check_s3.rb
```diff
@@ -66,9 +69,9 @@ def self.delete(*paths)
     client.delete_objects(bucket: CheckConfig.get('storage_bucket'), delete: { objects: objects })
   end
 
-  def self.write_presigned(path, content_type, content, expires_in)
-    self.write(path, content_type, content)
-    bucket = CheckConfig.get('storage_bucket')
+  def self.write_presigned(path, content_type, content, expires_in, bucket=nil)
```
caiosba: I think it's better to pass an acl parameter that is public-read by default and then the sunset task can explicitly pass acl = 'private' instead of leaving it nil.

lib/tasks/data/check_sunset.rake
```diff
@@ -17,5 +65,389 @@ namespace :check do
...
+            tn.enabled,
+            tn.created_at,
+          ]
+        end
+        export_dir = Rails.root.join('tmp', 'sunset', team.slug)
+        file = export_dir.join('workspace_tipline_newsletter_data.csv')
+        headers = {
+          'ID' => 'Unique identifier for the TiplineNewsletter.',
+          'Header type' => 'Type of header used in the newsletter.',
+          'Hheader file' => 'File used as the newsletter header.',
```
caiosba: Typo: `Hheader`

lib/tasks/data/check_sunset.rake
```diff
@@ -17,5 +65,389 @@ namespace :check do
...
+          compress_folder(folder_path, zip_path)
+          # Save to S3
+          bucket_name = ENV.fetch('EXPORT_OUTPUT_BUCKET')
+          zip_content = File.binread(zip_path)
+          s3_url = CheckS3.write_presigned("#{team.slug}/#{SecureRandom.hex(16)}/#{team.slug}.zip", 'application/zip', zip_content, 7.days.to_i, bucket_name)
+          key = Shortener::ShortenedUrl.generate!(s3_url).unique_key
+          download_url = CheckConfig.get('short_url_host') + '/' + key
+          puts "Download link (valid for 7 days): #{download_url}"
+          # Delete a directory for exported data and zip file
+          FileUtils.rm_rf(Rails.root.join('tmp', 'sunset'))
```
caiosba: It's better to delete only the directory related to this export, in case we have parallel exports. Also, this should be inside an `ensure` block, this way we confirm that the files can be deleted even if the task fails.

lib/tasks/data/check_sunset.rake
```diff
@@ -17,5 +65,389 @@ namespace :check do
...
+          'RSS feed url' => 'URL of the RSS feed used to populate the newsletter content.',
+          'Number of articles' => 'Number of articles included in the newsletter.',
+          'First article' => 'Content of the first article',
+          'Second article' => 'Content of the second article',
+          'Third article' => 'Content of the third article',
+          'Footer' => 'Footer text displayed in the newsletter.',
+          'Last sent at' => 'Date and time when the newsletter was last sent.',
+          'Language' => 'Language of the TiplineNewsletter in two-letter format, such as en/ar/fr.',
+          'Enabled' => 'Indicates whether the newsletter is enabled',
+          'Creation data' => 'Date and time when the newsletter was created.',
```
caiosba: Typo: Should be "Creation date"

lib/tasks/data/check_sunset.rake
```diff
@@ -17,5 +65,389 @@ namespace :check do
...
+            tu.role
+          ]
+        end
+        export_dir = Rails.root.join('tmp', 'sunset', team.slug)
+        file = export_dir.join('workspace_user_data.csv')
+        headers = {
+          'ID' => 'Unique identifier of the user.',
+          'Name' => 'Name of the user.',
+          'Email' => 'Email address of the user.',
+          'Joined data' => 'Date when the user joined the workspace.',
```
caiosba: Typo: Should be a "joined date"

caiosba (changes requested):
Almost there @melsawy, I still found a few outstanding issues.

lib/check_s3.rb
melsawy: I removed the `acl` parameter. By removing it, the upload will fall back to the bucket’s default permissions.

caiosba: I think we should force it as `private` for the sunset task, otherwise if there is a misconfiguration on the S3 bucket, we're exposed. Or is there any technical limitation/complication here?

melsawy: By leaving it unset, the upload will fall back to the bucket’s default permissions

lib/tasks/data/check_sunset.rake
melsawy: Done

lib/tasks/data/check_sunset.rake
melsawy: Done

lib/tasks/data/check_sunset.rake
melsawy: Done

lib/tasks/data/check_sunset.rake
melsawy: Done

lib/check_s3.rb
melsawy: I pass acl with default `public-read` to keep current calling works as expected and pass it with `nil` t avoid setting acl 
for sunset task as by leaving it unset, the upload will fall back to the bucket’s default permissions

caiosba (commented):
@melsawy I think we should force the `acl` as private for the sunset task, otherwise if there is a misconfiguration on the S3 bucket, we're exposed. Or is there any technical limitation/complication here?

melsawy:
> @melsawy I think we should force the `acl` as private for the sunset task, otherwise if there is a misconfiguration on the S3 bucket, we're exposed. Or is there any technical limitation/complication here?

Recommendation to leave the acl parameter out of  `put_object`  to make it compatible with modern cloud security standards, and safe from unexpected crashes.

caiosba:
> > @melsawy I think we should force the `acl` as private for the sunset task, otherwise if there is a misconfiguration on the S3 bucket, we're exposed. Or is there any technical limitation/complication here?
> 
> Recommendation to leave the acl parameter out of  `put_object`  to make it compatible with modern cloud security standards, and safe from unexpected crashes.

But even for our use case? It would be better if it crashes if the bucket allows for public-read objects.

caiosba (commented):
> Recommendation to leave the acl parameter out of  `put_object`  to make it compatible with modern cloud security standards, and safe from unexpected crashes.

But even for our use case? It would be better if it crashes if the bucket allows for public-read objects.

lib/check_s3.rb
melsawy: Done and set acl:private

lib/check_s3.rb
```diff
@@ -45,16 +45,18 @@ def self.get(path)
...
       body: content,
-      bucket: CheckConfig.get('storage_bucket'),
+      bucket: bucket,
       content_type: content_type
-    )
-    begin client.put_object_acl(acl: 'public-read', key: path, bucket: CheckConfig.get('storage_bucket')) rescue nil end
+    }
+    object[:acl] = acl unless acl.nil?
+    client.put_object(object)
+    begin client.put_object_acl(acl: 'public-read', key: path, bucket: bucket) rescue nil end unless acl.nil?
```
caiosba: `acl` is hard-coded as `public-read` here, you should pass `acl: acl`.

caiosba (changes requested):

lib/check_s3.rb
melsawy: Sorry Caio forgot to change this line

caiosba (approved):
Thanks @melsawy , everything looks good to me now! Once you merge/deploy, please share a download URL so we can do a final review.

PR #2455, opened 2026-09-02
CV2-6740: Add a download button for the Check export

melsawy:
## Description

Add a download link to Check app to allow admin user to download exported data

- [x] Update the `export_upload_and_send_workspace_data` rake task to accept workspace slug and email.
- [x] Verify that the email belongs to an active member of that workspace with admin role.
- [x] Add a new CheckExportedData model
- [x] Regenerate a download link 
- [x] Modify notification email copy

References: CV2-6740

## How to test?

Add unit tests

## Checklist

### DIFF
--- a/app/graph/types/me_type.rb
+++ b/app/graph/types/me_type.rb
@@ -171,4 +171,14 @@ class MeType < DefaultObject
     FeedInvitation.where(email: object.email)
   end
 
+  field :workspaces_download_url, JsonStringType, null: true
+
+  def workspaces_download_url
+    ability = context[:ability] || Ability.new
+    ret = {}
+    object.check_data_exports.includes(:team).find_each do |de|
+      ret[de.team.name] = de.download_url if ability.can?(:read, de)
+    end
+    ret
+  end
 end
--- a/app/graph/types/me_type.rb
+++ b/app/graph/types/me_type.rb
@@ -174,10 +174,12 @@ class MeType < DefaultObject
   field :workspaces_download_url, JsonStringType, null: true
 
   def workspaces_download_url
-    ability = context[:ability] || Ability.new
     ret = {}
+    ability = context[:ability] || Ability.new
+    expire_days = CheckConfig.get('check_sunset_download_expire_days', 15, :integer)
     object.check_data_exports.includes(:team).find_each do |de|
-      ret[de.team.name] = de.download_url if ability.can?(:read, de)
+      expired = Time.current > de.generated_at + expire_days.days
+      ret[de.team.name] = de.download_url if !expired && ability.can?(:read, de)
     end
     ret
   end
--- a/app/graph/types/me_type.rb
+++ b/app/graph/types/me_type.rb
@@ -178,8 +178,7 @@ class MeType < DefaultObject
     ret = {}
     ability = context[:ability] || Ability.new
     object.check_data_exports.includes(:team).find_each do |de|
-      expired = Time.current > de.expired_at
-      ret[de.team.name] = de.download_url if !expired && ability.can?(:read, de)
+      ret[de.team.name] = de.download_url if ability.can?(:read, de)
     end
     ret
   end
--- a/app/graph/types/me_type.rb
+++ b/app/graph/types/me_type.rb
@@ -171,14 +171,14 @@ class MeType < DefaultObject
     FeedInvitation.where(email: object.email)
   end
 
-  field :workspaces_download_url, JsonStringType, null: true
+  field :workspaces_data_export_url, JsonStringType, null: true
 
-  def workspaces_download_url
+  def workspaces_data_export_url
+    # List the CheckDataExport download URL for each workspace based on permissions and expiration date, using the key => value format
     ret = {}
     ability = context[:ability] || Ability.new
-    expire_days = CheckConfig.get('check_sunset_download_expire_days', 15, :integer)
     object.check_data_exports.includes(:team).find_each do |de|
-      expired = Time.current > de.generated_at + expire_days.days
+      expired = Time.current > de.expired_at
       ret[de.team.name] = de.download_url if !expired && ability.can?(:read, de)
     end
     ret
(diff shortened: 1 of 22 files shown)


app/graph/types/me_type.rb
```diff
@@ -171,4 +171,16 @@ def feed_invitations
     FeedInvitation.where(email: object.email)
   end
 
+  field :workspaces_data_export_url, JsonStringType, null: true
+
+  def workspaces_data_export_url
+    # List the CheckDataExport download URL for each workspace based on permissions and expiration date, using the key => value format
+    ret = {}
+    ability = context[:ability] || Ability.new
```
caiosba, line 179: I would just double-confirm here that `User.current`exists.

lib/check_basic_abilities.rb
```diff
@@ -39,6 +39,11 @@ def authenticated_perms
       tmp.archived = CheckArchivedFlags::FlagCodes::NONE
       can?(:update, tmp)
     end
+
+    can :read, CheckDataExport do |obj|
+      is_admin_member = TeamUser.where(user_id: @user.id, team_id: obj.team_id, role: 'admin', status: 'member').exists?
+      obj.user_id == @user.id && is_admin_member
+    end
```
caiosba, line 46: Please add a test for this.

app/models/check_data_export.rb
```diff
@@ -0,0 +1,45 @@
+class CheckDataExport < ApplicationRecord
+  belongs_to :team
+  belongs_to :user
+
+  validates_presence_of :team_id, :user_id, :download_url
+  validates_uniqueness_of :team_id
```
caiosba, line 6: Please also add a unique index for it in the database, just to confirm we won't have duplicate exports.

db/migrate/20260901165331_create_check_data_exports.rb
```diff
@@ -0,0 +1,14 @@
+class CreateCheckDataExports < ActiveRecord::Migration[6.1]
+  def change
+    create_table :check_data_exports do |t|
+      t.references :user, foreign_key: true
+      t.references :team, foreign_key: true
+      t.string :s3_key
+      t.string :download_url, null: false
+      t.datetime :generated_at
+      t.datetime :expired_at
```
caiosba: It's safer to make sure this is `NOT NULL`.

lib/check_basic_abilities.rb
```diff
@@ -39,6 +39,11 @@ def authenticated_perms
       tmp.archived = CheckArchivedFlags::FlagCodes::NONE
       can?(:update, tmp)
     end
+
+    can :read, CheckDataExport do |obj|
+      is_admin_member = TeamUser.where(user_id: @user.id, team_id: obj.team_id, role: 'admin', status: 'member').exists?
+      obj.user_id == @user.id && is_admin_member
```
caiosba: I think we should add the `expired_at` condition here.

caiosba (approved):
Approving to unblock testing on QA as you asked @melsawy , but there are a few things we need to fix/change in a follow-up PR before production (and that don't change the emails or user-facing stuff)...

To be honest, after reading the code, there is complexity and fragility in regenerating the presigned URL in order to have it available for 15 days when the max S3 validity is 7 days... and I'm not even sure if 7 days is the maximum for us... did you confirm if we're using long-lived IAM user credentials or an ECS task role?

So I had another idea... we could have a route and controller method that returns the presigned URL, valid for just a few minutes (so we don't need to worry about my question above, which I'm afraid we don't support for 7 days)... every time we load that URL, generates a new short-living presigned URL. This way we can have authentication for that endpoint, which means we don't need to worry about revocation, leakage, regeneration, enumeration, etc. I'm concerned that the search space for our URL shortener is too small (just 5 characters), and there is no authentication for that URL itself.

app/models/check_data_export.rb
melsawy: Done

db/migrate/20260901165331_create_check_data_exports.rb
melsawy: Done

lib/check_basic_abilities.rb
melsawy: I already added `expired_at`  to graphql field but yes I think here is better

melsawy: Done

app/graph/types/me_type.rb
melsawy: No need to confirm about `User.current` as `ability` will return false if User.current is nil

lib/check_basic_abilities.rb
melsawy: Done

melsawy:
> Approving to unblock testing on QA as you asked @melsawy , but there are a few things we need to fix/change in a follow-up PR before production (and that don't change the emails or user-facing stuff)...
> 
> To be honest, after reading the code, there is complexity and fragility in regenerating the presigned URL in order to have it available for 15 days when the max S3 validity is 7 days... and I'm not even sure if 7 days is the maximum for us... did you confirm if we're using long-lived IAM user credentials or an ECS task role?
> 
I Confirmed that 7 days is the maximum for us and yes regenerating code is complex so I tried to add a unit test to cover our case which is available for 15 days when the max S3 validity is 7 days and all [tests passed ](https://github.com/meedan/check-api/blob/CV2-6740-add-a-download-button-for-the-check-export/test/models/check_data_export_test.rb#L78-L114)

> So I had another idea... we could have a route and controller method that returns the presigned URL, valid for just a few minutes (so we don't need to worry about my question above, which I'm afraid we don't support for 7 days)... every time we load that URL, generates a new short-living presigned URL. This way we can have authentication for that endpoint, which means we don't need to worry about revocation, leakage, regeneration, enumeration, etc. I'm concerned that the search space for our URL shortener is too small (just 5 characters), and there is no authentication for that URL itself.

Agreed

PR #2458, opened 2026-09-15
CV2-6741: Add CheckDataExport mutation

melsawy:
## Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context, type of change (bug fix, performance, feature, etc.) and things to pay attention to during review.

References: TICKET-ID-1, TICKET-ID-2, …, TICKET-ID-N

## How to test?

## Checklist


db/migrate/20260915075134_change_column_null_in_check_data_exports.rb
```diff
@@ -0,0 +1,9 @@
+class ChangeColumnNullInCheckDataExports < ActiveRecord::Migration[6.1]
+  def change
+    change_column_null :check_data_exports, :download_url, true
+    change_column_null :check_data_exports, :expired_at, true
```
caiosba, line 4: Now that this column can be null, the ability may crash... see this condition:

```ruby
obj.user_id == @user.id && is_admin_member && Time.current < obj.expired_at
```

We might need to check that `expired_at` is not null, otherwise the `/me` query can crash.

app/graph/mutations/check_data_export_mutations.rb
```diff
@@ -0,0 +1,8 @@
+module CheckDataExportMutations
+  MUTATION_TARGET = 'check_data_export'.freeze
+  PARENTS = [].freeze
+
+  class Create < Mutations::CreateMutation
+    argument :status, GraphQL::Types::String, required: false
```
caiosba: Blocker: We shouldn't allow the API client set the status... please make it a default, or model callback.

app/graph/types/team_type.rb
```diff
@@ -432,4 +432,10 @@ def webhooks
     webhook_installations = object.team_users.joins(:user).where('users.type' => 'BotUser', 'users.default' => false).select{ |team_user| team_user.user.events.present? && team_user.user.get_request_url.present? && !team_user.user.get_approved }
     webhook_installations.map(&:user)
   end
+
+  field :check_data_export, CheckDataExportType, null: true
+
+  def check_data_export
+    object.check_data_export
```
caiosba: We should check for a `read` permission here as well, so non-admins don't see the export information.

caiosba (changes requested):

db/migrate/20260915075134_change_column_null_in_check_data_exports.rb
melsawy: Added `obj.expired_at &&`  to condition

app/graph/mutations/check_data_export_mutations.rb
melsawy: The default status is requested

app/graph/types/team_type.rb
melsawy: Done

caiosba (approved):
Thanks Sawy for addressing the comments, all good!

